PR #1, opened 2019-09-24
Update k8s libraries and use patch instead of update in cluster resource controller

katrogan:


Gopkg.toml
```diff
@@ -104,6 +108,10 @@
   name = "github.com/json-iterator/go"
   version = "^1.1.5"
 
+[[constraint]]
```
EngHabu:
```suggestion
[[override]]
```

PR #3, opened 2019-09-26
Add Namespace Mapping Configuration

migueltol22:
Currently Flyte makes a 1 to 1 mapping of project-domain to kubernetes namespace. This should be   a configurable option where users can map the domain to a kubernetes namespace or keep the existing mapping.


pkg/runtime/interfaces/propeller_executor_configuration.go
```diff
@@ -0,0 +1,11 @@
+// package interfaces
```
migueltol22: @katrogan I'd like to abstract the config for propeller executor to a propeller executor config that contains both cluster configuration and namespace mapping configuration. This follows more of a OOP paradigm and I'm a little rusty with go, so could you recommend a way to make this as go idiomatic as possible.

pkg/common/namespace.go
```diff
@@ -0,0 +1,15 @@
+package common
+
+import "fmt"
+
+const namespaceFormat = "%s-%s"
+const namespaceVariable = "namespace"
+const domainVariable = "domain"
+
+// Return kubernetes namespace name
+func GetNamespaceName(mapping, project, domain string) string {
```
katrogan: stylistic nit: instead of using a string for mapping can we use an enum (https://yourbasic.org/golang/iota/) so we have a defined set of mapping modes in one location?

also can we use a switch statement below?

pkg/runtime/cluster_resource_provider.go
```diff
@@ -43,6 +43,6 @@ func (p *ClusterResourceConfigurationProvider) GetRefreshInterval() time.Duratio
 	return time.Minute
 }
 
-func NewNamespaceConfigurationProvider() interfaces.ClusterResourceConfiguration {
+func NewClusterResourceConfigurationProvider() interfaces.ClusterResourceConfiguration {
```
katrogan: thank you :-)

pkg/runtime/interfaces/propeller_executor_configuration.go
katrogan: this looks fine!

migueltol22:
@katrogan Mind taking another look. Looks like with the changes made propeller executor config is no longer necessary so keeping with only namespace mapping configuration and will remove propeller config files if everything else looks good.

pkg/common/namespace_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+func TestGetNamespaceName(t *testing.T) {
+	testCases := []struct {
+		mapping NamespaceMapping
+		project string
+		domain  string
+		want    string
+	}{
+		{Domain, "project", "production", "project-production"},
+		{"bad-value", "project", "development", "project-development"},
```
katrogan: does this still compile?

pkg/runtime/namespace_config_provider.go
```diff
@@ -0,0 +1,40 @@
+package runtime
```
katrogan: does this file need to be updated?

pkg/clusterresource/controller.go
```diff
@@ -282,7 +282,8 @@ func (c *controller) Sync(ctx context.Context) error {
 	var errs = make([]error, 0)
 	for _, project := range projects {
 		for _, domain := range *domains {
-			namespace := fmt.Sprintf(namespaceFormat, project.Identifier, domain.Name)
+			namespaceMapping = c.config.NamespaceMappingConfiguration().GetNamespaceMappingConfig()
+			namespace := common.GetNamespaceName(namespaceMapping, project.Identifier, domain.Name)
```
katrogan: nit, no need to declare a new variable for the namespaceMapping, you can always just inline the GetNamespaceMappingConfig call

pkg/common/namespace.go
```diff
@@ -0,0 +1,25 @@
+package common
+
+import "fmt"
+
+type NamespaceMapping int
```
katrogan: looks like the type you declare in the runtime config is string. might be easier to make this `= string` so you can convert easily

pkg/common/namespace_test.go
migueltol22: Updated.

pkg/runtime/namespace_config_provider.go
migueltol22: This is already updated and basically doing the mapping. So if the value is "domain" in the config it will return a Domain enum value.

pkg/clusterresource/controller.go
migueltol22: Removed namespaceMapping variable.

pkg/common/namespace.go
migueltol22: The type is string so I could either have a mapping from string to enum in order to keep the us of an enum. I currently do this mapping in `namespace_config_provider.go`

pkg/runtime/namespace_config_provider.go
katrogan: oh I mean when I pull this locally this file doesn't compile

pkg/runtime/namespace_config_provider.go
```diff
@@ -0,0 +1,41 @@
...
+
+func (p *NamespaceMappingConfigurationProvider) GetNamespaceMappingConfig() common.NamespaceMapping {
+	var mapping string
+	if namespaceMappingConfig != nil && namespaceMappingConfig.GetConfig() != nil {
+		mapping = namespaceMappingConfig.GetConfig().(*interfaces.NamespaceMappingConfig).Mapping
+	}
+
+	switch mapping {
+	case domainVariable:
+		logger.Warningf(context.Background(), "Failed to find namespace mapping in config, defaulting to <project>-<domain>")
```
katrogan: is this log statement still accurate?

pkg/runtime/namespace_config_provider.go
```diff
@@ -0,0 +1,41 @@
...
+
+type NamespaceMappingConfigurationProvider struct{}
+
+func (p *NamespaceMappingConfigurationProvider) GetNamespaceMappingConfig() common.NamespaceMapping {
+	var mapping string
+	if namespaceMappingConfig != nil && namespaceMappingConfig.GetConfig() != nil {
+		mapping = namespaceMappingConfig.GetConfig().(*interfaces.NamespaceMappingConfig).Mapping
+	}
+
+	switch mapping {
```
katrogan: can we also add an option for project-domain too that doesn't log as an unsupported option?

pkg/runtime/namespace_config_provider.go
migueltol22: Nope. Removed.

pkg/runtime/namespace_config_provider.go
migueltol22: added project-domain option

pkg/common/namespace.go
katrogan: sure, sounds good

migueltol22:
@katrogan Can you give another look. Thanks!

pkg/common/namespace.go
```diff
@@ -0,0 +1,25 @@
...
+	Domain        NamespaceMapping = iota
+)
+
+// GetNamespaceName returns kubernetes namespace name
+func GetNamespaceName(mapping NamespaceMapping, project, domain string) string {
+	switch mapping {
+	case Domain:
+		return domain
+	case ProjectDomain:
+		return fmt.Sprintf(namespaceFormat, project, domain)
```
katrogan: nit: you can use `fallthrough` here to avoid duplicate code with the default statement

migueltol22: Done.

PR #5, opened 2019-09-27
Update the executions model

anandswaminathan:


pkg/repositories/config/migrations.go
```diff
@@ -109,4 +109,11 @@ var Migrations = []*gormigrate.Migration{
 			return tx.Exec("update node_executions set parent_task_execution_id = NULL where parent_task_execution_id = 0").Error
 		},
 	},
+	// Update executions table to add cluster
+	{
+		ID: "2019-09-27-executions",
```
katrogan: can you update the id in line 59 instead?

katrogan: very well, can you add a rollback command that drops the new column?

pkg/repositories/config/migrations.go
```diff
@@ -109,4 +109,14 @@ var Migrations = []*gormigrate.Migration{
...
+		Migrate: func(tx *gorm.DB) error {
+			return tx.AutoMigrate(&models.Execution{}).Error
+		},
+		Rollback: func(tx *gorm.DB) error {
+			return tx.AutoMigrate(&models.Execution{}).Error
```
wild-endeavor: @katrogan gorm deletes columns too?

katrogan: nope, this rollback function is identical to the migrate. i think you need to explicitly drop the new column.

PR #7, opened 2019-09-29
Remove extra return

wild-endeavor:
`make server` was not working


wild-endeavor:
ptal @anandswaminathan @katrogan

anandswaminathan:
+1

PR #9, opened 2019-10-02
Add index for ParentTaskExecutionID

katrogan:
Not worth adding a new migration for since no one else is using this currently.


anandswaminathan:
+1 (Guess migration is not needed, as nobody else has tables created)

PR #13, opened 2019-10-09
Adding support for project descriptions

schottra:
This updates all model/manager/validation/migration code related to `Project` to support the `description` field.

Description is an optional, variable length string which does not exceed 300 characters.

### DIFF
--- a/pkg/manager/impl/project_manager.go
+++ b/pkg/manager/impl/project_manager.go
@@ -27,7 +27,7 @@ func (m *ProjectManager) CreateProject(ctx context.Context, request admin.Projec
 	if err := validation.ValidateProjectRegisterRequest(request); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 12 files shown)


pkg/manager/impl/project_manager.go
```diff
@@ -28,7 +27,7 @@ func (m *ProjectManager) CreateProject(ctx context.Context, request admin.Projec
 	if err := validation.ValidateProjectRegisterRequest(request); err != nil {
 		return nil, err
 	}
-	projectModel := transformers.CreateProjectModel(request.Project.Id, request.Project.Name)
+	projectModel := transformers.CreateProjectModel(request.Project.Id, request.Project.Name, request.Project.Description)
```
katrogan: nit: might just be simpler to pass in a pointer to request.Project instead of decomposing nearly all the fields

schottra: That does make sense. I think I was blindly extending the existing pattern :-)

PR #15, opened 2019-10-14
Deprecate LiteralMap in ExecutionSpec and ExecutionClosure

matthewphsmith:
Due to the size of some literal maps, stowing it inside the closure and spec protobufs can result in very large messages from list endpoints. Also, it makes fetching the inputs of a workflow execution slightly different that for task and node executions.

This change offloads the literal map objects into Admin's blob store (i.e. s3 or similar) and stores a reference to it. It updates the GetExecutionData endpoints to match the behavior of node and task executions.

Additionally, it provides for a smooth migration. The only 'breaking' change is for users who rely on reading inputs of executions from the ListExecutions endpoint. We estimate this to impact no one. While clients are upgraded, we will continue to:

1. Accept inputs from old clients starting executions.
2. Return inputs _only_ in the spec returned from GetExecution, thus not breaking old clients that read those fields for inputs of a specific execution.


pkg/manager/impl/execution_manager.go
```diff
@@ -216,6 +231,16 @@ func (m *ExecutionManager) launchExecutionAndPrepareModel(
 	// Dynamically assign execution queues.
 	m.populateExecutionQueue(ctx, *workflow.Id, workflow.Closure.CompiledWorkflow)
 
+	inputsUri, err := m.offloadInputs(ctx, executionInputs, &workflowExecutionID, shared.Inputs)
```
anandswaminathan: Is this idempotent?

If launch in K8s fails, and user retries, this should be no-op right?

pkg/manager/impl/execution_manager.go
```diff
@@ -578,6 +630,23 @@ func (m *ExecutionManager) GetExecution(
 		return nil, transformerErr
 	}
 
+	// TODO: Remove the publishing to deprecated fields (Inputs) after a smooth migration has been completed of our existing users
```
anandswaminathan: Add comment - TO BE DELETED

pkg/manager/impl/execution_manager.go
anandswaminathan: Discussed offline, storageClient overwrites.

anandswaminathan:
Looks good to me. @matthewphsmith Would recommend directly deploying and testing this in staging before merging to master. (Also you might need a +1 from someone else after you push more tests - new github setting)

pkg/manager/impl/execution_manager.go
```diff
@@ -328,12 +358,28 @@ func (m *ExecutionManager) RelaunchExecution(
 	if executionSpec.Metadata == nil {
 		executionSpec.Metadata = &admin.ExecutionMetadata{}
 	}
+	var inputs *core.LiteralMap
+	if len(existingExecutionModel.UserInputsUri) > 0 {
+		inputs = new(core.LiteralMap)
```
chanadian: Interesting you used `new` here. Not a problem but generally in our codebase we have been doing `&core.LiteralMap{}`

pkg/manager/impl/execution_manager.go
```diff
@@ -328,12 +358,28 @@ func (m *ExecutionManager) RelaunchExecution(
...
+		if err := m.storageClient.ReadProtobuf(ctx, existingExecutionModel.InputsUri, inputs); err != nil {
+			return nil, err
+		}
+	} else {
+		// For old data, inputs are held in the spec
```
chanadian: In the rare case that an execution doesn't have inputs, would `spec.Inputs` would be nil or empty here? If nil that may be a problem?

pkg/manager/impl/execution_manager.go
matthewphsmith: Ah yeah, I wanted to ensure that golang realized to place it on the heap...although i think by observing the pointer escaping, golang will automatically allocate on the heap anyway.  So i can meet convention.

PR #16, opened 2019-10-14
Implementation of Weighted Clusters and domain sharding

anandswaminathan:
This PR implements the following:

- Support for adding weights for the clusters, to determine the number of executions to send to a particular Kubernetes cluster.
- Support for any cluster to accept requests for particular domains. Default is all domains.
- For a given execution, the logic first extracts all eligible clusters and then based on weight selects one. For a given executionID (name), the same cluster will be returned.

cc @katrogan @EngHabu


pkg/executioncluster/impl/cluster_execution_target_provider.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/lyft/flyteadmin/pkg/executioncluster"
+	"github.com/lyft/flyteadmin/pkg/flytek8s"
+	runtime "github.com/lyft/flyteadmin/pkg/runtime/interfaces"
+	flyteclient "github.com/lyft/flytepropeller/pkg/client/clientset/versioned"
+	"github.com/lyft/flytestdlib/promutils"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ClusterExecutionTargetProvider struct{}
```
EngHabu: docs on public types, please

pkg/executioncluster/impl/random_cluster_selector.go
```diff
@@ -0,0 +1,173 @@
...
+	if spec.TargetID != "" {
+		if val, ok := s.executionTargetMap[spec.TargetID]; ok {
+			return &val, nil
+		}
+		return nil, fmt.Errorf("invalid cluster target %s", spec.TargetID)
+	}
+	if spec.ExecutionID != nil {
+		if weightedRandomList, ok := s.domainWeightedRandomMap[spec.ExecutionID.GetDomain()]; ok {
+			executionName := spec.ExecutionID.GetName()
+			if executionName != "" {
```
EngHabu: Is it legit to have an execution ID with no name?

pkg/executioncluster/interfaces/cluster.go
```diff
@@ -0,0 +1,10 @@
+package interfaces
+
+import (
+	"github.com/lyft/flyteadmin/pkg/executioncluster"
+)
+
+type ClusterInterface interface {
```
EngHabu: docs plz

pkg/executioncluster/impl/random_cluster_selector.go
anandswaminathan: Not in general, but someone can use `random_cluster_selector.go` without passing in executionID

PR #17, opened 2019-10-18
Support signed URL's from minio.

kumare3:
Currently we return data from the Admin server as signed URLs. Flyte in its sandbox environment uses minio. Minio support the AWS SDK completely, but needs a couple more settings. support for signed url existed for AWS S3, this change adds it for Minio.

NOTE: Other cloud providers are not really supported. They return a plain url. 
TODO: migrate this API to flytestdlib and normalize across all cloud providers. We should have an alternate api that could return the literalmap - decoded (and it could use the size as a limiting factor)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -18,12 +18,13 @@
   version = "v10.2.1-beta"
 
 [[projects]]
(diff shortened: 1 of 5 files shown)


jonathanburns:
+1

PR #20, opened 2019-10-24
Adds endpoints for NamedEntities

schottra:
This adds endpoints for working with `NamedEntity` objects. These are an extension of the `NamedEntityIdentifier` concept, which is a combination of project/domain/name/resource type. The formal entity also includes an attached `Metadata` field, which for now only includes a description.

Note that a `NamedEntity` is really a combination of data from multiple tables. The project/domain/name are derived from a table based on resource type (workflow/launch plan/ task), while the metadata fields are taken from a new table `named_entity_metadata`.

* Added a new table `named_entity_metadata` and a migration for it
* Added RPC handlers for `GET`, `PUT`, and a list endpoint for named entities
* Added a manger class
* Added repository and model classes
* Added transformers, validators
* Refactored/extended some common code related to tests and GORM queries
* Unit tests for most of the above functionality


pkg/rpc/adminservice/named_entity.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/lyft/flyteadmin/pkg/rpc/adminservice/util"
+	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/admin"
+	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/core"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func (m *AdminService) GetNamedEntity(ctx context.Context, request *admin.NamedEntityGetRequest) (*admin.NamedEntity, error) {
+	defer m.interceptPanic(ctx, request)
```
katrogan: this shouldn't be necessary, we start the admin server with a panic recovery wrapper: https://github.com/lyft/flyteadmin/blob/master/pkg/rpc/adminservice/base.go#L60

pkg/rpc/adminservice/named_entity.go
```diff
@@ -0,0 +1,78 @@
...
+	if request == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Incorrect request, nil requests not allowed")
+	}
+
+	// We require a valid resource type to retrieve metadata
+	if request.ResourceType == core.ResourceType_UNSPECIFIED {
```
katrogan: can you move this into the validation in the manager method to be consistent with other request validation? the only reason other adminservice/ wrappers inject the ResourceType for other requests is because the url references the entity type even though the request body doesn't specify the ResourceType

pkg/manager/impl/validation/validation.go
```diff
@@ -121,6 +145,22 @@ func ValidateNamedEntityIdentifierListRequest(request admin.NamedEntityIdentifie
...
+	if err := ValidateEmptyStringField(request.Project, shared.Project); err != nil {
+		return err
+	}
+	if err := ValidateEmptyStringField(request.Domain, shared.Domain); err != nil {
+		return err
+	}
+	if err := ValidateResourceType(request.ResourceType); err != nil {
+		return err
+	}
+	if request.Limit == 0 {
```
katrogan: nit: you can use ValidateLimit https://github.com/lyft/flyteadmin/blob/master/pkg/manager/impl/validation/validation.go#L184

pkg/repositories/gormimpl/common.go
```diff
@@ -114,3 +142,39 @@ func applyScopedFilters(tx *gorm.DB, inlineFilters []common.InlineFilter, mapFil
 	}
 	return tx, nil
 }
+
+func getGroupByForNamedEntity(tableName string) string {
```
katrogan: do these methods need to belong in common if they're only used in one place?

Gopkg.toml
```diff
@@ -56,7 +56,8 @@
 [[override]]
   name = "github.com/lyft/flyteidl"
   source = "https://github.com/lyft/flyteidl"
-  version = "^0.16.x"
+  # version = "^0.17.x"
+  revision = "4d2184dc1727f86cb17d8d711fa135695aa16744"
```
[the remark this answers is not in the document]
schottra: Ah. Yeah v0.16.1 adds a new endpoint that wasn't implemented in master yet. So I was pinned here so I could build :-D

pkg/manager/impl/named_entity_manager.go
```diff
@@ -0,0 +1,130 @@
...
+	config runtimeInterfaces.Configuration,
+	scope promutils.Scope) interfaces.NamedEntityInterface {
+
+	metrics := NamedEntityMetrics{
+		Scope: scope,
+	}
+	return &NamedEntityManager{
+		db:      db,
+		config:  config,
+		metrics: metrics,
```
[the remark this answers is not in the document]
schottra: There are some at the RPC and Repo levels, just not in the manager. I wasn't sure what to measure at this layer 😬

pkg/manager/impl/named_entity_manager.go
```diff
@@ -0,0 +1,130 @@
...
+func (m *NamedEntityManager) UpdateNamedEntity(ctx context.Context, request admin.NamedEntityUpdateRequest) (
+	*admin.NamedEntityUpdateResponse, error) {
+	if err := validation.ValidateResourceType(request.ResourceType); err != nil {
+		return nil, err
+	}
+	if err := validation.ValidateNamedEntityIdentifier(request.Id); err != nil {
+		return nil, err
+	}
+
+	// TODO: Validate that the provided project/domain/name exists in the table
```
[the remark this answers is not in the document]
schottra: Ah, neat. That would at least get me the project/domain. But the name would require a lookup.
For the GET case I'm effectively validating it at the Repo level by returning a NotFound for an invalid combination of project/domain/name. But I couldn't come up with an easy way to do that in one query for the UPDATE case :-/

pkg/repositories/gormimpl/named_entity_repo.go
```diff
@@ -0,0 +1,131 @@
...
+type NamedEntityRepo struct {
+	db               *gorm.DB
+	errorTransformer errors.ErrorTransformer
+	metrics          gormMetrics
+}
+
+func (r *NamedEntityRepo) Update(ctx context.Context, input models.NamedEntity) error {
+	timer := r.metrics.UpdateDuration.Start()
+	var metadata models.NamedEntityMetadata
+	tx := r.db.Debug().Where(&models.NamedEntityMetadata{
```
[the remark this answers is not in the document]
schottra: It causes GORM to spit out the generated query and results. I meant to take it out, thanks!

pkg/repositories/gormimpl/named_entity_repo.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return nil
+}
+
+func (r *NamedEntityRepo) Get(ctx context.Context, input interfaces.GetNamedEntityInput) (models.NamedEntity, error) {
+	var namedEntity models.NamedEntity
+
+	// This is implemented as a filter+join to ensure that when the provided
+	// resource_type/project/domain/name combination don't map to a valid
+	// entry in the target table, we return NotFound instead of a fake record
```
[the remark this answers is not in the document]
schottra: Misleading comment!
In the case that we don't find any metadata, we have enough information in the request to construct a response that includes the identifier and an empty metadata object. I originally had it implemented that way and realized it was a weird pattern. I'm not sure why I needed a comment to explain how I _didn't_ end up doing it... 😁

pkg/repositories/models/named_entity.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/core"
+)
+
+// IMPORTANT: If you update the model below, be sure to double check model definitions in
+// pkg/repositories/config/migration_models.go
+
+// NamedEntityMetadata primary key
+type NamedEntityMetadataKey struct {
+	ResourceType core.ResourceType `gorm:"primary_key;named_entity_metadata_type_project_domain_name_idx"`
```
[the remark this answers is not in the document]
schottra: I was following the pattern set in other model classes. I'm not sure if it's necessary or not!

pkg/repositories/models/named_entity.go
```diff
@@ -0,0 +1,40 @@
...
+	Description string `gorm:"type:varchar(300)"`
+}
+
+// Database model to encapsulate metadata associated with a NamedEntity
+type NamedEntityMetadata struct {
+	BaseModel
+	NamedEntityMetadataKey
+	NamedEntityMetadataFields
+}
+
```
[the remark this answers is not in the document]
schottra: Done

pkg/repositories/models/named_entity.go
```diff
@@ -0,0 +1,40 @@
...
+type NamedEntityKey struct {
+	ResourceType core.ResourceType
+	Project      string
+	Domain       string
+	Name         string
+}
+
+type NamedEntity struct {
```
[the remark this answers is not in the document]
schottra: NamedEntityMetadata is the model used to do lookups on the metadata table. NamedEntity is the model type returned by the Repository layer. I organized it that way because we may compose more fields into a NamedEntity that are derived from other tables.
(NamedEntity) isn't an actual model in our database the way Workflow/Task are.

pkg/rpc/adminservice/named_entity.go
schottra: I can remove it. I was following the pattern in other RPC handlers.

pkg/rpc/adminservice/named_entity.go
schottra: Sure. Another case of following/copying code from another handler

schottra: Oh actually, the manager already validates the resource type. I'm just going to remove this.

pkg/repositories/gormimpl/common.go
schottra: Maybe not. Seemed like all of the joins/groups were in this file either way. And Go's package mechanism will expose them at package scope no matter which file they live in. So it seemed like a good idea to keep this near the other Db helper functions.

pkg/manager/impl/util/shared.go
```diff
@@ -117,6 +114,30 @@ func GetLaunchPlan(
 	return transformers.FromLaunchPlanModel(launchPlanModel)
 }
 
+func GetNamedEntityModel(
```
katrogan: nit: maybe add a test for these new methods?

pkg/manager/impl/validation/validation.go
```diff
@@ -121,6 +145,42 @@ func ValidateNamedEntityIdentifierListRequest(request admin.NamedEntityIdentifie
 	return nil
 }
 
+func ValidateNamedEntityGetRequest(request admin.NamedEntityGetRequest) error {
```
katrogan: in following with other specific validations, can you put this in its own file e.g. https://github.com/lyft/flyteadmin/blob/master/pkg/manager/impl/validation/execution_validator.go

pkg/manager/impl/validation/validation.go
```diff
@@ -59,6 +59,30 @@ func ValidateIdentifier(id *core.Identifier, expectedType common.Entity) error {
 	return nil
 }
 
+// Validates that all required fields for an identifier are present.
+func ValidateNamedEntityIdentifier(id *admin.NamedEntityIdentifier) error {
```
katrogan: Please add tests for all these new methods!

pkg/repositories/gormimpl/common.go
katrogan: sure but i mean logically do they need to belong in common? i'd say if they're specific to one repo impl it makes sense for them to live in that corresponding file

pkg/repositories/gormimpl/common.go
```diff
@@ -114,3 +142,39 @@ func applyScopedFilters(tx *gorm.DB, inlineFilters []common.InlineFilter, mapFil
...
+		fmt.Sprintf("%s.%s", tableName, Name),
+		fmt.Sprintf("%s.%s", namedEntityMetadataTableName, Description),
+	}
+}
+
+func getNamedEntityFilters(resourceType core.ResourceType, project string, domain string, name string) ([]common.InlineFilter, error) {
+	entity := common.ResourceTypeToEntity[resourceType]
+
+	filters := make([]common.InlineFilter, 0)
+	projectFilter, err := common.NewSingleValueFilter(entity, common.Equal, Project, project)
```
katrogan: btw you can use https://github.com/lyft/flyteadmin/blob/master/pkg/manager/impl/util/filters.go#L146 for these

pkg/repositories/gormimpl/common.go
```diff
@@ -60,6 +62,32 @@ var leftJoinTaskToTaskExec = fmt.Sprintf(
 	taskTableName, taskExecutionTableName, taskTableName, taskExecutionTableName, taskTableName,
 	taskExecutionTableName, taskTableName, taskExecutionTableName, taskTableName)
 
+// Creates a JOIN clause between a table which can hold NamedEntityIdentifiers
+// and the metadata associated between them. Metadata is optional, so a left
+// join is used.
+func CreateEntityMetadataJoin(entityTableName string, resourceType core.ResourceType) string {
```
katrogan: Please add tests for all these new methods too!

pkg/repositories/models/named_entity.go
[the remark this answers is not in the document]
katrogan: oh haha 🙂 
you can actually run the migration locally with your sandbox postgres deployment if you want to verify

pkg/repositories/models/named_entity.go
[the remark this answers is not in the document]
katrogan: coolio (and the code comments are very appreciated!)

pkg/rpc/adminservice/named_entity.go
katrogan: oops, disregard then!

pkg/repositories/gormimpl/common.go
schottra: Makes sense. I'll move them!

pkg/manager/impl/validation/validation.go
schottra: Done!

pkg/manager/impl/validation/validation.go
schottra: Yep, done.

pkg/repositories/gormimpl/common.go
schottra: So, I was trying to maintain package separation. It seemed to me like the idea is that Manager can reach down and use methods from Repo. But we had no instances of Repo reaching *up* to use things from Manager. I'm not sure how Go handles circular dependencies, but that would create one. I only really saw the Repo using methods from the common package.
But it also seemed overkill/unnecessary to try to move all of those things out of the Manager layer when they seem pretty clearly specific to it.

pkg/repositories/gormimpl/common.go
schottra: I'm not certain a test is necessary for this one. It's only used to generate three string constants that are used in queries which are already covered under testing in the NamedEntityRepo class. I actually might just remove the helper function and make the three constants stand on their own

schottra: But I will add some for the other methods!

tests/named_entity_test.go
```diff
@@ -0,0 +1,300 @@
+// +build integration
```
katrogan: thank you for adding these tests! looks great

tests/shared.go
```diff
@@ -3,4 +3,139 @@
...
+
+	"github.com/lyft/flyteadmin/pkg/manager/impl/testutils"
+	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/admin"
+	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/service"
+	"github.com/stretchr/testify/assert"
+)
+
 const project, domain, name = "project", "domain", "execution name"
+
+var entityProjects = []string{"admintests"}
```
katrogan: thanks for refactoring all these

pkg/repositories/gormimpl/named_entity_repo.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/lyft/flyteadmin/pkg/repositories/errors"
+	"github.com/lyft/flyteadmin/pkg/repositories/interfaces"
+	"github.com/lyft/flyteadmin/pkg/repositories/models"
+	"github.com/lyft/flytestdlib/promutils"
+)
+
+// Creates a JOIN clause between a table which can hold NamedEntityIdentifiers
+// and the metadata associated between them. Metadata is optional, so a left
+// join is used.
+func CreateEntityMetadataJoin(entityTableName string, resourceType core.ResourceType) string {
```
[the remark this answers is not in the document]
schottra: Oh, no it's not. I updated the statements below to not use it and forgot to delete it. Good catch.

pkg/manager/impl/named_entity_manager_test.go
```diff
@@ -0,0 +1,136 @@
...
+	})
+	assert.Error(t, err)
+	assert.Nil(t, response)
+}
+
+func TestNamedEntityManager_Update(t *testing.T) {
+	repository := getMockRepositoryForNETest()
+	manager := NewNamedEntityManager(repository, getMockConfigForNETest(), mockScope.NewTestScope())
+	updatedDescription := "updated description"
+
```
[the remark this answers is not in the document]
schottra: Good idea!

pkg/repositories/interfaces/named_entity_repo.go
```diff
@@ -0,0 +1,31 @@
...
+	Entities []models.NamedEntity
+}
+
+// Defines the interface for interacting with NamedEntity models
+type NamedEntityRepoInterface interface {
+	// Returns NamedEntity objects matching the provided query. A limit is
+	// required
+	List(ctx context.Context, resourceType core.ResourceType, input ListResourceInput) (NamedEntityCollectionOutput, error)
+	// Updates metadata associated with a NamedEntity
+	Update(ctx context.Context, input models.NamedEntity) error
```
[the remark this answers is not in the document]
schottra: I added a comment. It's meant to Update the entire NamedEntity model (which may involve multiple tables at some point). Right now, it's only setting the metadata (and creating if it exists).

PR #21, opened 2019-10-29
Support postgres extra options [gcp]

tnsetting:
Right now it is enforced to use sslmode when flyte admin is connecting to postgres and password provided. The PR make it possible to connect to postgres when disabling sslmode.


tnsetting:
> Thank you @tnsetting :)

@kumare3 I am not sure why the integration test failed for the error`error: flag docker-username is required`.

PR #23, opened 2019-10-29
Upgrade flytestdlib and golang version

katrogan:
Also temporarily remove integration tests until the story around dockerhub integration if fully fleshed out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@
 
 vendor/
 node_modules/
+
+.virtualgo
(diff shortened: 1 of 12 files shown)


wild-endeavor:
boilerplate was updated just by running the update script right?

katrogan:
> boilerplate was updated just by running the update script right?
@wild-endeavor no - we'll update boilerplate once everything's been migrated to go 1.13

PR #24, opened 2019-10-30
Add More Template Strings for Email Notifications

matthewphsmith:
Added support for inserting the following values into the notification upon the failure of the workflow:

1. Workflow Name
2. Workflow Project
3. Workflow Domain
4. Workflow Version
5. Launch Plan Name
6. Launch Plan Project
7. Launch Plan Domain
8. Launch Plan Version

### DIFF
--- a/pkg/async/notifications/email.go
+++ b/pkg/async/notifications/email.go
@@ -9,9 +9,12 @@ import (
 	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/admin"
 )
 
+type GetTemplateValue func(admin.WorkflowExecutionEventRequest, *admin.Execution) string
+
 const executionError = " The execution failed with error: [%s]."
 
 const substitutionParam = "{{ %s }}"
+const substitutionParamNoSpaces = "{{%s}}"
 const project = "project"
 const domain = "domain"
 const name = "name"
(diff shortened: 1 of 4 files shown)


pkg/async/notifications/email.go
```diff
@@ -18,14 +17,30 @@ const domain = "domain"
...
+const launchPlanName = "launch_plan.name"
+const launchPlanVersion = "launch_plan.version"
 const replaceAllInstances = -1
 
-func substituteEmailParameters(message string, request admin.WorkflowExecutionEventRequest, execution models.Execution) string {
-	response := strings.Replace(message, fmt.Sprintf(substitutionParam, project), execution.Project, replaceAllInstances)
-	response = strings.Replace(response, fmt.Sprintf(substitutionParam, domain), execution.Domain, replaceAllInstances)
-	response = strings.Replace(response, fmt.Sprintf(substitutionParam, name), execution.Name, replaceAllInstances)
+func substituteEmailParameters(message string, request admin.WorkflowExecutionEventRequest, execution *admin.Execution) string {
+	response := strings.Replace(message, fmt.Sprintf(substitutionParam, project), execution.Id.Project, replaceAllInstances)
```
katrogan: nit, this is kind of verbose and I'm worried error prone because we have to keep these Replace calls in sync with the constants above. can we put all the placeholder string constants in an array defined at the top of the file and then iterate through the array and perform the string substitution in a loop?

PR #27, opened 2019-10-31
Support postgres extra options [gcp]

tnsetting:
Right now it is enforced to use sslmode when flyte admin is connecting to postgres and password provided. The PR make it possible to connect to postgres when disabling sslmode.

### DIFF
--- a/pkg/repositories/config/postgres.go
+++ b/pkg/repositories/config/postgres.go
@@ -53,8 +53,8 @@ func (p *PostgresConfigProvider) GetArgs() string {
 		return fmt.Sprintf("host=%s port=%d dbname=%s user=%s sslmode=disable",
 			p.config.Host, p.config.Port, p.config.DbName, p.config.User)
 	}
-	return fmt.Sprintf("host=%s port=%d dbname=%s user=%s password=%s",
-		p.config.Host, p.config.Port, p.config.DbName, p.config.User, p.config.Password)
+	return fmt.Sprintf("host=%s port=%d dbname=%s user=%s password=%s %s",
+		p.config.Host, p.config.Port, p.config.DbName, p.config.User, p.config.Password, p.config.ExtraOptions)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/repositories/config/postgres.go
```diff
@@ -53,8 +53,8 @@ func (p *PostgresConfigProvider) GetArgs() string {
 		return fmt.Sprintf("host=%s port=%d dbname=%s user=%s sslmode=disable",
 			p.config.Host, p.config.Port, p.config.DbName, p.config.User)
 	}
-	return fmt.Sprintf("host=%s port=%d dbname=%s user=%s password=%s",
-		p.config.Host, p.config.Port, p.config.DbName, p.config.User, p.config.Password)
+	return fmt.Sprintf("host=%s port=%d dbname=%s user=%s password=%s %s",
+		p.config.Host, p.config.Port, p.config.DbName, p.config.User, p.config.Password, p.config.ExtraOptions)
```
wild-endeavor: @katrogan will this break our production deployment?  We do have `options: sslmode=disable` in our yaml settings I think.

Also, post IAM role integration, the password empty check is no longer a proxy for sandbox deployment.

wild-endeavor:
Any chance you could also update the `postgres_test.go` file please?  Thank you.

pkg/repositories/config/postgres.go
honnix: I'm a bit confused. How does [this](https://github.com/lyft/flyte/blob/master/kustomize/overlays/sandbox/admindeployment/flyteadmin_config.yaml#L21) work today without this PR?

tnsetting: If password is provided, it will discard the option and always use sslmode. But cloudsql proxy does not use sslmode

honnix: I mean I couldn't find anywhere in the code reading `ExtraOptions`. I guess I'm missing something.

PR #30, opened 2019-11-07
Fix invalid filter function error message

slai:
Currently, if you use an invalid filter function, e.g. `neq(phase,RUNNING)`, the error message (via the GRPC gateway) is -

```
{"error":"unsupported filter expression: contains","code":3,"message":"unsupported filter expression: contains"}
```

This is confusing as it names a function that isn't actually the one that's invalid. This happens because the message takes a FilterExpression value, but when the function is invalid, the filter expression value is its zero-value, `0`, which happens to correspond to the value for `Contains`.

This PR fixes this by clarifying the message and passes the function string directly.

### DIFF
--- a/pkg/common/filters.go
+++ b/pkg/common/filters.go
@@ -80,6 +80,7 @@ var executionIdentifierFields = map[string]bool{
 	"name":    true,
 }
 
+const unrecognizedFilterFunction = "unrecognized filter function: %s"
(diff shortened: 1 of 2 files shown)


katrogan:
fyi to fix the lint errors you can run the lint target locally: https://github.com/lyft/flyteadmin#before-check-in

PR #33, opened 2019-11-12
Metadata url and base url as items in the auth context

wild-endeavor:
Was incorrectly using `path.Join` in order to join URL paths.  This would've failed on windows, and also the redirect for the metadata and user info endpoints ended up being like `http:/blah.okta.com/...` instead of having the correct `//`.

Added a couple functions to the context interface to get these.
Added more todo's around config object to implement the actual use of the discovery endpoints. This will get rid of a lot of the URLs in the config.

### DIFF
--- a/pkg/auth/handlers_test.go
+++ b/pkg/auth/handlers_test.go
@@ -69,6 +69,6 @@ func TestGetMetadataEndpointRedirectHandler(t *testing.T) {
 	assert.NoError(t, err)
 	w := httptest.NewRecorder()
 	handler(w, req)
-	assert.Equal(t, 303, w.Code)
+	assert.Equal(t, http.StatusSeeOther, w.Code)
 	assert.Equal(t, "http://www.google.com/.well-known/oauth-authorization-server", w.Header()["Location"][0])
 }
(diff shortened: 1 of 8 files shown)


wild-endeavor:
ptal @EngHabu @katrogan

pkg/auth/handlers_test.go
```diff
@@ -53,3 +54,21 @@ func TestGetHttpRequestCookieToMetadataHandler(t *testing.T) {
...
+	assert.NoError(t, err)
+	mockAuthCtx := mocks.AuthenticationContext{}
+	mockAuthCtx.On("GetBaseURL").Return(baseURL)
+	mockAuthCtx.On("GetMetadataURL").Return(metadataPath)
+	handler := GetMetadataEndpointRedirectHandler(ctx, &mockAuthCtx)
+	req, err := http.NewRequest("GET", "/xyz", nil)
+	assert.NoError(t, err)
+	w := httptest.NewRecorder()
+	handler(w, req)
+	assert.Equal(t, 303, w.Code)
```
katrogan: nit: use the named constant instead? https://golang.org/pkg/net/http/#pkg-constants

wild-endeavor:
thanks

PR #34, opened 2019-11-20
add server reflection

honnix:
The same as https://github.com/lyft/datacatalog/pull/19

With [server reflection](https://github.com/grpc/grpc-go/blob/master/Documentation/server-reflection-tutorial.md), it would be easier to interact with GRPC service using e.g. `grpcurl`.

flyteadmin does come with HTTP endpoint as well, but interacting with GRPC directly makes it more handy.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -921,7 +921,7 @@
   revision = "919d9bdd9fe6f1a5dd95ce5d5e4cdb8fd3c516d0"
 
 [[projects]]
-  digest = "1:0c0ea0f0ccec8a112245dd3ca0c3b304224ede9afd4b7a2b482a7b48471ef894"
+  digest = "1:ca45e0a856db4fdac5d23492fed11be07e5c86c62a26ed0bbd63f89fbb23b6a4"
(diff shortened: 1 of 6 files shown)


honnix:
@kumare3 PTAL

EngHabu:
CC @wild-endeavor, @katrogan 

@honnix can you add this behind a config flag? I'm not sure if it has implications on performance/security... etc.

katrogan:
+1 to making configurable

honnix:
This seems like a standard as part of GRCP protocol and I didn't find any thing around performance hit and security. But sure, we can have a config flag.

BTW I did the same to datacatalog and it was already merged. I will send a follow up PR to make it configurable.

honnix:
https://github.com/lyft/datacatalog/pull/20

wild-endeavor:
sorry for the back and forth - could you add grpcServerReflection: true to the config file at the base of the repo please?  It's meant to be the config that one would use when running locally (`make server`) and it'd be nice to have it turned on.

Thanks!

PR #37, opened 2019-12-04
Fix bugs encountered with Auth flow

schottra:
We weren't setting CORS headers for the `/me` endpoint. In attempting to implement that, I found myself duplicating some code and went in search for a better way to handle CORS.

I removed the custom CORS logic and replaced it with a simpler middleware-based implementation from `gorilla`. It still pulls in the allowed headers and origins lists, but implements the origin matching in a way that is compatible with credentials _and_ specifying multiple domains.

This will also apply CORS to all future endpoints we add, since it's done at the level above the mux.


Gopkg.toml
```diff
@@ -41,6 +41,10 @@
   name = "github.com/golang/protobuf"
   version = "1.1.0"
 
+[[constraint]]
```
wild-endeavor: why do we need this specific version?

schottra: I guess we don't, but it's the latest, and also the version that was already included by one of our dependencies. We also seem to be pinning the rest of our dependencies to specific versions, so it seemed consistent. 🤷‍♂

cmd/entrypoints/serve.go
```diff
@@ -216,7 +198,20 @@ func serveGatewayInsecure(ctx context.Context, cfg *config.ServerConfig) error {
 	if err != nil {
 		return err
 	}
-	err = http.ListenAndServe(cfg.GetHostAddress(), httpServer)
+
+	var handler http.Handler
+	if cfg.Security.AllowCors {
+		handler = handlers.CORS(
+			handlers.AllowCredentials(),
+			handlers.AllowedHeaders(cfg.Security.AllowedHeaders),
```
katrogan: how does this work if admin is misconfigured and either the allowed headers or origins happens to be empty?

Gopkg.toml
schottra: Sure I'll pin it to the latest major with a caret 👍

cmd/entrypoints/serve.go
```diff
@@ -142,31 +144,11 @@ func newHTTPServer(ctx context.Context, cfg *config.ServerConfig, authContext in
...
-		}
 	}
 
 	// Create the grpc-gateway server with the options specified
 	gwmux := runtime.NewServeMux(gwmuxOptions...)
 
-	// Enable CORS if necessary on the gateway mux by adding a handler for all OPTIONS requests.
-	if cfg.Security.AllowCors {
-		globPattern := server.GetGlobPattern()
-		decorator := auth.GetCorsDecorator(ctx, cfg.Security.AllowedOrigins, cfg.Security.AllowedHeaders)
```
wild-endeavor: how does it handle OPTIONS calls if this bit is removed?

schottra: The Cors middleware wraps the mux and processes all requests, passing them through to the mux unless there's an error of some sort (disallowed origin, etc). It will handle all OPTIONS requests and add the appropriate headers based on the configuration.

cmd/entrypoints/serve.go
schottra: I tested this. It will still work (I can cURL from Postman). But it will make it not possible for the UI to make valid CORS-enabled calls.
The domain needs to be user-configured no matter what. But I'm considering changing it so that the content-type header is always sent. We need it for the UI in all cases.

PR #41, opened 2019-12-11
Use StrategicMergePatch when updating existing cluster resources

katrogan:
Kubectl defaults to using the [StrategicMergePatch](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md) strategy. However the controller-runtime only has an [implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/patch.go) for [MergePatch](https://tools.ietf.org/html/rfc7386) which we were formerly using.

This change introduces a strategicMergeFromPatch type (which interestingly [doesn't exist](https://github.com/kubernetes-sigs/controller-runtime/search?q=StrategicMergeFrom&unscoped_q=StrategicMergeFrom) in controller-runtime) to mimic the behavior of kubectl so we can successfully patch serviceaccounts.

Tested in flyte sandbox locally


katrogan:
ping @EngHabu @wild-endeavor

pkg/clusterresource/controller.go
```diff
@@ -285,18 +283,24 @@ func (c *controller) syncNamespace(ctx context.Context, namespace NamespaceName,
...
+				k8sObj.GetObjectKind().GroupVersionKind().Kind, k8sObj, target.ID, namespace)
 			err = target.Client.Create(ctx, k8sObjCopy)
 			if err != nil {
 				if k8serrors.IsAlreadyExists(err) {
 					logger.Debugf(ctx, "Resource [%+v] in namespace [%s] already exists - attempting update instead",
 						k8sObj.GetObjectKind().GroupVersionKind().Kind, namespace)
 					c.metrics.AppliedTemplateExists.Inc()
-					err = target.Client.Patch(ctx, k8sObjCopy, client.MergeFrom(k8sObjCopy))
+					// Use a strategic-merge-patch to mimic `kubectl apply` behavior.
+					err = target.Client.Patch(ctx, k8sObjCopy, StrategicMergeFrom(k8sObjCopy))
```
wild-endeavor: Can you copy/paste maybe the PR description here as well?  And also explain the issue we were seeing before?

PR #42, opened 2019-12-12
Auth Logout Handler

wild-endeavor:
This adds a logout handler for the auth component.  It will clear the two encrypted cookies used in the oauth flow by setting the expiration to something in the past.

The endpoint also takes the same `redirect_url` query parameter.
To use:
`http://localhost:8088/logout?redirect_url=http://www.google.com`

If there is no redirect url, it just returns a 200 with an empty body.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ linux_compile:
 
 .PHONY: server
 server:
-	go run cmd/main.go --logtostderr --server.kube-config ~/.kube/config  --config ~/flyteadmin_config.yaml serve
+	go run cmd/main.go --logtostderr --server.kube-config ~/.kube/config  --config flyteadmin_config.yaml serve
 
 .PHONY: migrate
 migrate:
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ linux_compile:
 
 .PHONY: server
 server:
-	go run cmd/main.go --logtostderr --server.kube-config ~/.kube/config  --config flyteadmin_config.yaml serve
+	go run cmd/main.go --logtostderr --server.kube-config ~/.kube/config  --config ~/flyteadmin_config.yaml serve
 
 .PHONY: migrate
 migrate:
--- a/pkg/auth/cookie_manager.go
+++ b/pkg/auth/cookie_manager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"net/http"
+	"time"
 
 	"github.com/lyft/flytestdlib/errors"
 	"github.com/lyft/flytestdlib/logger"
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -25,7 +25,7 @@ linux_compile:
 
 .PHONY: server
 server:
-	go run cmd/main.go --logtostderr --server.kube-config ~/.kube/config  --config flyteadmin_config.yaml serve
+	go run cmd/main.go --logtostderr --server.kube-config ~/.kube/config  --config ~/flyteadmin_config.yaml serve
```
katrogan: is this change intentional?

pkg/auth/cookie_manager.go
```diff
@@ -93,3 +94,28 @@ func (c CookieManager) SetTokenCookies(ctx context.Context, writer http.Response
 	}
 	return nil
 }
+
+func getLogoutAccessCookie() *http.Cookie {
+	return &http.Cookie{
+		Name:     accessTokenCookieName,
+		Value:    "",
```
katrogan: do you need to initialize these zero values?

Makefile
wild-endeavor: no no, just debugging.

pkg/auth/cookie_manager.go
wild-endeavor: no, but may as well make this one explicit.  just because the value is set when you log in, I feel like I like explicitly unsetting it, but that's the irrational part of me talking.

katrogan: okie dokes

pkg/auth/handlers.go
```diff
@@ -252,15 +251,18 @@ func GetMetadataEndpointRedirectHandler(ctx context.Context, authCtx interfaces.
 	}
 }
 
-// These are here for CORS handling. Actual serving of the OPTIONS request will be done by the gorilla/handlers package
-type CorsHandlerDecorator func(http.Handler) http.Handler
+func GetLogoutEndpointHandler(ctx context.Context, authCtx interfaces.AuthenticationContext) http.HandlerFunc {
+	return func(writer http.ResponseWriter, request *http.Request) {
+		logger.Debugf(ctx, "Deleting auth cookies")
```
schottra: Just out of curiosity, are we in the habit of leaving in debug statements? They don't blog our staging/production logs right?

pkg/auth/handlers.go
```diff
@@ -252,15 +251,18 @@ func GetMetadataEndpointRedirectHandler(ctx context.Context, authCtx interfaces.
...
+		// Redirect if one was given
+		queryParams := request.URL.Query()
+		if redirectURL := queryParams.Get(RedirectURLParameter); redirectURL != "" {
+			http.Redirect(writer, request, redirectURL, http.StatusTemporaryRedirect)
+		}
+	}
 }
+
+// These are here for CORS handling. Actual serving of the OPTIONS request will be done by the gorilla/handlers package
+type CorsHandlerDecorator func(http.Handler) http.Handler
```
schottra: Is this still needed?

pkg/rpc/adminservice/base.go
```diff
@@ -57,7 +57,7 @@ func NewAdminServer(kubeConfig, master string) *AdminService {
 	defer func() {
 		if err := recover(); err != nil {
 			adminScope.MustNewCounter("initialization_panic",
-				"panics encountered initializating the admin service").Inc()
```
schottra: I like this word better. :-p

pkg/auth/handlers.go
wild-endeavor: It's not but I'm gonna remove it later.  Don't want to get more approvals.  Already dismissed this approval enough times

pkg/auth/handlers.go
wild-endeavor: I am yeah, it helps with debugging.  The logging level is controlled through a runtime config so we can play with it on the fly.

PR #43, opened 2019-12-13
Add auth enforcement config options

wild-endeavor:
This is the last bit of auth work on the Admin side of things.  Before this PR, if you present Admin with a bad or expired token, you will get a 401.  However, if you present no token whatsoever, Admin will let the request go through.  This PR begins enforcement.

To help with migration, we are incrementally going to require auth.  We were originally going to use the gRPC endpoint name to decide whether or not to enforce things, but ultimately it made more sense to divide up by protocol - HTTP vs gRPC. Since most programmatic (more problematic) traffic goes through gRPC and only Flyte Console uses HTTP, we can start by just doing HTTP.

In order to do this, the code tags incoming HTTP requests because the token validation code is further downstream.

Also
* Adding the bit of code to look for bearer tokens in http headers as well as cookies to support use-cases where we use the HTTP Admin API.  Note however, that Admin will be deployed with a custom config so that the header you should send is `Flyte-Authorization` not the default `Authorization`, so the complete header would look something like `Flyte-Authorization: Bearer j.w.t`
(CC @haoyuez )

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ linux_compile:
 
 .PHONY: server
 server:
-	go run cmd/main.go --logtostderr --server.kube-config ~/.kube/config  --config ~/flyteadmin_config.yaml serve
+	go run cmd/main.go --logtostderr --server.kube-config ~/.kube/config  --config flyteadmin_config.yaml serve
 
 .PHONY: migrate
 migrate:
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ linux_compile:
 
 .PHONY: server
 server:
-	go run cmd/main.go --logtostderr --server.kube-config ~/.kube/config  --config flyteadmin_config.yaml serve
+	go run cmd/main.go --logtostderr --server.kube-config ~/.kube/config  --config ~/flyteadmin_config.yaml serve
 
 .PHONY: migrate
 migrate:
--- a/pkg/auth/config/config.go
+++ b/pkg/auth/config/config.go
@@ -56,8 +56,8 @@ type OAuthOptions struct {
 	// dimension that made the most sense to cut by at time of writing is HTTP vs gRPC as the web UI mainly used HTTP
 	// and the backend used mostly gRPC.  Cutting by individual endpoints is another option but it possibly falls more
 	// into the realm of authorization rather than authentication.
-	EnforceHttp bool `json:"enforceHttp"`
-	EnforceGrpc bool `json:"enforceGrpc"`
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -25,7 +25,7 @@ linux_compile:
 
 .PHONY: server
 server:
-	go run cmd/main.go --logtostderr --server.kube-config ~/.kube/config  --config flyteadmin_config.yaml serve
+	go run cmd/main.go --logtostderr --server.kube-config ~/.kube/config  --config ~/flyteadmin_config.yaml serve
```
EngHabu: do you want this change in here?

pkg/auth/config/config.go
```diff
@@ -51,6 +51,13 @@ type OAuthOptions struct {
 	// name. Instead, there is a gRPC interceptor, GetAuthenticationCustomMetadataInterceptor, that will translate
 	// incoming metadata headers with this config setting's name, into that standard header
 	GrpcAuthorizationHeader string `json:"grpcAuthorizationHeader"`
+
+	// To help ease migration, it was helpful to be able to only selectively enforce authentication.  The
+	// dimension that made the most sense to cut by at time of writing is HTTP vs gRPC as the web UI mainly used HTTP
+	// and the backend used mostly gRPC.  Cutting by individual endpoints is another option but it possibly falls more
+	// into the realm of authorization rather than authentication.
+	EnforceHttp bool `json:"enforceHttp"`
```
EngHabu: Is the plan to completely deprecate these options once Auth is enforced and tested across the board? or are we going to leave them here? if we are going to leave them (my vote is to deprecate them), then I think the default value for all bool-options should be false. In other words, I think we should inverse the names here `DisableForHttp` and `DisableForGrpc` or something to that effect..

pkg/auth/handlers.go
```diff
@@ -150,38 +150,35 @@ func GetAuthenticationCustomMetadataInterceptor(authCtx interfaces.Authenticatio
...
-			logger.Debugf(ctx, "Could not retrieve bearer token from metadata %v", err)
-			return ctx, nil
-		}
 
-		// Currently auth is optional...
-		if tokenStr == "" {
-			logger.Debugf(ctx, "Bearer token is empty, skipping parsing")
-			return ctx, nil
-		}
+		fromHTTP := metautils.ExtractIncoming(ctx).Get("from_http")
```
EngHabu: nit: move "from_http" to some constant

Makefile
katrogan: +1 please revert

PR #46, opened 2019-12-20
Refactor bootstrapped application config options

katrogan:


flyteadmin_config.yaml
```diff
@@ -25,6 +25,10 @@ server:
         iss: "https://idp.com"
         aud: "api://default"
       idpUserInfoEndpoint: "/v1/userinfo"
+  applicationConfig:
```
wild-endeavor: can we call this thirdPartyConfig?

katrogan: done

PR #50, opened 2020-01-03
Add audit logging

katrogan:


pkg/audit/log.go
```diff
@@ -0,0 +1,113 @@
...
+		tokenIssuedAt, err := ptypes.TimestampProto(clientMeta.(AuthenticatedClientMeta).TokenIssuedAt)
+		if err != nil {
+			logger.Warningf(ctx, "Failed to convert authenticated TokenIssuedAt to timestamp proto: %v", err)
+		}
+		b.auditLog.Principal = &admin.Principal{
+			Subject:       ctx.Value(common.PrincipalContextKey).(string),
+			TokenIssuedAt: tokenIssuedAt,
+		}
+		if len(clientMeta.(AuthenticatedClientMeta).ClientIds) > 0 {
+			b.auditLog.Principal.ClientId = clientMeta.(AuthenticatedClientMeta).ClientIds[0]
```
wild-endeavor: will we ever store more than the first one?

pkg/config/config.go
```diff
@@ -22,10 +22,11 @@ type ServerConfig struct {
...
 type ServerSecurityOptions struct {
-	Secure  bool                 `json:"secure"`
-	Ssl     SslOptions           `json:"ssl"`
-	UseAuth bool                 `json:"useAuth"`
-	Oauth   config2.OAuthOptions `json:"oauth"`
+	Secure      bool                 `json:"secure"`
+	Ssl         SslOptions           `json:"ssl"`
+	UseAuth     bool                 `json:"useAuth"`
+	Oauth       config2.OAuthOptions `json:"oauth"`
+	AuditAccess bool                 `json:"auditAccess"`
```
wild-endeavor: will we ever want this off?  this pr doesn't write to the database right?  so long as it's just a log line, can it just always be on if auth is on?

pkg/audit/log.go
katrogan: I'm not sure - is the convention ever to have multiple client ids? i can make this an array/slice, what do you think?

PR #51, opened 2020-01-03
Implement get and delete for attributes

katrogan:
Currently ProjectAttributes, ProjectDomainAttributes and WorkflowAttributes can be used to specify resource overrides. This change implements a delete API for all these customizable attributes AND a get to check what the existing values are currently set as.

In the future, only system administrators will be able to mutate these values once we have authorization rolled out.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -453,7 +453,7 @@
   version = "v1.2.0"
 
 [[projects]]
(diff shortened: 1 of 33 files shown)


wild-endeavor:
can we add an info level log line for all deletes?

PR #54, opened 2020-01-14
Refactor the resources model of FlyteAdmin

anandswaminathan:
This is first part of the placement manager approach to help Admin map executions to clusters.

This PR replaces the existing three table approach and brings it down to a single Resource Table.

The resource repo handles the logic of mapping resource based on hierarchy, and falling back when needed. The idea is if someone overrides resources at LaunchPlan/Workflow layer, that should take precedence over the Project/Domain. The ordering is LaunchPlan > Workflow > Project > Domain. (Added Comments for this logic at the Repo layer)

The lowest scope is a particular Domain. If one wants to override values for a Project, then one has to explicitly add all the Project-Domain values. 

Cleaning existing handlers and managers.

TODO: Move ResourceRequest and Response to IDL, once we expose.

cc @akhurana001 @katrogan @bnsblue


pkg/repositories/models/resource.go
```diff
@@ -0,0 +1,25 @@
...
+	ResourcePriorityProjectDomainLevel ResourcePriority = 10
+	ResourcePriorityWorkflowLevel      ResourcePriority = 100
+	ResourcePriorityLaunchPlanLevel    ResourcePriority = 1000
+)
+
+// Represents Flyte resources repository.
+// In this model, the combination of (Project, Domain, Workflow, LaunchPlan, ResourceType) is unique
+type Resource struct {
+	BaseModel
+	ID           int64  `gorm:"AUTO_INCREMENT;column:id;primary_key"`
```
katrogan: how does this field work with the BaseModel ID field? do the struct tags here still get applied?

anandswaminathan: Oh yes. I don't believe I can use the BaseModel then. Should I just copy over 

```	CreatedAt time.Time
	UpdatedAt time.Time
	DeletedAt *time.Time `sql:"index"```

Also that is one of the things I wanted to ask - how can I do end to end test?

File: pkg/repositories/gormimpl/resource_repo_test.go [hunk 2]
```diff
@@ -65,14 +65,43 @@ func TestGetWorkflowAttributes(t *testing.T) {
 	assert.Equal(t, []byte("attrs"), output.Attributes)
 }
 
+func TestProjectDomainAttributes(t *testing.T) {
+	resourceRepo := NewResourceRepo(GetDbForTest(t), errors.NewTestErrorTransformer(), mockScope.NewTestScope())
+	GlobalMock := mocket.Catcher.Reset()
+	GlobalMock.Logging = true
```
katrogan: nit: remove?

tests/attributes_test.go
```diff
@@ -103,19 +58,36 @@ func TestUpdateProjectDomainAttributes(t *testing.T) {
...
+		Domain:       "development",
+		Workflow:     "workflow",
+		ResourceType: admin.MatchableResource_TASK_RESOURCE,
+	})
+	assert.Nil(t, err)
+	assert.True(t, proto.Equal(&admin.WorkflowAttributesGetResponse{
+		Attributes: &admin.WorkflowAttributes{
+			Project:            "admintests",
+			Domain:             "development",
+			Workflow:           "",
```
katrogan: why is the workflow name empty?

tests/bootstrap.go
```diff
@@ -53,7 +53,8 @@ func truncateAllTablesForTestingOnly() {
 	TruncateNodeExecutions := fmt.Sprintf("TRUNCATE TABLE node_executions;")
 	TruncateNodeExecutionEvents := fmt.Sprintf("TRUNCATE TABLE node_execution_events;")
 	TruncateTaskExecutions := fmt.Sprintf("TRUNCATE TABLE task_executions;")
-	db := database_config.OpenDbConnection(database_config.NewPostgresConfigProvider(getDbConfig(), adminScope))
+	TruncateResources := fmt.Sprintf("TRUNCATE TABLE resources;")
+	db := database_config.OpenDbConnection(database_config.NewPostgresConfigProvider(getLocalDbConfig(), adminScope))
```
katrogan: revert?

tests/bootstrap.go
```diff
@@ -64,11 +65,12 @@ func truncateAllTablesForTestingOnly() {
 	db.Exec(TruncateNodeExecutions)
 	db.Exec(TruncateNodeExecutionEvents)
 	db.Exec(TruncateTaskExecutions)
+	db.Exec(TruncateResources)
 }
 
 func populateWorkflowExecutionForTestingOnly(project, domain, name string) {
 	InsertExecution := fmt.Sprintf(insertExecutionQueryStr, project, domain, name, "UNDEFINED", 1, 2)
-	db := database_config.OpenDbConnection(database_config.NewPostgresConfigProvider(getDbConfig(), adminScope))
+	db := database_config.OpenDbConnection(database_config.NewPostgresConfigProvider(getLocalDbConfig(), adminScope))
```
katrogan: revert this too?

tests/execution_test.go
```diff
@@ -156,7 +156,7 @@ func populateWorkflowExecutionsForTestingOnly() {
 		fmt.Sprintf(insertExecutionQueryStr, "project1", "domain2", "name1", "RUNNING", 1, 2),
 		fmt.Sprintf(insertExecutionQueryStr, "project2", "domain2", "name1", "SUCCEEDED", 1, 2),
 	}
-	db := database_config.OpenDbConnection(database_config.NewPostgresConfigProvider(getDbConfig(), adminScope))
+	db := database_config.OpenDbConnection(database_config.NewPostgresConfigProvider(getLocalDbConfig(), adminScope))
```
katrogan: and revert this one?

tests/attributes_test.go
anandswaminathan: Added comment

PR #59, opened 2020-02-03
Adding special logic for L5 to get OAuth to work

wild-endeavor:


pkg/auth/auth_context.go
```diff
@@ -172,3 +172,16 @@ func GetOauth2Config(options config.OAuthOptions) (oauth2.Config, error) {
 		},
 	}, nil
 }
+
+func GetL5Oauth2Config(mainConfig *oauth2.Config) oauth2.Config {
```
katrogan: this is admin public. why are we including l5 specific logic here?

katrogan:
this whole pr shouldn't be in admin public. if we need l5 logic handling can we re-use the existing designated lyft branch? `Lyft-Flyte`

pkg/auth/auth_context.go
wild-endeavor: Can we port this to the Lyft private version?

wild-endeavor:
let me re-open this PR into the lyft specific branch

PR #69, opened 2020-02-12
[MB-2105] List all matchable attributes

katrogan:


katrogan:
@anandswaminathan

katrogan:
ping @anandswaminathan @EngHabu

pkg/repositories/gormimpl/resource_repo.go
```diff
@@ -20,6 +20,8 @@ type ResourceRepo struct {
 	metrics          gormMetrics
 }
 
+const priorityDescending = "priority desc"
```
EngHabu: nit: Is there an `ID.Priority` that has the `priority` string to use here?

katrogan: from the docs it looks like you still have to explicitly use the string field name :\

on the plus side, model fields should change infrequently :-)

PR #70, opened 2020-02-12
Remove primary queue references

katrogan:
This value isn't used for anything since only dynamic tasks use queues

### DIFF
--- a/flyteadmin_config.yaml
+++ b/flyteadmin_config.yaml
@@ -104,18 +104,17 @@ queues:
     - dynamic: "default"
       attributes:
         - defaultclusters
-  workflowConfigs      -
-    - primary: "my_queue_1"
+  workflowConfigs
+    - project: "my_queue_1"
       domain: "production"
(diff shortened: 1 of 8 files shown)


flyteadmin_config.yaml
```diff
@@ -95,29 +95,26 @@ storage:
...
+    - dynamic: "critical"
       attributes:
         - critical
-    - primary: "default"
-      dynamic: "default"
+    - dynamic: "default"
       attributes:
         - defaultclusters
-    - primary: "my_queue_1"
+  workflowConfigs
```
EngHabu: Is this a new change?

anandswaminathan: workflowConfigs is not relevant anymore right?

katrogan: this file specifically was borked. we still read per-domain defaults from the config https://github.com/lyft/flyteadmin/blob/master/pkg/manager/impl/executions/queues.go#L90

PR #71, opened 2020-02-27
[Backward Incompatible] Implementation of Placement manager and multi cluster refactor 

anandswaminathan:
This PR uses the resource manager from https://github.com/lyft/flyteadmin/pull/54 and refactors the multi cluster implementation. The main value addition is the ability to route executions based on Project/Domain/Workflow/LaunchPlan - just like other resource overrides.

What has changed ?
- Weights and allowedDomains are removed from the cluster config.
- Introduces `labelClusterMap`, in which a label points to a list of clusters with weights.
- Labels can be set to resources at Project/Domain/Workflow/LaunchPlan level.

```
  labelClusterMap:
    test:
      - id: testcluster
        weight: 1
    all:
      - id: testcluster2
        weight: 0.5
      - id: testcluster3
        weight: 0.5
```
* Added/Updated unit tests.

The backward incompatibility comes in the fact that all the settings from the Config needs to be moved to the Database. 

@katrogan @kumare3

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -33,14 +33,6 @@
   version = "v13.3.1"
 
 [[projects]]
-  digest = "1:9f3b30d9f8e0d7040f729b82dcbc8f0dead820a133b3147ce355fc451f32d761"
-  name = "github.com/BurntSushi/toml"
-  packages = ["."]
-  pruneopts = "UT"
-  revision = "3012a1dbe2e4bd1391d42b32f0577cb7bbc7f005"
-  version = "v0.3.1"
-
-[[projects]]
   digest = "1:4d8aa8bc01f60d0fd7f764e1838f26dbc5a5dec428217f936726007cdf3929f0"
   name = "github.com/NYTimes/gizmo"
   packages = [
@@ -54,28 +46,12 @@
   version = "v0.4.3"
 
 [[projects]]
-  digest = "1:18fca7f682febe8a0f74f7d63408993043ad68a1566671ecafc1e35a59eca6b4"
-  name = "github.com/OpenPeeDeeP/depguard"
-  packages = ["."]
-  pruneopts = "UT"
-  revision = "2c5871f4f2dd78afab3c5075f292cdadb1f1d2d8"
-  version = "v1.0.1"
-
-[[projects]]
   digest = "1:7e704bce17074e862cfe9e4c2849320c2628fc3501b7d0795c589a427ef2bf50"
   name = "github.com/Selvatico/go-mocket"
   packages = ["."]
   pruneopts = "UT"
   revision = "c368d4162be502eea110ae12fb85e98567b0f1e6"
   version = "v1.0.7"
-
-[[projects]]
-  digest = "1:03a1d1b3090ed320c6e3cd5b936729430e4e255ca06e32d0a50a3832d88fb1ac"
-  name = "github.com/alvaroloes/enumer"
-  packages = ["."]
-  pruneopts = "UT"
-  revision = "642f2c04365f76460301ce0ea10d0dbbdf0b0508"
-  version = "v1.1.2"
 
 [[projects]]
   digest = "1:39cc2836b0733b809ab56eb991d3d58c96c08b6381fded8dc636c53656a0f5cc"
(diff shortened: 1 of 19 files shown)


Gopkg.lock
```diff
@@ -32,6 +32,14 @@
   revision = "21d4b01533b1005be0d020da67a6d3f8ebdf0141"
   version = "v13.3.1"
 
+[[projects]]
```
katrogan: aren't we on gomod?

pkg/executioncluster/execution_target.go
```diff
@@ -11,7 +10,11 @@ import (
 // Spec to determine the execution target
 type ExecutionTargetSpec struct {
 	TargetID    string
-	ExecutionID *core.WorkflowExecutionIdentifier
+	ExecutionID string
```
katrogan: why not keep the ExecutionID to make it backwards compatible and add the Workflow Identifier + Launch Plan Identifier as additional params? the project+domain aren't always identical across execution, workflow and launch plan anyways, no?

pkg/executioncluster/impl/random_cluster_selector.go
```diff
@@ -123,51 +107,72 @@ func (s RandomClusterSelector) GetAllValidTargets() []executioncluster.Execution
...
-	return nil, fmt.Errorf("invalid executionTargetSpec %v", *spec)
+	var weightedRandomList random.WeightedRandomList
+	if resource != nil && resource.Attributes.GetExecutionClusterLabel() != nil {
+		label := resource.Attributes.GetExecutionClusterLabel().Value
 
+		if _, ok := s.labelWeightedRandomMap[label]; ok {
+			weightedRandomList = s.labelWeightedRandomMap[label]
+		}
+	}
+	// If there is no label associated (or) if the label is invalid, choose from all enabled clusters.
```
katrogan: should we log when the label is invalid?

pkg/executioncluster/execution_target.go
anandswaminathan: Ohhhhh interesting. I guess will then use the launch plan's value.

anandswaminathan: @katrogan I thought more. This is actually a weird state for resource overrides. 

I will use the Domain, project from the execution. And name of launchplan and workflow.
What that would mean is Execution launched in a different project/domain would need a separate override even if the workflow does not exist there.

pkg/executioncluster/impl/random_cluster_selector.go
anandswaminathan: Sure

pkg/executioncluster/execution_target.go
katrogan: ok that sounds reasonable. how are you planning on migrating the db? (we can talk about this separately from the PR offline)

pkg/executioncluster/impl/random_cluster_selector.go
```diff
@@ -123,51 +109,77 @@ func (s RandomClusterSelector) GetAllValidTargets() []executioncluster.Execution
...
+		Domain:       spec.Domain,
+		Workflow:     spec.Workflow,
+		LaunchPlan:   spec.LaunchPlan,
+		ResourceType: admin.MatchableResource_EXECUTION_CLUSTER_LABEL,
+	})
+	if err != nil {
+		if flyteAdminError, ok := err.(errors.FlyteAdminError); !ok || flyteAdminError.Code() != codes.NotFound {
+			return nil, err
 		}
+		return nil, err
```
honnix, line 133: Is this intended? The logic here is, if a resource is not found, an error will be returned, but it doesn't seem to be the intention given there is a null check for `resource` on line 136. My understanding is, if there is no override, and no label is configured, all defined clusters will be used as fallback.

honnix: @anandswaminathan @katrogan

kumare3: @honnix you seem to have a point, the `return nil, err` in both cases
1. When err = Not Found
2. all other errors seems fishy.

Let me Talk to @katrogan and @anandswaminathan

PR #76, opened 2020-03-12
Add execution cluster to system meta

katrogan:
# TL;DR
Store and retrieve system information associated with the execution

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/206

## Follow-up issue
NA


kumare3:
Awesome! So what happens in sandbox case, what is the default value for cluster?

katrogan:
> Awesome! So what happens in sandbox case, what is the default value for cluster?

Should just be an empty string then

PR #77, opened 2020-03-16
Use Update in sync cluster resources by default

katrogan:
# TL;DR
It turns out that updating kubernetes resources regularly (and successfully) requires the Update rest call. However, serviceaccount, special snowflake that it is, doesn't respect Update and will actually create a new corresponding secret each time Update is called, even if nothing changes across the original and to-Update resource. This change whitelists a set of resources for which to use StrategicMergePatch with when updating. This mimics the behavior `kubectl apply` uses when updating serviceaccounts. In the future, if we need to use other Patch strategies we can refactor this code - if necessary - to have the whitelist map to Patch Types.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested (local sandbox with docker desktop)
 - [x] Unit tests added

## Complete description
See PR description

## Tracking Issue
N/A

## Follow-up issue
N/A

### DIFF
--- a/pkg/clusterresource/controller.go
+++ b/pkg/clusterresource/controller.go
@@ -90,6 +90,10 @@ var descCreatedAtSortParam, _ = common.NewSortParameter(admin.Sort{
 // Kubectl defaults to using the StrategicMergePatch strategy.
 // However the controller-runtime only has an implementation for MergePatch which we were formerly
 // using but failed to actually always merge resources in the Patch call.
+// INTERESTINGLY Patch doesn't actually appear to update the majority of resources. We default to using Update but
+// whitelist the specific set of resources that require a Patch to work instead.
+// If you use update with a ServiceAccount - *every* call to Update results in a new corresponding secret being created
+// which has the (not so) fun side-effect of overwhelming API server when this Sync script is run as a cron.
 var strategicPatchTypes = map[string]bool {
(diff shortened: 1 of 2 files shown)


pkg/clusterresource/controller.go
```diff
@@ -85,6 +87,18 @@ var descCreatedAtSortParam, _ = common.NewSortParameter(admin.Sort{
...
 
+// Use a strategic-merge-patch to mimic `kubectl apply` behavior for serviceaccounts.
+// Kubectl defaults to using the StrategicMergePatch strategy.
+// However the controller-runtime only has an implementation for MergePatch which we were formerly
+// using but failed to actually always merge resources in the Patch call.
+// INTERESTINGLY Patch doesn't actually appear to update the majority of resources. We default to using Update but
+// whitelist the specific set of resources that require a Patch to work instead.
+// If you use update with a ServiceAccount - *every* call to Update results in a new corresponding secret being created
+// which has the (not so) fun side-effect of overwhelming API server when this Sync script is run as a cron.
+var strategicPatchTypes = map[string]bool{
```
EngHabu: I would have used sets.String{} instead but I guess too late :D

katrogan: golang has sets ?!?

PR #82, opened 2020-03-30
Add configurable schedule name prefix

katrogan:
# TL;DR
This removes the hard coded schedule name prefix and makes it configurable.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
For multiple deployments of flyteadmin (e.g. a staging, production, etc) the same schedule name prefix is applied without configuration across deployments. This means enabling or disabling a schedule in one deployment has the effect of adding or removing across all deployments such that any matching workflows will be scheduled/unscheduled.

## Tracking Issue
https://github.com/lyft/flyte/issues/227

## Follow-up issue
_NA_

### DIFF
--- a/flyteadmin_config.yaml
+++ b/flyteadmin_config.yaml
@@ -52,6 +52,7 @@ scheduler:
     region: "my-region"
     scheduleRole: "arn:aws:iam::abc123:role/my-iam-role"
     targetName: "arn:aws:sqs:my-region:abc123:my-queue"
+    scheduleNamePrefix: "flyte"
   workflowExecutor:
     scheme: local
     region: "my-region"
--- a/pkg/runtime/interfaces/application_configuration.go
+++ b/pkg/runtime/interfaces/application_configuration.go
@@ -31,10 +31,11 @@ type ApplicationConfig struct {
 }
 
 type EventSchedulerConfig struct {
-	Scheme       string `json:"scheme"`
-	Region       string `json:"region"`
-	ScheduleRole string `json:"scheduleRole"`
-	TargetName   string `json:"targetName"`
+	Scheme             string `json:"scheme"`
+	Region             string `json:"region"`
+	ScheduleRole       string `json:"scheduleRole"`
+	TargetName         string `json:"targetName"`
+	ScheduleNamePrefix string `json:"scheduleNamePrefix"`
 }
 
 type WorkflowExecutorConfig struct {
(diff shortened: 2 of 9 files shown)


flyteadmin_config.yaml
```diff
@@ -52,6 +52,7 @@ scheduler:
     region: "my-region"
     scheduleRole: "arn:aws:iam::abc123:role/my-iam-role"
     targetName: "arn:aws:sqs:my-region:abc123:my-queue"
+    scheduleNamePrefix: "flyte"
```
kumare3: can we make this optional? and the default can be `flyte`

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -31,10 +31,11 @@ type ApplicationConfig struct {
 }
 
 type EventSchedulerConfig struct {
```
kumare3: Documentation please

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -31,10 +31,11 @@ type ApplicationConfig struct {
...
-	Scheme       string `json:"scheme"`
-	Region       string `json:"region"`
-	ScheduleRole string `json:"scheduleRole"`
-	TargetName   string `json:"targetName"`
+	Scheme             string `json:"scheme"`
```
kumare3: Can we document every element in the config?

flyteadmin_config.yaml
katrogan: It is optional, if it's empty we just don't join "<prefix>_" with the identifier hash

anandswaminathan:
+1

Internally we have worked around this using different launch plan names.

flyteadmin_config.yaml
kumare3: but i think having "flyte_" is better than just an identifier hash. In some organizations, this will help them figure out how the schedules were created

katrogan: the schedule description includes a descriptive identifier that pinpoints it's a workflow execution e.g. `Schedule for Project:flytekit Domain:production Name:workflows.app.MyWorkflow`

katrogan:
@kumare3 added documentation to the config file, PTAL

pkg/runtime/interfaces/application_configuration.go
katrogan: done

kumare3:
Thank you @katrogan - your rock!

PR #89, opened 2020-04-13
Downgrade info logs in clusterresource controller

katrogan:
# TL;DR
This change downgrades info logs in clusterresource controller to debug level to reduce its overall spamminess and address disk space pressure for docker-for-mac deployments.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/250

## Follow-up issue
_NA_

### DIFF
--- a/pkg/clusterresource/controller.go
+++ b/pkg/clusterresource/controller.go
@@ -239,14 +239,14 @@ func (c *controller) syncNamespace(ctx context.Context, namespace NamespaceName,
 		templateFileName := templateFile.Name()
 		if filepath.Ext(templateFileName) != ".yaml" {
 			// nothing to do.
-			logger.Infof(ctx, "syncing namespace [%s]: ignoring unrecognized filetype [%s]",
+			logger.Debugf(ctx, "syncing namespace [%s]: ignoring unrecognized filetype [%s]",
 				namespace, templateFile.Name())
 			continue
 		}
 
 		if c.templateAlreadyApplied(namespace, templateFile) {
 			// nothing to do.
-			logger.Infof(ctx, "syncing namespace [%s]: templateFile [%s] already applied, nothing to do.", namespace, templateFile.Name())
+			logger.Debugf(ctx, "syncing namespace [%s]: templateFile [%s] already applied, nothing to do.", namespace, templateFile.Name())
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/clusterresource/controller.go
```diff
@@ -239,14 +239,14 @@ func (c *controller) syncNamespace(ctx context.Context, namespace NamespaceName,
 		templateFileName := templateFile.Name()
 		if filepath.Ext(templateFileName) != ".yaml" {
 			// nothing to do.
-			logger.Infof(ctx, "syncing namespace [%s]: ignoring unrecognized filetype [%s]",
+			logger.Debugf(ctx, "syncing namespace [%s]: ignoring unrecognized filetype [%s]",
```
kumare3, line 242: is this not a warning? We are ignoring something?

pkg/clusterresource/controller.go
```diff
@@ -315,7 +315,7 @@ func (c *controller) syncNamespace(ctx context.Context, namespace NamespaceName,
 
 					if err != nil {
 						c.metrics.TemplateUpdateErrors.Inc()
-						logger.Infof(ctx, "Failed to update resource [%+v] in namespace [%s] with err :%v",
+						logger.Warningf(ctx, "Failed to update resource [%+v] in namespace [%s] with err :%v",
```
kumare3, line 318: `%+v`, maybe that is a problem?

pkg/clusterresource/controller.go
katrogan: not necessarily, we only look at yaml files (because of how kubernetes mounts configmaps with symlinks this was causing problems in the sandbox mode)

pkg/clusterresource/controller.go
katrogan: how so?

kumare3: we are dumping out the entire object, is that want you intend to do?

kumare3:
@katrogan if you think the intention matches the code, I am good to go with this. Ping me and we can approve it

pkg/clusterresource/controller.go
katrogan: we're only dumping out the Kind

PR #95, opened 2020-04-30
ErrorCode and ErrorKind columns to executions

kumare3:
# TL;DR
The intention is to make it possible to calculate errors for executions or node_executions by kind and eventually categorize by error codes

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Refer to issue

## Tracking Issue
https://github.com/lyft/flyte/issues/300

## Follow-up issue
NA


pkg/repositories/config/migrations.go
```diff
@@ -199,4 +199,14 @@ var Migrations = []*gormigrate.Migration{
 			return tx.Exec("ALTER TABLE workflows ADD COLUMN IF NOT EXISTS state integer;").Error
 		},
 	},
+	// Modify the executions & node_executison table, if necessary
+	{
+		ID: "2020-04-29-executions",
+		Migrate: func(tx *gorm.DB) error {
+			return tx.Exec("DO we need to do migrations for nullable column adds?").Error
```
katrogan, line 206: yep, we still need to add the schema to the table - without this the most recent version of the executions model will determine the table schema

pkg/repositories/transformers/execution.go
```diff
@@ -135,6 +136,10 @@ func UpdateExecutionModelState(
 				},
 			},
 		}
+	} else if request.Event.GetError() != nil {
```
EngHabu, line 139: nit: I think this should move to the next if block

katrogan, line 139: isn't this condition checked in line 145 below? can you merge the blocks?

pkg/repositories/config/migrations.go
kumare3: so alter table?

kumare3: and how about indexes? This sounds a little weird. Can we just not run auto-migration on the model?

pkg/repositories/transformers/execution.go
kumare3: done

kumare3: done

pkg/repositories/models/execution.go
```diff
@@ -51,4 +51,8 @@ type Execution struct {
 	InputsURI storage.DataReference
 	// User specified inputs. This map might be incomplete and not include defaults applied
 	UserInputsURI storage.DataReference
+	// Execution Error Kind. nullable
```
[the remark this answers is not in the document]
kumare3: yes, is there a way to document that, reference to flyteidl?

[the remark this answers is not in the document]
katrogan: that would be great if you don't mind (sorry I realized it's missing from Phase above)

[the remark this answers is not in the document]
kumare3: done

PR #96, opened 2020-05-12
 Single task execution (admin-generated workflow implementation)

katrogan:
# TL;DR
This PR implements single task execution for CreateExecutionRequests with reference entities of type 'TASK'.

Because this is v1 of single task execution this code is compartmentalized and designed to be easy to rip out. Therefore not everything that could necessarily be refactored from the existing CreateExecution, CreateWorfklow and CreateLaunchPlan logic is changed and some duplicate code exists. This is (ugly) by design.

**TODO** If this overall design looks okay I will add more unit tests to this PR - I just wanted to get an overall approval of the code before investing time in unit tests in case it changes significantly.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
This code takes an existing, registered task and generates a skeleton workflow definition with a single node referencing the task. The workflow inputs and outputs are mapped to those of the task definition. For UI purposes, a default launch plan with a matching interface is also generated.

## Tracking Issue
https://github.com/lyft/flyte/issues/256

## Follow-up issue
https://github.com/lyft/flyte/issues/221


wild-endeavor:
Is this really "v1".  I'd rather not call it that.

katrogan:
@wild-endeavor updated but names are hard. let me know if you have suggestions

pkg/manager/impl/util/single_task_execution.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+	return bindings, nil
+}
+
+func CreateOrGetWorkflowModel(
+	ctx context.Context, request admin.ExecutionCreateRequest, db repositories.RepositoryInterface,
+	workflowManager interfaces.WorkflowInterface, namedEntityManager interfaces.NamedEntityInterface, taskIdentifier *core.Identifier,
+	task *admin.Task) (*models.Workflow, error) {
+	workflowModel, err := db.WorkflowRepo().Get(ctx, repositoryInterfaces.GetResourceInput{
+		Project: taskIdentifier.Project,
```
EngHabu, line 185: We talked about creating these skeleton WFs in a dummy project/domain. So that we don't clutter the UI. Did that not work?

pkg/manager/impl/util/single_task_execution.go
```diff
@@ -0,0 +1,340 @@
...
+		bindings = append(bindings, binding)
+	}
+	return bindings, nil
+}
+
+func CreateOrGetWorkflowModel(
+	ctx context.Context, request admin.ExecutionCreateRequest, db repositories.RepositoryInterface,
+	workflowManager interfaces.WorkflowInterface, namedEntityManager interfaces.NamedEntityInterface, taskIdentifier *core.Identifier,
+	task *admin.Task) (*models.Workflow, error) {
+	workflowModel, err := db.WorkflowRepo().Get(ctx, repositoryInterfaces.GetResourceInput{
```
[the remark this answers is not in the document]
katrogan: true but i'm not sure how to guarantee against collisions using any naming scheme. couldn't a random hash or prefix/suffix pattern fall into the same trap? (i suppose with less likelihood)

pkg/manager/impl/util/single_task_execution.go
katrogan: unfortunately projects are registered dynamically although we could seed the database with a system project like we do in the sandbox but we also have https://github.com/lyft/flyteidl/blob/4c689bdb271db83c1934505282d5e72981f23a51/protos/flyteidl/admin/common.proto#L33 as a separate status denoting origin/status

pkg/manager/impl/util/single_task_execution.go
```diff
@@ -0,0 +1,340 @@
...
+const maxNodeIDLength = 63
+
+var defaultRetryStrategy = core.RetryStrategy{
+	Retries: 3,
+}
+var defaultTimeout = ptypes.DurationProto(24 * time.Hour)
+
+func generateNodeNameFromTask(taskName string) string {
+	if len(taskName) >= maxNodeIDLength {
+		taskName = taskName[:maxNodeIDLength]
```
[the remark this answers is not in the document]
katrogan: excellent point! updated

pkg/manager/impl/util/single_task_execution.go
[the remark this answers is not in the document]
EngHabu: True, the entropy is much higher with random strings, but I guess introducing a random string means we will create this many many times (with every execution) creating a lot of garbage... 

Maybe just introducing a known constant obscure prefix/suffix is fine then?

pkg/manager/impl/util/single_task_execution.go
```diff
@@ -0,0 +1,340 @@
...
+		bindings = append(bindings, binding)
+	}
+	return bindings, nil
+}
+
+func CreateOrGetWorkflowModel(
+	ctx context.Context, request admin.ExecutionCreateRequest, db repositories.RepositoryInterface,
+	workflowManager interfaces.WorkflowInterface, namedEntityManager interfaces.NamedEntityInterface, taskIdentifier *core.Identifier,
+	task *admin.Task) (*models.Workflow, error) {
+	workflowModel, err := db.WorkflowRepo().Get(ctx, repositoryInterfaces.GetResourceInput{
```
EngHabu, line 184: Doesn't this introduce a race condition? if I tried to run the same task twice, both of them go past the "Get" call and realize there isn't an existing WF model, so they will both try to create one... Should the error check for Create() call below exclude "ErrExistingRow" or whatever the error is for that?

katrogan:
> Are you also going to add the automatic filter for System Generated in the ListWF Endpoint?
Added the filter to the ListNamedEntity endpoint which is what the console uses to list workflows and is the entrypoint for individual workflows 

I'm more hesitant about adding it to the existing ListWorkflows endpoint by default because now that will force an additional join and I don't want to add a performance penalty to an endpoint that is really only available in the cli (aka power users who hopefully know what they're doing :) )

pkg/manager/impl/util/single_task_execution.go
katrogan: yes, good call. done

pkg/manager/impl/util/single_task_execution.go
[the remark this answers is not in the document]
katrogan: sure done. let me know if you'd prefer a different name/format

katrogan:
friendly ping @EngHabu

pkg/manager/impl/util/single_task_execution.go
```diff
@@ -0,0 +1,353 @@
...
+			ResourceType: core.ResourceType_WORKFLOW,
+			Id: &admin.NamedEntityIdentifier{
+				Project: workflowIdentifier.Project,
+				Domain:  workflowIdentifier.Domain,
+				Name:    workflowIdentifier.Name,
+			},
+			Metadata: &admin.NamedEntityMetadata{State: admin.NamedEntityState_SYSTEM_GENERATED},
+		})
+		if err != nil {
+			logger.Warningf(ctx, "Failed to set skeleton workflow state to archived: %v", err)
```
EngHabu, line 270: nit:
```suggestion
			logger.Warningf(ctx, "Failed to set skeleton workflow state to system-generated: %v", err)
```

katrogan:
@EngHabu mind reviewing one last time? your suggested commit dismissed your review :)

PR #99, opened 2020-05-21
Ensure task resource requests are always less than limits

katrogan:
# TL;DR
For a specific resource (cpu, memory) the task requests should always be less than or equal to the corresponding limit value. Otherwise this results in a pod spec that is invalid and fails at execution time.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/264

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/execution_manager.go
+++ b/pkg/manager/impl/execution_manager.go
@@ -6,6 +6,8 @@ import (
 	"strconv"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/resource"
+
 	"github.com/lyft/flyteadmin/pkg/manager/impl/resources"
 
 	"github.com/golang/protobuf/ptypes"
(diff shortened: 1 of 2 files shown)


pkg/manager/impl/execution_manager.go
```diff
@@ -292,17 +336,34 @@ func setCompiledTaskDefaults(ctx context.Context, config runtimeInterfaces.Confi
...
 	}
-	task.Template.GetContainer().Resources.Limits = assignResourcesIfUnset(
+
+	var resourceLimitIndices = newTaskResourceIndices()
+	task.Template.GetContainer().Resources.Limits, resourceLimitIndices = assignResourcesIfUnset(
 		ctx, task.Template.Id, createTaskDefaultLimits(ctx, task), task.Template.GetContainer().Resources.Limits,
 		taskResourceSpec)
+
+	if !(resourceRequestIndices.isValid() && resourceLimitIndices.isValid()) {
+		logger.Warningf(ctx, "Validating task resource requests & limits that are incomplete [%+v]", task.Template.Id)
```
bnsblue, line 355: This warning message is a present progressive so the meaning is a bit unclear to me. Is it reporting what it is trying to do? Shouldn't it report that those indices are invalid?

pkg/manager/impl/execution_manager.go
```diff
@@ -213,24 +215,42 @@ func createTaskDefaultLimits(ctx context.Context, task *core.CompiledTask) runti
 	return runtimeInterfaces.TaskResourceSet{CPU: cpuLimit, Memory: memoryLimit}
 }
 
+// This type stores the locations of where a CPU and memory resource are defined in a []core.Resources_ResourceEntry.
+type taskResourceIndices struct {
```
bnsblue, line 219: Could you say a bit more why we need these indices?

pkg/manager/impl/execution_manager_test.go
```diff
@@ -2409,6 +2409,71 @@ func TestAssignResourcesIfUnset(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+		}
+		var resourceRequestIndex, resourceLimitIndex = 1, 0
+		err := finalizeTaskResources(context.Background(), task, resourceRequestIndex, resourceLimitIndex, "my imaginary resource")
+		assert.Nil(t, err)
+		assert.EqualValues(t, task.Template.GetContainer().Resources.Requests[resourceRequestIndex].Value,
```
bnsblue, line 2445: In your function the logic is that when request > limit, you **lower** the value of request to the same value as limit. I think in this test that behavior should be stated more expliciltly.

bnsblue:
Overall it looks great. Just a couple of nitpicks. Thanks so much for taking the action to get this fixed :D

pkg/manager/impl/execution_manager.go
```diff
@@ -213,24 +215,42 @@ func createTaskDefaultLimits(ctx context.Context, task *core.CompiledTask) runti
...
+		cpuIndex:    -1,
+		memoryIndex: -1,
+	}
+}
+
+func (i taskResourceIndices) isValid() bool {
+	return i.cpuIndex >= 0 && i.memoryIndex >= 0
+}
+
 func assignResourcesIfUnset(ctx context.Context, identifier *core.Identifier,
```
wild-endeavor, line 235: this function makes me hate go.

bnsblue: why

katrogan: haha any ugliness here is on me, not the language :P

pkg/manager/impl/execution_manager.go
```diff
@@ -255,8 +276,31 @@ func assignResourcesIfUnset(ctx context.Context, identifier *core.Identifier,
...
+	if err != nil {
+		return errors.NewFlyteAdminErrorf(codes.InvalidArgument, "Unable to parse requested cpu quantity [%s]: [%v]",
+			task.Template.GetContainer().Resources.Requests[resourceRequestIndex].Value, err)
+	}
+	limit, err := resource.ParseQuantity(task.Template.GetContainer().Resources.Limits[resourceLimitIndex].Value)
+	if err != nil {
+		return errors.NewFlyteAdminErrorf(codes.InvalidArgument, "Unable to parse cpu limit quantity [%s]: [%v]",
+			task.Template.GetContainer().Resources.Limits[resourceLimitIndex].Value, err)
+	}
+	if limit.Cmp(requested) < 0 {
```
EngHabu, line 298: This behavior needs to be documented somewhere... Somebody someday will wonder why it didn't happen the other way around and we need to be able to point to a doc that clearly states that.

bnsblue: Good point. And as said in my comment above that the corresponding unit test needs to reflect this behavior explicitly.

pkg/manager/impl/execution_manager.go
bnsblue: isn't this pretty common lol

katrogan:
> remind me again cuz i don't want to look - we do this check at registration time as well right?

we do the check at execution time so the defaults are applied dynamically

pkg/manager/impl/execution_manager.go
katrogan: sure the resources are a list (rather than a map) so we need to know where to reference specific resource values by resource type

katrogan:
given the feedback on this pr i'm going to drop it and rewrite to use the requested value when the limits are empty (and vice-versa) & with proper grammar, etc

PR #100, opened 2020-05-29
[do not use] Raise gRPC limit to 16 MB

wild-endeavor:
Temporarily raising the gRPC limit to 16MBs.  This PR should be reverted after the error truncation fix goes in.

### DIFF
--- a/cmd/entrypoints/serve.go
+++ b/cmd/entrypoints/serve.go
@@ -205,7 +205,7 @@ func serveGatewayInsecure(ctx context.Context, cfg *config.ServerConfig) error {
 	}()
 
 	logger.Infof(ctx, "Starting HTTP/1 Gateway server on %s", cfg.GetHostAddress())
-	httpServer, err := newHTTPServer(ctx, cfg, authContext, cfg.GetGrpcHostAddress(), grpc.WithInsecure())
+	httpServer, err := newHTTPServer(ctx, cfg, authContext, cfg.GetGrpcHostAddress(), grpc.WithInsecure(), grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(16777216)))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


cmd/entrypoints/serve.go
```diff
@@ -205,7 +205,7 @@ func serveGatewayInsecure(ctx context.Context, cfg *config.ServerConfig) error {
 	}()
 
 	logger.Infof(ctx, "Starting HTTP/1 Gateway server on %s", cfg.GetHostAddress())
-	httpServer, err := newHTTPServer(ctx, cfg, authContext, cfg.GetGrpcHostAddress(), grpc.WithInsecure())
+	httpServer, err := newHTTPServer(ctx, cfg, authContext, cfg.GetGrpcHostAddress(), grpc.WithInsecure(), grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(16777216)))
```
anandswaminathan, line 208: Const variable please :)

wild-endeavor: yes yes, but should we do this?

EngHabu: What's the default?
Does this number mean anything in particular :-D ?
Are you trying to limit it or make it more permissible?

katrogan: what happens when a response exceeds the size? does this get logged? does an error get returned?

wild-endeavor: Go to the link I'm about to send you on Slack.

wild-endeavor: The default size limit for grpc is 4MB, which is where the 4194304 number comes from in the error message (4*2^20).  This is just four times that so 16MB.  wondering if the truncation change is equally quick, in which case we should do that instead?  or another workaround that people would rather see?

anandswaminathan:
Congrat. PR: 100

PR #106, opened 2020-07-06
Update async retry mechanisms

katrogan:
# TL;DR
Add retries to notifications processor to handle underlying SQS connection failures.

In general, add user-config specified retries for initializing the initial AWS/GCP client but otherwise indefinitely retry in the case an open channel hiccups.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/376

## Follow-up issue
_NA_

### DIFF
--- a/pkg/async/notifications/factory.go
+++ b/pkg/async/notifications/factory.go
@@ -40,8 +40,7 @@ type EmailerConfig struct {
 	BaseURL     string
 }
 
-func GetEmailer(config runtimeInterfaces.NotificationsConfig, scope promutils.Scope,
-	reconnectAttempts int, reconnectDelay time.Duration) interfaces.Emailer {
+func GetEmailer(config runtimeInterfaces.NotificationsConfig, scope promutils.Scope) interfaces.Emailer {
 	switch config.Type {
 	case common.AWS:
 		awsConfig := aws.NewConfig().WithRegion(config.Region).WithMaxRetries(maxRetries)
--- a/pkg/async/notifications/factory.go
+++ b/pkg/async/notifications/factory.go
@@ -45,13 +45,7 @@ func GetEmailer(config runtimeInterfaces.NotificationsConfig, scope promutils.Sc
 	switch config.Type {
 	case common.AWS:
 		awsConfig := aws.NewConfig().WithRegion(config.Region).WithMaxRetries(maxRetries)
-		var awsSession *session.Session
-		var err error
-		err = async.Retry(reconnectAttempts, reconnectDelay, func() error {
-			awsSession, err = session.NewSession(awsConfig)
-			return err
-		})
-
+		awsSession, err := session.NewSession(awsConfig)
 		if err != nil {
 			panic(err)
 		}
--- a/pkg/async/notifications/factory.go
+++ b/pkg/async/notifications/factory.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"time"
 
+	"github.com/lyft/flyteadmin/pkg/async"
+
 	"github.com/lyft/flyteadmin/pkg/async/notifications/implementations"
 	"github.com/lyft/flyteadmin/pkg/async/notifications/interfaces"
 	runtimeInterfaces "github.com/lyft/flyteadmin/pkg/runtime/interfaces"
@@ -38,11 +40,18 @@ type EmailerConfig struct {
 	BaseURL     string
 }
 
-func GetEmailer(config runtimeInterfaces.NotificationsConfig, scope promutils.Scope) interfaces.Emailer {
+func GetEmailer(config runtimeInterfaces.NotificationsConfig, scope promutils.Scope,
+	reconnectAttempts int, reconnectDelay time.Duration) interfaces.Emailer {
 	switch config.Type {
 	case common.AWS:
 		awsConfig := aws.NewConfig().WithRegion(config.Region).WithMaxRetries(maxRetries)
-		awsSession, err := session.NewSession(awsConfig)
+		var awsSession *session.Session
+		var err error
+		err = async.Retry(reconnectAttempts, reconnectDelay, func() error {
+			awsSession, err = session.NewSession(awsConfig)
+			return err
+		})
+
 		if err != nil {
 			panic(err)
 		}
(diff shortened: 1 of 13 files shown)


pkg/rpc/adminservice/base.go
```diff
@@ -98,11 +98,18 @@ func NewAdminServer(kubeConfig, master string) *AdminService {
 	publisher := notifications.NewNotificationsPublisher(*configuration.ApplicationConfiguration().GetNotificationsConfig(), adminScope)
 	processor := notifications.NewNotificationsProcessor(*configuration.ApplicationConfiguration().GetNotificationsConfig(), adminScope)
 	go func() {
+		logger.Info(context.Background(), "Started processing notifications.")
 		err = processor.StartProcessing()
```
kumare3, line 102: shouldnt start processing auto-handle connection loss and reconnect failures? Ofcourse it should log

katrogan: sure, refactored @kumare3

kumare3:
Looks good to me, just one thing, in case an error connection, should FlyteAdmin process not crash/exit?

katrogan:
I'm not sure - is a running admin with partial functionality better or should we fail fast?

kumare3:
> I'm not sure - is a running admin with partial functionality better or should we fail fast?

IMO, it could be a network partition or some other issue. If we fail fast, we might be able to recover on a different node? Also, since we have multiple replica's others should be fine.

On the other hand in sandbox mode we should probably not crash as users may not be able to figure out what the problem is. I pinged Ruslan. I guess if we do it for this, we should probably do it for all other cases?

katrogan:
Updated so that we retry initializing the initial AWS/GCP client the number of times specified in the config but otherwise indefinitely retry in the case an open channel hiccups.

pkg/async/notifications/factory.go
```diff
@@ -2,6 +2,9 @@ package notifications
 
 import (
 	"context"
+	"time"
+
+	"github.com/lyft/flyteadmin/pkg/async"
```
kumare3, line 7: looks like a future flytestdlib entity

pkg/async/notifications/factory.go
```diff
@@ -37,11 +40,18 @@ type EmailerConfig struct {
 	BaseURL     string
 }
 
-func GetEmailer(config runtimeInterfaces.NotificationsConfig, scope promutils.Scope) interfaces.Emailer {
+func GetEmailer(config runtimeInterfaces.NotificationsConfig, scope promutils.Scope,
+	reconnectAttempts int, reconnectDelay time.Duration) interfaces.Emailer {
 	switch config.Type {
 	case common.AWS:
 		awsConfig := aws.NewConfig().WithRegion(config.Region).WithMaxRetries(maxRetries)
```
kumare3, line 47: it seems aws config already takes retries and delay?

katrogan: derp, thanks for the catch. updated

pkg/async/notifications/implementations/processor.go
```diff
@@ -38,7 +41,16 @@ type Processor struct {
 // Currently only email is the supported notification because slack and pagerduty both use
 // email client to trigger those notifications.
 // When Pagerduty and other notifications are supported, a publisher per type should be created.
-func (p *Processor) StartProcessing() error {
+func (p *Processor) StartProcessing() {
+	for {
+		logger.Warningf(context.Background(), "Starting notifications processor")
+		err := p.run()
+		logger.Errorf(context.Background(), "error with running processor err: [%v] ", err)
```
wild-endeavor, line 48: should we log a metric here too?

pkg/async/schedule/aws/workflow_executor.go
```diff
@@ -168,6 +171,15 @@ func (e *workflowExecutor) formulateExecutionCreateRequest(
 }
 
 func (e *workflowExecutor) Run() {
+	for {
+		logger.Warningf(context.Background(), "Starting workflow executor")
+		err := e.run()
```
wild-endeavor, line 176: and here

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -149,6 +149,10 @@ type NotificationsConfig struct {
 	NotificationsPublisherConfig NotificationsPublisherConfig `json:"publisher"`
 	NotificationsProcessorConfig NotificationsProcessorConfig `json:"processor"`
 	NotificationsEmailerConfig   NotificationsEmailerConfig   `json:"emailer"`
+	// Number of times to attempt recreating a notifications processor client should there be any disruptions.
+	ReconnectAttempts int `json:"reconnectAttempts"`
+	// Specifies the time interval to wait before attempting to reconnect the notifications processor client.
+	ReconnectDelaySeconds int `json:"reconnectDelaySeconds"`
```
wild-endeavor, line 155: should this be int?  We can make it this if you want https://github.com/lyft/flytestdlib/blob/537f86093d86af270aab300742ee1a56f2905885/config/duration.go#L10

pkg/async/schedule/aws/workflow_executor.go
wild-endeavor: and this is a layer above the async auto-retry layer right?  so if we exhaust all retries, then we chill for half an hour and then try again?

katrogan: wow half an hour is a goof, meant it to be half a minute - will update

pkg/runtime/interfaces/application_configuration.go
katrogan: sure, but it's a question of how much granularity we need to expose - are people really going to configure retry delays on the order or many minutes or even hours?

pkg/async/notifications/implementations/processor.go
katrogan: we already log a metric when the channel closes

pkg/async/schedule/aws/workflow_executor.go
katrogan: added a metric on channel close

wild-endeavor: "async retry is for initializing clients a finite amount of times"

PR #107, opened 2020-07-08
Add quality of service logic to admin

katrogan:
## TL;DR
This PR implements translation from a user-specified quality of service to a queueing budget used for cost optimization.

Users can specify the quality of service for an execution (in order of decreasing specificity)

- At CreateExecution request time
- In the LaunchPlan spec
- In the Workflow spec
- As an overridable [MatchableResource](https://lyft.github.io/flyte/administrator/install/managing_customizable_resources.html)

System administrators can specify default [quality of service spec](https://github.com/lyft/flyteidl/blob/e9727afcedf8d4c30a1fc2eeac45593e426d9bb0/protos/flyteidl/core/execution.proto#L92)s for different [tiers](https://github.com/lyft/flyteidl/blob/e9727afcedf8d4c30a1fc2eeac45593e426d9bb0/protos/flyteidl/core/execution.proto#L101,L107) **and** default tiers for different domains.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
https://docs.google.com/document/d/1rYqr01x0XL6n1AdmEOUbTIGeS9Z2hJCcnG5NMy8aQ68/edit#

## Tracking Issue
https://github.com/lyft/flyte/issues/343

## Follow-up issue
_NA_


pkg/manager/impl/executions/quality_of_service.go
```diff
@@ -0,0 +1,164 @@
...
+			return QualityOfServiceSpec{
+				QueuingBudget: duration,
+			}, nil
+		}
+		qualityOfServiceTier = input.Workflow.Closure.CompiledWorkflow.Primary.Template.Metadata.QualityOfService.GetTier()
+	}
+
+	// If nothing in the hierarchy has set the quality of service, see if an override exists in the matchable attributes
+	// resource table.
+	resource, err := q.resourceManager.GetResource(ctx, interfaces.ResourceRequest{
```
EngHabu, line 105: Should we only do this if qualityOfServiceTier == UNDEFINED ?

katrogan: good call, thanks! refactored this before opening the PR and missed the check here

katrogan:
friendly ping @EngHabu @migueltol22

flyteadmin_config.yaml
```diff
@@ -159,3 +159,16 @@ cluster_resources:
       valueFrom:
         env: SHELL
   refresh: 3s
+qualityOfService:
```
wild-endeavor, line 162: Not a fan of this name - I just don't like the connotation of a low quality of service.  Is it too late to change this?

pkg/manager/impl/executions/quality_of_service.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	executionValues, ok := q.config.QualityOfServiceConfiguration().GetTierExecutionValues()[qualityOfServiceTier]
+	if !ok {
+		// No queueing budget to set when no default is specified
+		return QualityOfServiceSpec{}, nil
+	}
+	// Config values should always be vetted so there's no need to check the error from conversion.
+	duration, _ := ptypes.Duration(executionValues.QueueingBudget)
+
+	return QualityOfServiceSpec{
```
wild-endeavor, line 172: can we add a debug log line to each case so we can more easily understand what was used?

pkg/runtime/mocks/mock_quality_of_service_provider.go
```diff
@@ -0,0 +1,26 @@
+package mocks
```
wild-endeavor, line 1: why doesn't this file have those generated file comments? is this not generated?

flyteadmin_config.yaml
katrogan: it's not used so technically i can go make a breaking change :) any suggestions instead? basic/baseline?

pkg/runtime/mocks/mock_quality_of_service_provider.go
katrogan: nope, see https://github.com/lyft/flyte/issues/149

flyteadmin_config.yaml
katrogan: actually we've had bad luck with supposedly no-op breaking changes so I would rather not (this is checked into flyteidl, unfortunately)

pkg/manager/impl/executions/quality_of_service.go
katrogan: done

katrogan:
@migueltol22 @EngHabu

pkg/manager/impl/execution_manager_test.go
```diff
@@ -217,12 +218,33 @@ func TestCreateExecution(t *testing.T) {
 				"annotation3": "3",
 				"annotation4": "4",
 			}, inputs.Annotations)
+			assert.EqualValues(t, 10*time.Minute, inputs.QueueingBudget)
 			return &workflowengineInterfaces.ExecutionInfo{
 				Cluster: testCluster,
 			}, nil
 		})
+	//qosProvider := runtimeMocks.NewMockQualityOfServiceProvider()
```
EngHabu, line 226: nit: deadcode... delete?

katrogan: done, thanks

PR #109, opened 2020-07-24
Return full execution data on every request if under max specified size

katrogan:
# TL;DR
This change returns fully populated input & output data on GetExecutionData requests in addition to signed URLs.

Not to be merged until https://github.com/lyft/flyteidl/pull/73 is checked in & released

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
This is useful to return inputs & outputs in sandbox mode when the signed url is host-specific. Also fixes the leaky abstraction presented by requiring a caller to concern themselves with where the data stored and having to parse the proto on their own.

## Tracking Issue
https://github.com/lyft/flyte/issues/419

## Follow-up issue
_NA_
OR


pkg/manager/impl/execution_manager.go
```diff
@@ -1019,10 +1021,32 @@ func (m *ExecutionManager) GetExecutionData(
 	if err != nil {
 		return nil, err
 	}
-	return &admin.WorkflowExecutionGetDataResponse{
+	response := &admin.WorkflowExecutionGetDataResponse{
```
kumare3, line 1024: @katrogan one reason to do this is to avoid having to sign the URL, so can we just do a `HEAD`, check the size and if that is ok then do a protobuf get using flytestdlib.storage?

kumare3: or maybe i am misunderstanding, do we actually store the size of the payload in our database?

katrogan: until the UI transitions to using the new field, we should return both for backwards compatibility 
but good point on using flytestdlib, will update

katrogan: no, we don't store the size

katrogan: oh wait, and the storage client is using flytestdlib already

PR #114, opened 2020-08-13
Add UpdateProject functionality to Flyte control plane

kosigz-lyft:
# TL;DR
Add UpdateProject functionality to Flyte control plane.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Issue
https://github.com/lyft/flyte/issues/474

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added


EngHabu:
Can you create an issue in github.com/flyte/issues (it's public) and link all of these PRs to it?

pkg/repositories/gormimpl/project_repo.go
```diff
@@ -70,3 +70,33 @@ func NewProjectRepo(db *gorm.DB, errorTransformer errors.ErrorTransformer,
...
+		return models.Project{}, flyteAdminErrors.NewFlyteAdminErrorf(codes.NotFound, "project [%s] not found", updatedProject.Identifier)
+	}
+
+	// TODO: update the project with the fields and persist to the DB.
+	if updatedProject.Description != "" {
+		project.Description = updatedProject.Description;
+	}
+
+	if len(updatedProject.Labels) > 0 {
+		project.Labels = updatedProject.Labels;
```
kumare3, line 98: should we be combining the labels?

anandswaminathan: This is fine. But I have a question on overall update API - it's very tricky one.

In your current implementation, fields cannot be cleared through update. You design is only fields that need to be updated will be passed in ? 

@katrogan Is that model we follow? Or should the recommendation be - do a GET, and change field, and call update. Server does simple overwrite?

kumare3: great point

kumare3: but if we want to do the entity modification, then we will need to use like a update-tag?

kosigz-lyft: @schottra advised to use the existing admin.Project structure to construct the request.
> You design is only fields that need to be updated will be passed in.
Correct, if something like ID or Domain is passed on the admin.Project object it will be ignored.
> Or should the recommendation be - do a GET, and change field, and call update. Server does simple overwrite?
I'm not 100% sure what the difference between those two is. In principle, it shouldn't matter if a user of the API knows an ID they want to modify or if they first send a GET request, change the object, then send it back, unless I'm missing something.
In terms of how to actually do the DB write, that's something I'll need some help with since I don't know the API well.

pkg/manager/impl/project_manager.go
```diff
@@ -60,6 +60,19 @@ func (m *ProjectManager) ListProjects(ctx context.Context, request admin.Project
 	}, nil
 }
 
+func (m *ProjectManager) UpdateProject(ctx context.Context, project admin.Project) (*admin.ProjectUpdateResponse, error) {
+	var response *admin.ProjectUpdateResponse
+
+	projectModel := transformers.CreateProjectModel(&project);
```
katrogan, line 66: are we assuming a fully formed project in the request and not a patch type operation?

pkg/manager/impl/project_manager.go
```diff
@@ -60,6 +60,19 @@ func (m *ProjectManager) ListProjects(ctx context.Context, request admin.Project
...
+	_, err := m.db.ProjectRepo().UpdateProject(ctx, projectModel)
```
katrogan, line 67: why have the repo method return anything if it's unused?

pkg/repositories/gormimpl/project_repo.go
```diff
@@ -70,3 +70,33 @@ func NewProjectRepo(db *gorm.DB, errorTransformer errors.ErrorTransformer,
...
+	if tx.Error != nil {
+		return models.Project{}, r.errorTransformer.ToFlyteAdminError(tx.Error)
+	}
+
+	if tx.RecordNotFound() {
+		return models.Project{}, flyteAdminErrors.NewFlyteAdminErrorf(codes.NotFound, "project [%s] not found", updatedProject.Identifier)
+	}
+
+	// TODO: update the project with the fields and persist to the DB.
+	if updatedProject.Description != "" {
```
katrogan, line 93: can we move these checks to an appropriate transformer?

pkg/repositories/gormimpl/project_repo.go
```diff
@@ -70,3 +70,33 @@ func NewProjectRepo(db *gorm.DB, errorTransformer errors.ErrorTransformer,
 		metrics:          metrics,
 	}
 }
+
+func (r *ProjectRepo) UpdateProject(ctx context.Context, updatedProject models.Project) (models.Project, error) {
+	var project models.Project
+
+	timer := r.metrics.GetDuration.Start()
```
katrogan, line 77: ideally the get should be called at the manager layer

pkg/repositories/gormimpl/project_repo.go
katrogan: i think we use typically use Update (http://gorm.io/docs/update.html#Update-Changed-Fields) which should only update changed fields

anandswaminathan: @kosigz Then I think we would need to make all fields pointers, like say "Description"? 
@katrogan I was referring to the admin API.

katrogan: sorry @anandswaminathan i think the model we're following is using PUT (https://github.com/lyft/flyteidl/pull/78/files#diff-fde1983afefdc036089ed82714330441R320)

i think according to https://github.com/lyft/flyteidl/pull/78#discussion_r469471031 we don't allow for clearing fields in this model

question is - should we? in the sdk we allow for explicitly overriding things like notifications with an empty value

pkg/repositories/gormimpl/project_repo.go
kosigz-lyft: Sounds goods.

pkg/manager/impl/project_manager.go
kosigz-lyft: Not necessarily. We expect the ID and at least one of labels and description but there's no validation (yet). Domains are disregarded.

pkg/repositories/gormimpl/project_repo.go
kosigz-lyft: I'm following the structure of the `Get` operation defined in the same file here.

katrogan: sure but instead of duplicating the logic of the Get operation we could have the repo method caller (the project manager) call get and then update

katrogan:
Mind adding an integration test here: https://github.com/lyft/flyteadmin/blob/master/tests/project.go to validate the Update works to only overwrite set fields?

pkg/repositories/gormimpl/project_repo.go
kosigz-lyft: Got it, updated.

pkg/repositories/gormimpl/project_repo.go
```diff
@@ -70,3 +70,24 @@ func NewProjectRepo(db *gorm.DB, errorTransformer errors.ErrorTransformer,
 		metrics:          metrics,
 	}
 }
+
+func (r *ProjectRepo) UpdateProject(ctx context.Context, project models.Project, projectUpdate models.Project) (error) {
```
katrogan, line 74: since we're going with a merge patch type strategy i think you don't even need the existing project model to begin with (http://gorm.io/docs/update.html#Update-Changed-Fields)

tests/project.go
```diff
@@ -43,3 +43,92 @@ func TestCreateProject(t *testing.T) {
...
+func TestUpdateProjectDescription(t *testing.T) {
+	ctx := context.Background()
+	client, conn := GetTestAdminServiceClient()
+	defer conn.Close()
+
+	// Create a new project.
+	req := admin.ProjectRegisterRequest{
+		Project: &admin.Project{
+			Id:   "potato",
+			Name: "spud",
```
katrogan, line 56: nit: for completeness, want to add Labels here so we make sure that emtpy labels in the update request below doesn't delete them?

tests/project.go
```diff
@@ -43,3 +43,92 @@ func TestCreateProject(t *testing.T) {
 	}
 	assert.True(t, sawNewProject)
 }
+
+func TestUpdateProjectDescription(t *testing.T) {
```
katrogan, line 47: thank you for adding both unit tests! 😄

pkg/repositories/gormimpl/project_repo.go
kosigz-lyft: Question: also for @anandswaminathan and folks who had opinions — do we want to update fields even if the update is null? In that case, we could skip both the if branches in this function as well.

tests/project.go
```diff
@@ -43,3 +43,108 @@ func TestCreateProject(t *testing.T) {
...
+	assert.Nil(t, err)
+	assert.NotEmpty(t, projects.Projects)
+
+	// Attempt to modify the name of the Project. Modifying the Name should be a
+	// no-op, while the Description is modified. Labels should be a no-op.
+	_, err = client.UpdateProject(ctx, &admin.Project{
+		Id: "potato",
+		Name: "foobar",
+		Description: "a-new-description",
+		Labels: &admin.Labels{},
```
kosigz-lyft, line 79: @katrogan — according to our hypothesis about how gorm works, the update with empty labels should be a no-op.

tests/project.go
kosigz-lyft: Yup, added [here](https://github.com/lyft/flyteadmin/pull/114/files#r471760725).

kumare3:
@kosigz-lyft please add a link to the issue - https://github.com/lyft/flyte/issues/474

go.mod
```diff
@@ -12,7 +12,8 @@ require (
 	github.com/evanphx/json-patch v4.5.0+incompatible
 	github.com/gogo/protobuf v1.3.1
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
-	github.com/golang/protobuf v1.3.5
+	github.com/golang/protobuf v1.4.2
```
katrogan, line 15: why change the protobuf version?

go.mod
```diff
@@ -22,14 +23,14 @@ require (
...
-	github.com/lyft/flyteidl v0.18.2
+	github.com/lyft/flyteidl v0.18.3
 	github.com/lyft/flytepropeller v0.3.7
 	github.com/lyft/flytestdlib v0.3.9
 	github.com/magiconair/properties v1.8.1
 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.5.0
-	github.com/spf13/cobra v0.0.6
+	github.com/spf13/cobra v1.0.0
```
katrogan, line 33: why update this?

kosigz-lyft: Reverted.

go.mod
kosigz-lyft: Reverted.

pkg/repositories/config/migrations.go
```diff
@@ -251,4 +251,13 @@ var Migrations = []*gormigrate.Migration{
 			return tx.Model(&models.NodeExecution{}).DropColumn("parent_id").DropColumn("node_execution_metadata").Error
 		},
 	},
+	{
+		ID: "2020-08-17-labels-addition",
+		Migrate: func(tx *gorm.DB) error {
+			return tx.AutoMigrate(&models.NodeExecution{}).Error
```
katrogan, line 257: models.Project{} ?

pkg/repositories/config/migrations.go
```diff
@@ -251,4 +251,13 @@ var Migrations = []*gormigrate.Migration{
...
+		},
+		Rollback: func(tx *gorm.DB) error {
+			return tx.Model(&models.NodeExecution{}).DropColumn("labels").Error
```
katrogan, line 260: here too

pkg/repositories/gormimpl/project_repo.go
```diff
@@ -70,3 +70,15 @@ func NewProjectRepo(db *gorm.DB, errorTransformer errors.ErrorTransformer,
 		metrics:          metrics,
 	}
 }
+
+func (r *ProjectRepo) UpdateProject(ctx context.Context, projectUpdate models.Project) error {
+	// Use gorm client to update the two fields that are changed.
+	writeTx := r.db.Model(&projectUpdate).Update("Description", "Labels")
+
+	// Return error if applies.
+	if writeTx != nil {
```
katrogan, line 79:
```suggestion
	if writeTx.Error != nil {
```

pkg/repositories/gormimpl/project_repo.go
```diff
@@ -70,3 +70,15 @@ func NewProjectRepo(db *gorm.DB, errorTransformer errors.ErrorTransformer,
...
+		_ = r.errorTransformer.ToFlyteAdminError(writeTx.Error)
```
katrogan, line 80:
```suggestion
		return r.errorTransformer.ToFlyteAdminError(writeTx.Error)
```

pkg/repositories/gormimpl/project_repo.go
```diff
@@ -70,3 +70,15 @@ func NewProjectRepo(db *gorm.DB, errorTransformer errors.ErrorTransformer,
 		metrics:          metrics,
 	}
 }
+
+func (r *ProjectRepo) UpdateProject(ctx context.Context, projectUpdate models.Project) error {
+	// Use gorm client to update the two fields that are changed.
+	writeTx := r.db.Model(&projectUpdate).Update("Description", "Labels")
```
katrogan, line 76: i think this is not quite right (see http://gorm.io/docs/update.html#Update-Changed-Fields)
this translates to //// UPDATE projects SET Description='Labels'

can you do something similar to https://github.com/lyft/flyteadmin/blob/master/pkg/repositories/gormimpl/launch_plan_repo.go#L43

tests/project.go
```diff
@@ -43,3 +43,108 @@ func TestCreateProject(t *testing.T) {
...
+	assert.Nil(t, err)
+	assert.NotEmpty(t, projects.Projects)
+
+	// Attempt to modify the name of the Project. Modifying the Name should be a
+	// no-op, while the Description is modified. Labels should be a no-op.
+	_, err = client.UpdateProject(ctx, &admin.Project{
+		Id:          "potato",
+		Name:        "foobar",
+		Description: "a-new-description",
+		Labels:      &admin.Labels{},
```
katrogan, line 79: nit: no need to set to the empty value, golang does this by default

tests/project.go
```diff
@@ -43,3 +43,107 @@ func TestCreateProject(t *testing.T) {
 	}
 	assert.True(t, sawNewProject)
 }
+
+func TestUpdateProjectDescription(t *testing.T) {
```
katrogan, line 47: can you add a truncate projects table statement here? https://github.com/lyft/flyteadmin/blob/master/tests/bootstrap.go#L43 otherwise your tests are not idempotent

kosigz-lyft: Yup! Good callout. Had no idea that would be needed.

katrogan: np! it's not really intuitive unless you've worked with the code before

kosigz-lyft:
@katrogan thanks for helping me through it! There's no way I could have shipped this feature without your guidance. 😄

PR #117, opened 2020-08-24
Allow random cluster selection when no override

honnix:
# TL;DR
In case there is no override in resource table, we should allow the code to go through and apply non-label based weighted random selection.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
As [commented](https://github.com/lyft/flyteadmin/pull/71/files#r475657367), it makes more sense to let the code go through and apply different selection.

BTW these unit tests relies on `GOPATH` and we should fix that.

## Tracking Issue
_Will create later_
https://github.com/lyft/flyte/issues/<number>

## Follow-up issue
_NA_

### DIFF
--- a/pkg/executioncluster/impl/random_cluster_selector.go
+++ b/pkg/executioncluster/impl/random_cluster_selector.go
@@ -130,7 +130,6 @@ func (s RandomClusterSelector) GetTarget(ctx context.Context, spec *executionclu
 		if flyteAdminError, ok := err.(errors.FlyteAdminError); !ok || flyteAdminError.Code() != codes.NotFound {
 			return nil, err
 		}
-		return nil, err
 	}
 	var weightedRandomList random.WeightedRandomList
 	if resource != nil && resource.Attributes.GetExecutionClusterLabel() != nil {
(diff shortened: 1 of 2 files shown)


pkg/executioncluster/impl/random_cluster_selector.go
```diff
@@ -130,7 +130,6 @@ func (s RandomClusterSelector) GetTarget(ctx context.Context, spec *executionclu
 		if flyteAdminError, ok := err.(errors.FlyteAdminError); !ok || flyteAdminError.Code() != codes.NotFound {
 			return nil, err
 		}
-		return nil, err
```
EngHabu, line 133: I'm not comfortable with ignoring any error here... Can we be specific about the error that maps to "not found" or whatever... something along the lines of:
```
if err != nil {
		if flyteAdminError, ok := err.(errors.FlyteAdminError); !ok || flyteAdminError.Code() != codes.NotFound {
			return nil, err
		}

                if err != ErrNoOverride {
                    return nil, err
                }
}
```

anandswaminathan: @EngHabu We are only ignoring not found error here. Line 133 is a bug

honnix: I'm resolving this comment.

PR #119, opened 2020-08-31
Add slack validation

kosigz-lyft:
# TL;DR
Add the ability to validate that a tag is present on Flyte project creation.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

### DIFF
--- a/pkg/manager/impl/validation/project_validator.go
+++ b/pkg/manager/impl/validation/project_validator.go
@@ -34,7 +34,7 @@ func ValidateProjectRegisterRequest(request admin.ProjectRegisterRequest) error 
 		return err
 	}
 	// TODO: this is not where we should be hardcoding a required tag
-	tag := "slack"
+	tag := "foo"
 	if err := ValidateProjectHasTag(request.Project, tag) {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/manager/impl/validation/project_validator.go
```diff
@@ -33,6 +33,11 @@ func ValidateProjectRegisterRequest(request admin.ProjectRegisterRequest) error
 	if err := ValidateMaxLengthStringField(request.Project.Description, projectDescription, maxDescriptionLength); err != nil {
 		return err
 	}
+	// TODO: this is not where we should be hardcoding a required tag
+	tag := "foo"
```
kumare3, line 37: move to configuration, and also make it optional?

pkg/manager/impl/validation/validation.go
```diff
@@ -218,3 +218,12 @@ func ValidateLimit(limit uint32) error {
 	}
 	return nil
 }
+
+func ValidateProjectHasTag(project admin.Project, tag string) error {
+	if project.Labels != nil && project.Labels.Values != nil {
```
kumare3, line 223: if the tag is empty `""` then you will force users to have an empty lablel?

kosigz-lyft: No, fixing that.

PR #120, opened 2020-09-01
Add index to optimize for list task executions for node execution

katrogan:
# TL;DR
Add index to optimize for list task executions for node execution

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/328

## Follow-up issue
_NA_

### DIFF
--- a/pkg/repositories/config/migration_models.go
+++ b/pkg/repositories/config/migration_models.go
@@ -54,7 +54,10 @@ type NodeExecution struct {
 
 type TaskExecutionKey struct {
 	TaskKey
-	NodeExecutionKey
+	Project string `gorm:"primary_key;column:execution_project;index:idx_task_executions_exec"`
+	Domain  string `gorm:"primary_key;column:execution_domain;index:idx_task_executions_exec"`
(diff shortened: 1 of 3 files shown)


anandswaminathan:
@katrogan The gorm joins are fine? Has the latency gone down completely?

wild-endeavor:
What happens when you try this locally?  Do you see a performance bump?

Ideally, I feel like this is a gorm issue rather than a postgres issue, is there no way to skip those tables when not necessary?  The query that gorm is delivering to postgres has the tables right?

katrogan:
@anandswaminathan yeah, i'm seeing latency on the order of formerly seconds to 0.1 ms now

katrogan:
@wild-endeavor I was misguided in the original reference bug - the query optimizer does indeed handle eliding the joins. adding this index improved query performance significantly

wild-endeavor:
+1

PR #121, opened 2020-09-01
Gcs remote data

honnix:
# TL;DR
Add support of signed GCS URL.
This is a continuation of #81 because CI doesn't seem to work in forked repo.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Right now only AWS/S3 presigned URL is support. This PR adds support for GCS URL.

The signing part looks a bit funky, mainly due to GCP lib does not support signing URL without
materialised service account key. So this PR uses https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/signBlob directly.

`SigningPrincipal` refers to the service account that signs the URL, as well as reads object size. Service account associated with flyteadmin instance needs to have the [role](https://cloud.google.com/iam/docs/service-accounts#the_service_account_token_creator_role) on the principal.

## Follow-up issue
NA


pkg/data/implementations/gcp_remote_url.go
```diff
@@ -0,0 +1,156 @@
...
+func (g *GCPRemoteURL) Get(ctx context.Context, uri string) (admin.UrlBlob, error) {
+	logger.Debugf(ctx, "Getting signed url for - %s", uri)
+	gcsURI, err := g.splitURI(ctx, uri)
+	if err != nil {
+		logger.Debugf(ctx, "failed to extract gcs bucket and object from uri: %s", uri)
+		return admin.UrlBlob{}, errors.NewFlyteAdminErrorf(codes.InvalidArgument, "invalid uri: %s", uri)
+	}
+
+	// First, get the size of the url blob.
+	attrs, err := g.gcsClient.Bucket(gcsURI.bucket).Object(gcsURI.object).Attrs(ctx)
```
honnix, line 91: I couldn't figure out an easy way to mock `Bucket` because it returns a struct.

honnix: I think I'm still catching up with golang. Any suggestion @kumare3 ?

honnix: OK I did it in a pretty stupid way.

pkg/manager/impl/util/digests_test.go
```diff
@@ -46,8 +46,8 @@ var compiledTaskDigest = []byte{
 	0xa, 0x22, 0x80, 0xb1, 0x8, 0x44, 0x53, 0xf3, 0xca, 0x60, 0x4, 0xf7, 0x6f}
 
 var compiledWorkflowDigest = []byte{
-	0x1c, 0x66, 0x45, 0x89, 0x8f, 0x5, 0xa5, 0x4f, 0xf5, 0xba, 0x6f, 0xd9, 0xd, 0x70, 0xf6, 0x86, 0xf, 0x8e, 0x7e, 0x1b,
-	0x80, 0x1d, 0xb9, 0x59, 0xe, 0x3, 0x50, 0x4d, 0x64, 0xeb, 0x13, 0xa2}
+	0xeb, 0x66, 0x44, 0xe8, 0x1c, 0xa8, 0x51, 0x7d, 0x3f, 0x33, 0xf0, 0x77, 0x95, 0x24, 0x84, 0xc2, 0xbe, 0x79, 0xcd,
```
honnix, line 49: This is due to introducing of `google.golang.org/protobuf` as transitive dependency.

kumare3: interesting, @katrogan will this affect all existing digests to be invalidated, should be fine, but we might cause some re-registrations?

pkg/data/factory.go
```diff
@@ -45,6 +46,12 @@ func GetRemoteDataHandler(cfg RemoteDataHandlerConfig) RemoteDataHandler {
 		return &remoteDataHandler{
 			remoteURL: implementations.NewAWSRemoteURL(awsConfig, presignedURLDuration),
 		}
+	case common.GCP:
```
kumare3, line 49: @katrogan / @honnix with the new change to return the entire data as part of the getData API, we should probably think of making signing optional (we have to keep it around to ensure that very large datasets can still be served). By optional I mean, that if it is not specified, the signing should just be skipped. 

I do not think my comment needs to be added as part of this commit, but as a follow up? @katrogan what do you think?

kumare3 (dismissed):
One comment, but LGTM

pkg/manager/impl/util/digests_test.go
honnix: This is where I most concerned about as well. I can dig a bit further to see whether this can be avoided.

pkg/data/implementations/gcp_remote_url.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		logger.Debugf(ctx, "failed to get object size for %s with %v", uri, err)
+		return admin.UrlBlob{}, errors.NewFlyteAdminErrorf(
+			codes.Internal, "failed to get object size for %s with %v", uri, err)
+	}
+
+	// The second return argument here is the GetObjectOutput, which we don't use below.
```
katrogan, line 136: nit: is this comment relevant?

pkg/data/implementations/gcp_remote_url.go
```diff
@@ -0,0 +1,189 @@
...
+
+const gcsScheme = "gs"
+
+type iamCredentialsInterface interface {
+	SignBlob(ctx context.Context, req *credentialspb.SignBlobRequest, opts ...gax.CallOption) (*credentialspb.SignBlobResponse, error)
+	GenerateAccessToken(ctx context.Context, req *credentialspb.GenerateAccessTokenRequest, opts ...gax.CallOption) (*credentialspb.GenerateAccessTokenResponse, error)
+}
+
+type gcsClientWrapper struct {
+	delegate *gcs.Client
```
katrogan, line 30: hey @honnix I'm not super familiar with the delegate pattern, what's the reason for using it here?

katrogan: is this so you can mock out the bucket like you commented below?

pkg/data/implementations/gcp_remote_url.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func (g *GCPRemoteURL) signURL(ctx context.Context, gcsURI GCPGCSObject) (string, error) {
+	opts := &gcs.SignedURLOptions{
+		Method:         "GET",
+		GoogleAccessID: g.signingPrincipal,
+		SignBytes: func(b []byte) ([]byte, error) {
+			req := &credentialspb.SignBlobRequest{
+				Payload: b,
+				Name:    "projects/-/serviceAccounts/" + g.signingPrincipal,
```
katrogan, line 106: what's the meaning of this string prefix?

pkg/manager/impl/util/digests_test.go
katrogan: oh interesting! this could be a problem because some folks will get a warning about the digest changing for a workflow with the same identifier, so they will have to change the version parameter to reregister - which is unfortunate

pkg/data/implementations/gcp_remote_url.go
honnix: Oops, sorry. Copy pasta. Will fix.

pkg/data/implementations/gcp_remote_url.go
honnix: It's a GCP resource name: https://cloud.google.com/iam/docs/reference/rest/v1/projects.serviceAccounts/signBlob

Basically, it means the `service account` in some GCP project (`-`) that doesn't matter.

pkg/data/implementations/gcp_remote_url.go
honnix: Yes. That is the only way I could figure out how to test this fluid API.

go.mod
```diff
@@ -3,6 +3,8 @@ module github.com/lyft/flyteadmin
 go 1.13
 
 require (
+	cloud.google.com/go v0.56.0
```
honnix, line 6: This is the last version staying on protobuf v1.3.5. Since v1.4.0, `github.com/golang/protobuf` started to depend on `google.golang.org/protobuf`: https://github.com/golang/protobuf/blob/v1.4.0/ptypes/timestamp/timestamp.pb.go#L9

So sooner or later we will need to face this issue again.

pkg/manager/impl/util/digests_test.go
honnix: I managed to revert this change by downgrading a few deps, in latest commit. Still we will need to face this thing later: https://github.com/lyft/flyteadmin/pull/121#discussion_r482317790

go.mod
katrogan: oh this is awesome, thank you for finding it

katrogan (approved):

PR #122, opened 2020-09-02
Do not depend on GOPATH to locate test data

honnix:
# TL;DR
GOPATH is not mandatory to have now and the source code may not be stored under GOPATH.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Use 	`os.Getwd()` to get current directory and then join path.

## Tracking Issue
https://github.com/lyft/flyte/issues/#498

### DIFF
--- a/pkg/executioncluster/impl/random_cluster_selector_test.go
+++ b/pkg/executioncluster/impl/random_cluster_selector_test.go
@@ -2,10 +2,8 @@ package impl
 
 import (
 	"context"
-	"go/build"
 	"os"
 	"path/filepath"
-	"strings"
 	"testing"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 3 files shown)


pkg/executioncluster/impl/random_cluster_selector_test.go
```diff
@@ -32,13 +30,13 @@ const testDomain = "domain"
...
-	for _, goPath := range strings.Split(build.Default.GOPATH, string(os.PathListSeparator)) {
-		searchPaths = append(searchPaths, filepath.Join(goPath, "src/github.com/lyft/flyteadmin/pkg/executioncluster/testdata/", fileName))
+	pwd, err := os.Getwd()
+	if err != nil {
+		return err
 	}
 
 	configAccessor := viper.NewAccessor(config.Options{
-		SearchPaths: searchPaths,
+		SearchPaths: []string{filepath.Join(pwd, "../testdata", fileName)},
```
kumare3, line 39: Thank you so much, but what do you think about moving the test yaml to this folder?

kumare3: does not need to part of this change, we can file a bug and handle it as a follow up.

kumare3 (approved):
lgtm

pkg/executioncluster/impl/random_cluster_selector_test.go
honnix: Yeah I prefer that in a separated PR because it is not really part of this fix.

PR #124, opened 2020-09-08
#patch Allow unscoped filtering on all NamedEntity(Metadata) fields

schottra:
This fixes a regression where attempting to filter the list endpoint for NamedEntities on any of the base fields (such as `name`) will incorrectly exclude any records which do not have a metadata entry.

This is due to the generated query containing a `where` clause pointing at the metadata table, instead of the primary table.
For `NamedEntitiy` objects, the behavior we want is that the core and metadata fields are treated all as a single struct. The fact that we store the metadata in a separate table shouldn't require that filters be scoped.

This fixes the issue by leveraging some customization logic we already have for execution fields, but allows replacing of the filter `entity` instead. For `state` and `description`, we want the filter to be against the `NamedEntityMetadata` type. For all others, we want to use `NamedEntity`.

### DIFF
--- a/pkg/common/filters.go
+++ b/pkg/common/filters.go
@@ -81,7 +81,7 @@ var executionIdentifierFields = map[string]bool{
 
 var entityMetadataFields = map[string]bool{
 	"description": true,
-	"state": true,
+	"state":       true,
 }
 
 const unrecognizedFilterFunction = "unrecognized filter function: %s"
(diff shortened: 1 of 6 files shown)


pkg/common/filters.go
```diff
@@ -235,14 +240,24 @@ func customizeField(field string, entity Entity) string {
 	return field
 }
 
+func customizeEntity(field string, entity Entity) Entity {
+	// NamedEntity is considered a single object, but the metdata
+	// is stored using a different entity type.
+	if entity == NamedEntity && entityMetadataFields[field] {
```
katrogan, line 246: can you just add this check to if statement customizeField() above instead of creating a separate method for each entity type?

schottra: `customizeField` is changing the field, but not the entity. But I can update it to do both if that's what you mean?

katrogan: oh yes sorry! ignore me then, your call either way

katrogan (approved):

PR #126, opened 2020-09-21
Ensure admin never sets task resource request > limit

katrogan:
# TL;DR
In the case where a user specifies a task resource limit but not the request value, admin will substitute the request value with the platform-configured default. In some cases this default can exceed the user-specified limit which is non-sensical and prevents kubernetes from ever scheduling the task.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Fixes user-reported bug.

## Tracking Issue
https://github.com/lyft/flyte/issues/264

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/execution_manager.go
+++ b/pkg/manager/impl/execution_manager.go
@@ -282,7 +282,8 @@ func resolveTaskLimitsAndPlatformRequestDefaults(ctx context.Context, identifier
 		limitValue, ok := resourceLimits[resourceEntry.Name]
 		if !ok {
 			// Unexpected - at this stage both requests and limits should be populated.
-			logger.Warningf(ctx, "No limit specified for [%s] although request was set", identifier)
+			logger.Warningf(ctx, "No limit specified for [%v] resource [%s] although request was set", identifier,
+				resourceEntry.Name)
 			continue
 		}
 		if quantity.Cmp(resource.MustParse(limitValue)) == 1 {
--- a/pkg/manager/impl/execution_manager.go
+++ b/pkg/manager/impl/execution_manager.go
@@ -265,7 +265,7 @@ func assignResourcesIfUnset(ctx context.Context, identifier *core.Identifier,
 	return resourceEntries
 }
 
-func resolveTaskLimitsAndPlatformRequestDefaults(ctx context.Context, identifier *core.Identifier,
+func checkTaskRequestsLessThanLimits(ctx context.Context, identifier *core.Identifier,
 	taskResources *core.Resources) {
 	// We choose the minimum of the platform request defaults or the limit itself for every resource request.
 	// Otherwise we can find ourselves in confusing scenarios where the injected platform request defaults exceed a
@@ -342,7 +342,7 @@ func (m *ExecutionManager) setCompiledTaskDefaults(ctx context.Context, task *co
 	task.Template.GetContainer().Resources.Limits = assignResourcesIfUnset(
 		ctx, task.Template.Id, createTaskDefaultLimits(ctx, task), task.Template.GetContainer().Resources.Limits,
 		taskResourceSpec)
-	resolveTaskLimitsAndPlatformRequestDefaults(ctx, task.Template.Id, task.Template.GetContainer().Resources)
+	checkTaskRequestsLessThanLimits(ctx, task.Template.Id, task.Template.GetContainer().Resources)
 }
 
 func (m *ExecutionManager) launchSingleTaskExecution(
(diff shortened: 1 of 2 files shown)


pkg/manager/impl/execution_manager.go
```diff
@@ -263,6 +265,41 @@ func assignResourcesIfUnset(ctx context.Context, identifier *core.Identifier,
 	return resourceEntries
 }
 
+func resolveTaskLimitsAndPlatformRequestDefaults(ctx context.Context, identifier *core.Identifier,
```
EngHabu, line 268: I don't see the code updating the limits (as the name suggests?) am I missing something?

katrogan: nope, I just struggled with naming. is 'considerTaskLimitsAndPlatformRequestDefaults` or `reconcileTaskLimitsAndPlatformRequestDefaults` any better? other suggestions welcomed

wild-endeavor: the limit is not supposed to change though... the request is.

wild-endeavor (approved):
just add a comment to the function to clear up confusion?

pkg/manager/impl/execution_manager.go
katrogan: okay updated

katrogan:
friendly ping @EngHabu @wild-endeavor @bnsblue

pkg/manager/impl/execution_manager.go
```diff
@@ -263,6 +265,41 @@ func assignResourcesIfUnset(ctx context.Context, identifier *core.Identifier,
...
+		limitValue, ok := resourceLimits[resourceEntry.Name]
+		if !ok {
+			// Unexpected - at this stage both requests and limits should be populated.
+			logger.Warningf(ctx, "No limit specified for [%v] resource [%s] although request was set", identifier,
+				resourceEntry.Name)
+			continue
+		}
+		if quantity.Cmp(resource.MustParse(limitValue)) == 1 {
+			// The quantity is greater than the limit! Course correct below.
+			logger.Debugf(ctx, "Updating requested value for task [%+v] resource [%s]. Overriding to [%s] from [%s]",
```
bnsblue, line 291: nit: can you add what causes this override to the log so that it's clearer when reading the log?

bnsblue: Also I feel the debug level should be warning/info? Not quite sure 🤔

pkg/manager/impl/execution_manager_test.go
```diff
@@ -2464,6 +2464,111 @@ func TestAssignResourcesIfUnset(t *testing.T) {
...
+					Name:  core.Resources_CPU,
+					Value: "2",
+				},
+				{
+					Name:  core.Resources_MEMORY,
+					Value: "1.5",
+				},
+			},
+		}, resources))
+	})
```
bnsblue, line 2569: Can you add a test that tests this logic? Specifically, can you add a test where one of the resource does not have the limit specified? 
https://github.com/lyft/flyteadmin/pull/126/files#diff-fc047e54b9dd82ca7c89ac9b32cb07b3R282-R288

bnsblue:
LGTM. Just some nits.

pkg/manager/impl/execution_manager_test.go
katrogan: at the stage where the method is called all values will have already been substituted with the application defaults. this is why my test only verifies that if the defaults are weirdly configured we choose the min of acceptable values

bnsblue: oh interesting. So the control flow should never enter this block?
https://github.com/lyft/flyteadmin/pull/126/files#diff-fc047e54b9dd82ca7c89ac9b32cb07b3R283-R288

katrogan: unless the user is specifying a resource we don't provide defaults for. but in that case we don't find ourselves in the original conundrum where the substituted request > limit since there will be no limit default value which is a permissible situation for scheduling

pkg/manager/impl/execution_manager.go
katrogan: updated & upgraded

pkg/manager/impl/execution_manager_test.go
bnsblue: From a unit-test point of view I think that it would still be nice to have a test for it since we don't know if in the future this function would be invoked through other code paths where the default limits were not pre-filled. But as you said it is not critical here.

bnsblue (approved):

PR #129, opened 2020-10-01
Set labels in workflow metadata

catalinii:
https://github.com/lyft/flyte/issues/521

### DIFF
--- a/pkg/manager/impl/execution_manager.go
+++ b/pkg/manager/impl/execution_manager.go
@@ -467,6 +467,11 @@ func (m *ExecutionManager) launchSingleTaskExecution(
 	if request.Spec.Labels != nil {
 		executeTaskInputs.Labels = request.Spec.Labels.Values
 	}
+	executeTaskInputs.Labels, err = m.addProjectLabels(ctx, request.Project, executeTaskInputs.Labels)
+	if err != nil {
+		return nil, nil, err
+	}
+
 	if request.Spec.Annotations != nil {
 		executeTaskInputs.Annotations = request.Spec.Annotations.Values
 	}
@@ -632,6 +637,11 @@ func (m *ExecutionManager) launchExecutionAndPrepareModel(
 	if err != nil {
 		return nil, nil, err
 	}
+	executeWorkflowInputs.Labels, err = m.addProjectLabels(ctx, request.Project, executeWorkflowInputs.Labels)
+	if err != nil {
+		return nil, nil, err
+	}
+
 	err = m.addPluginOverrides(ctx, &workflowExecutionID, launchPlan.GetSpec().WorkflowId.Name, launchPlan.Id.Name,
 		&executeWorkflowInputs)
 	if err != nil {
@@ -1345,4 +1355,26 @@ func NewExecutionManager(
 		resourceManager:           resourceManager,
 		qualityOfServiceAllocator: executions.NewQualityOfServiceAllocator(config, resourceManager),
 	}
+}
+
+// Adds project labels with higher precedence to workflow labels. Project labels are ignored if a corresponding label is set on the workflow.
+func (m *ExecutionManager) addProjectLabels(ctx context.Context, projectName string, initialLabels map[string]string) (map[string]string, error) {
+	project, err := m.db.ProjectRepo().Get(ctx, projectName)
+	if err != nil {
+		logger.Errorf(ctx, "Failed to get project for [%+v] with error: %v", project, err)
+		return nil, err
+	}
+	// passing nil domain as not needed to retrieve labels
+	projectLabels := transformers.FromProjectModel(project, nil).Labels.GetValues()
+
+	if initialLabels == nil {
+		initialLabels = make(map[string]string)
+	}
+
+	for k, v := range projectLabels {
+		if _, ok := initialLabels[k]; !ok {
+			initialLabels[k] = v
+		}
+	}
+	return initialLabels, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/workflowengine/impl/propeller_executor.go
```diff
@@ -136,6 +136,7 @@ func (c *FlytePropeller) ExecuteWorkflow(ctx context.Context, input interfaces.E
 
 	labels := addMapValues(input.Labels, flyteWf.Labels)
 	flyteWf.Labels = labels
+	flyteWf.WorkflowMeta.ProjectLabels = labels
```
katrogan, line 139: why are we duplicating the labels? isn't passing them once sufficient?

pkg/manager/impl/execution_manager.go
```diff
@@ -619,6 +619,12 @@ func (m *ExecutionManager) launchExecutionAndPrepareModel(
 		return nil, nil, err
 	}
 
+	projectLabels := make(map[string]string)
+	project, err := m.db.ProjectRepo().Get(ctx, request.Project)
+	if err != nil {
```
katrogan, line 624: if the err is not nil, the project definitely won't be populated. we should bubble up the err and return in this case

pkg/manager/impl/execution_manager.go
```diff
@@ -619,6 +619,12 @@ func (m *ExecutionManager) launchExecutionAndPrepareModel(
...
+		projectLabels = project.labels.values
```
katrogan, line 625: you should call the transformer: https://github.com/lyft/flyteadmin/blob/eb9236eea8af44791d87c0f4113b92d0da32fa0f/pkg/repositories/transformers/project.go#L28 in order to get the deserialized label values

pkg/manager/impl/execution_manager.go
catalinii: addressed

katrogan (commented):
looks good! mind adding to the existing unit test: https://github.com/lyft/flyteadmin/blob/master/pkg/manager/impl/execution_manager_test.go#L199 to make sure we pass the project labels as expected?

pkg/manager/impl/execution_manager_test.go
```diff
@@ -198,6 +198,13 @@ func getMockRepositoryForExecTest() repositories.RepositoryInterface {
 
 func TestCreateExecution(t *testing.T) {
 	repository := getMockRepositoryForExecTest()
+	repository.ProjectRepo().Create(context.Background(), models.Project{
```
katrogan, line 201: I think you need to mock out the Get call to return a project with the proper labels

catalinii: fixed, tested and working:
Running tool: /usr/local/opt/go/libexec/bin/go test -timeout 30s github.com/lyft/flyteadmin/pkg/manager/impl -run ^TestCreateExecution$

ok  	github.com/lyft/flyteadmin/pkg/manager/impl	0.503s

katrogan:
to fix lint can you run `make goimports` and then validate with `make lint`

go.mod
```diff
@@ -25,8 +25,8 @@ require (
 	github.com/jinzhu/gorm v1.9.12
 	github.com/kelseyhightower/envconfig v1.4.0 // indirect
 	github.com/lib/pq v1.3.0
-	github.com/lyft/flyteidl v0.18.6
-	github.com/lyft/flytepropeller v0.3.17
+	github.com/lyft/flyteidl v0.18.7
+	github.com/lyft/flytepropeller v0.4.2-0.20201001203502-c9d4b2a9b8dd
```
katrogan, line 29: let's wait until the propeller changes are in (with an official release or tag) before merging this

catalinii: sure

pkg/manager/impl/execution_manager_test.go
```diff
@@ -198,6 +199,17 @@ func getMockRepositoryForExecTest() repositories.RepositoryInterface {
 
 func TestCreateExecution(t *testing.T) {
 	repository := getMockRepositoryForExecTest()
+	labels := admin.Labels{
+		Values: map[string]string{
+			"label3": "3",
+			"label2": "1", // common label, will be dropped
```
katrogan, line 205: what do you mean by this comment? I see it in the expected values below?

katrogan, line 205: do we want project labels to override individual labels?

catalinii: They do not override, see the code: https://github.com/lyft/flyteadmin/pull/129/files#diff-fc047e54b9dd82ca7c89ac9b32cb07b3R1375

katrogan: ah thank you for clarifying!

katrogan (approved):

pkg/manager/impl/execution_manager.go
```diff
@@ -1346,3 +1356,25 @@ func NewExecutionManager(
 		qualityOfServiceAllocator: executions.NewQualityOfServiceAllocator(config, resourceManager),
 	}
 }
+
+func (m *ExecutionManager) addProjectLabels(ctx context.Context, projectName string, initialLabels map[string]string) (map[string]string, error) {
```
kumare3, line 1360: I know you have a comment below, but comments on the func become docs in golang.
So a comment like
// Adds project labels with higher precedence to workflow labels. Project labels are ignored if a corresponding label is set on the workflow.

kumare3:
thank you looks good to me, just one comment and then +1

pkg/manager/impl/execution_manager.go
catalinii: @kumare3 I addressed the comment

kumare3 (approved):

PR #130, opened 2020-10-08
Fix list executions for single task

katrogan:
# TL;DR
In the case of single task executions the reference `launch_plan_id` in the executions database model is 0. Likewise, for conventional executions the `task_id` is set to 0 and the launch plan is referenced instead. This poses a problem because we always inner join the executions table with the workflows and launch plans table (and as of this PR, now tasks too) when listing executions to permit filtering on these join table attributes. This change updates the joins to only be included when specific filters referencing these joinable entities exist.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/542

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/execution_manager.go
+++ b/pkg/manager/impl/execution_manager.go
@@ -1153,11 +1153,16 @@ func (m *ExecutionManager) ListExecutions(
 		return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument, "invalid pagination token %s for ListExecutions",
 			request.Token)
 	}
+	joinTableEntities := make(map[common.Entity]bool)
+	for _, filter := range filters {
+		joinTableEntities[filter.GetEntity()] = true
+	}
 	listExecutionsInput := repositoryInterfaces.ListResourceInput{
-		Limit:         int(request.Limit),
-		Offset:        offset,
-		InlineFilters: filters,
-		SortParameter: sortParameter,
+		Limit:             int(request.Limit),
+		Offset:            offset,
+		InlineFilters:     filters,
+		SortParameter:     sortParameter,
+		JoinTableEntities: joinTableEntities,
 	}
 	output, err := m.db.ExecutionRepo().List(ctx, listExecutionsInput)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/repositories/interfaces/common.go
```diff
@@ -20,8 +20,9 @@ type ListResourceInput struct {
 	// MapFilters refers to primary entity filters defined as map values rather than inline sql queries.
 	// These exist to permit filtering on "IS NULL" which isn't permitted with inline filter queries and
 	// pq driver value substitution.
-	MapFilters    []common.MapFilter
-	SortParameter common.SortParameter
+	MapFilters        []common.MapFilter
+	SortParameter     common.SortParameter
+	JoinTableEntities map[common.Entity]bool
```
kumare3, line 25: can we get some documentation here, that explain what this is for?

pkg/manager/impl/execution_manager.go
```diff
@@ -1153,11 +1153,16 @@ func (m *ExecutionManager) ListExecutions(
 		return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument, "invalid pagination token %s for ListExecutions",
 			request.Token)
 	}
+	joinTableEntities := make(map[common.Entity]bool)
+	for _, filter := range filters {
```
kumare3, line 1157: how is this filter specified? One thing I have been struggling with is filters documentation. Can you please point me to one or an example?

schottra (dismissed):

pkg/manager/impl/execution_manager.go
katrogan: docs here: https://docs.lyft.net/eng/flytedocs/contributor/components/admin.html#making-requests-to-flyteadmin let me know if i can move this somewhere more discoverable

pkg/repositories/interfaces/common.go
katrogan: done

katrogan:
can I get another +1 @schottra @kumare3?

pkg/manager/impl/execution_manager.go
kumare3: Ohh my bad, should always know Katrina has docs. Stand corrected

kumare3 (approved):

PR #132, opened 2020-10-23
Validate Project Requests on Create/Update

narape:
# TL;DR
Increase the validation done on `CreateProject` and `UpdateProject` in `project_manager.go`

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Increase the validation done on `CreateProject` and `UpdateProject` in `project_manager.go`:
* Validate `admin.Project` on `UpdateProject`
* Add length validation for `project-name`. It cannot contain more than 64 characters
* Add validation for number entries in `labels`. It cannot contain more than 16 labels

## Tracking Issue
https://github.com/lyft/flyte/issues/336

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/project_manager.go
+++ b/pkg/manager/impl/project_manager.go
@@ -70,9 +70,8 @@ func (m *ProjectManager) UpdateProject(ctx context.Context, projectUpdate admin.
 		return nil, err
 	}
 
-	// Run validation on the request, specifically checking for labels, and return err if validation does not succeed.
-	err = validation.ValidateProjectLabels(projectUpdate)
-	if err != nil {
+	// Run validation on the request and return err if validation does not succeed.
+	if err := validation.ValidateProject(projectUpdate); err != nil {
 		return nil, err
 	}
 
--- a/pkg/manager/impl/validation/validation.go
+++ b/pkg/manager/impl/validation/validation.go
@@ -34,6 +34,14 @@ func ValidateMaxLengthStringField(field string, fieldName string, limit int) err
 	return nil
 }
 
+// Validates that a map field does not exceed a certain amount of entries
+func ValidateMaxMapLengthField(m map[string]string, fieldName string, limit int) error {
+	if len(m) > limit {
+		return errors.NewFlyteAdminErrorf(codes.InvalidArgument, "%s map cannot exceed %d entries", fieldName, limit)
+	}
+	return nil
+}
+
 func ValidateIdentifierFieldsSet(id *core.Identifier) error {
 	if id == nil {
 		return shared.GetMissingArgumentError(shared.ID)
--- a/pkg/repositories/config/migrations.go
+++ b/pkg/repositories/config/migrations.go
@@ -199,7 +199,7 @@ var Migrations = []*gormigrate.Migration{
 			return tx.Exec("ALTER TABLE workflows ADD COLUMN IF NOT EXISTS state integer;").Error
 		},
 	},
-	// Modify the executions & node_executison table, if necessary
+	// Modify the executions & node_execution table, if necessary
 	{
 		ID: "2020-04-29-executions",
 		Migrate: func(tx *gorm.DB) error {
--- a/pkg/manager/impl/project_manager_test.go
+++ b/pkg/manager/impl/project_manager_test.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/lyft/flyteadmin/pkg/manager/impl/shared"
+
 	"github.com/lyft/flyteadmin/pkg/common"
 	"github.com/lyft/flyteadmin/pkg/manager/impl/testutils"
 	repositoryMocks "github.com/lyft/flyteadmin/pkg/repositories/mocks"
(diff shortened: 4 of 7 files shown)


EngHabu:
Hi @narape, thank you for starting this. Is this PR ready for review? or are you planning to send it once you implement the extra validation? if it's the latter, I would suggest you mark it as "Draft" until ready...

narape:
I created this PR to start the conversation. 
Right now, flyte-admin does the following validations:
* project-id must be a DNS 1123 label
* project-name must not be empty *but there is no limit for how long it could be*
* project-description shouldn't be longer than 300 characters
* All labels must be valid k8s labels but there is no limit on how many elements the array could have*

@katrogan / @kumare3 Do we want to limit name length and label array length? if so what will be the limit for those

katrogan:
I think it'd be nice to transparently offload too large free-form text fields to a blobstore. But that still leaves the question, what is a reasonable limit for text serialized directly in the db? I know Postgres already does something similar with [TOAST](https://www.postgresql.org/docs/8.1/storage-toast.html) however for entities with high read volume (so perhaps, not projects :) ) that still looks potentially expensive from a latency perspective.

@kumare3 @EngHabu any thoughts on what a 'reasonable' limit for project attribute fields could be if we wanted to enforce it?

kumare3:
> I think it'd be nice to transparently offload too large free-form text fields to a blobstore. But that still leaves the question, what is a reasonable limit for text serialized directly in the db? I know Postgres already does something similar with [TOAST](https://www.postgresql.org/docs/8.1/storage-toast.html) however for entities with high read volume (so perhaps, not projects :) ) that still looks potentially expensive from a latency perspective.
> 
> @kumare3 @EngHabu any thoughts on what a 'reasonable' limit for project attribute fields could be if we wanted to enforce it?

Ya these should not be offloaded. I think I want the offloading to part of the documentation story. 
These direct entities should have small manageable payload and maintain performance

narape:
@kumare3 Could you compute p75, p90 and p99 for project-name length and for number of labels among the projects at Lyft to have those numbers as reference?

kumare3:
@narape we can and share those numbers

katrogan:
hey @narape this is difficult to do for labels b/c we store them as serialized bytes but for name and description:

```
 select percentile_disc(0.75) within group (order by length(projects.name)) as p75, percentile_disc(0.9) within group (order by length(projects.name)) as p90,  percentile_disc(0.99) within group (order by length(projects.name)) as p99 from projects;
 p75 | p90 | p99 
-----+-----+-----
  19 |  24 |  39


 select percentile_disc(0.75) within group (order by length(projects.description)) as p75, percentile_disc(0.9) within group (order by length(projects.description)) as p90,  percentile_disc(0.99) within group (order by length(projects.description)) as p99 from projects;
 p75 | p90 | p99 
-----+-----+-----
  47 |  67 | 150

```

narape:
@kumare3 @katrogan I went ahead and picked 64 for name and 16 for number of entries in the map. Also do you know how can I fix the other build checks? they failed when trying to push a docker image to github registry

katrogan:
hey @narape can you follow these instructions to try the tests out in a fork? https://github.com/lyft/flyteadmin/tree/master/boilerplate/lyft/github_workflows

@EngHabu says he's working with @evalsocket to make this process less of a pain in the future :)

in the meantime, he adds:

> go to your fork -> settings -> secrets and create a secret: `FLYTEGITHUB_REPO` with `<fork>/flyteadmin` as the value
Create a PR from your branch `admin-check-limits` into the own fork's master, this should trigger all the checks and you should have permissions to run them all

mind linking your fork PR once it's ready? sorry again for the extra trouble

narape:
@katrogan now my PR is working (thanks to the work done by @EngHabu) can you review this? Specially about the arbitrary limit that I chose

README.rst
```diff
@@ -10,7 +10,7 @@ Flyteadmin uses a relational style Metadata Store abstracted by `GORM <http://go
 Before Check-In
 ~~~~~~~~~~~~~~~
 
-Flyte Admin has a few useful make targets for linting and testing. Please use these before checking in to help suss out
```
katrogan, line 13: mind reverting this? i think it makes more sense as the original :)

pkg/manager/impl/validation/validation.go
```diff
@@ -34,6 +34,14 @@ func ValidateMaxLengthStringField(field string, fieldName string, limit int) err
 	return nil
 }
 
+// Validates that a string field does not exceed a certain character count
```
katrogan, line 37: nit: update comment?

pkg/repositories/config/migrations.go
```diff
@@ -199,7 +199,7 @@ var Migrations = []*gormigrate.Migration{
 			return tx.Exec("ALTER TABLE workflows ADD COLUMN IF NOT EXISTS state integer;").Error
 		},
 	},
-	// Modify the executions & node_executison table, if necessary
+	// Modify the executions & node_execution table, if necessary
```
katrogan, line 202: thank you for fixing!

pkg/repositories/models/project.go
```diff
@@ -3,7 +3,7 @@ package models
 type Project struct {
 	BaseModel
 	Identifier  string `gorm:"primary_key"`
-	Name        string // Human-readable name, not a unique identifier.
+	Name        string `gorm:"type:varchar(64)"` // Human-readable name, not a unique identifier.
```
katrogan, line 6: can we leave the underlying db model as is? i'd like to make validation part of the admin business logic layer so that if we need to change the limits we can easily do that without requiring a db op

katrogan:
the limits you selected look good! we can always iterate based on user feedback too if they end up being onerous

README.rst
narape: Haha, I thought that it was a typo. TIL

katrogan (approved):
Thank you so much for adding all the thorough test cases too!

katrogan (approved):

pkg/manager/impl/project_manager.go
```diff
@@ -70,9 +70,8 @@ func (m *ProjectManager) UpdateProject(ctx context.Context, projectUpdate admin.
 		return nil, err
 	}
 
-	// Run validation on the request, specifically checking for labels, and return err if validation does not succeed.
-	err = validation.ValidateProjectLabels(projectUpdate)
-	if err != nil {
+	// Run validation on the request and return err if validation does not succeed.
+	if err := validation.ValidateProject(projectUpdate); err != nil {
```
honnix, line 74: This breaks archiving/activating project feature unfortunately because in that case, project.Name and project.Description is not required.

honnix: I think for any update operation, it is reasonable to allow certain fields to be unset.

katrogan: +1, we should only do a merge update in that case and not override empty fields

PR #135, opened 2020-10-27
Enable EventVersion '1' in Admin

anandswaminathan:
This change provides ability to enable new eventVersion to be set on workflows.

Issue: https://github.com/lyft/flyte/issues/218
PTAL @kumare3 @katrogan

### DIFF
--- a/pkg/runtime/interfaces/application_configuration.go
+++ b/pkg/runtime/interfaces/application_configuration.go
@@ -45,6 +45,8 @@ type ApplicationConfig struct {
 	// This defines the nested path on the configured external storage provider where workflow closures are remotely
 	// offloaded.
 	MetadataStoragePrefix []string `json:"metadataStoragePrefix"`
+	// Event version to be used for Flyte workflows
+	EventVersion int `json:"eventVersion"`
 }
 
 // This section holds common config for AWS
--- a/pkg/rpc/adminservice/base.go
+++ b/pkg/rpc/adminservice/base.go
@@ -86,7 +86,7 @@ func NewAdminServer(kubeConfig, master string) *AdminService {
 		applicationConfiguration.RoleNameKey,
 		execCluster,
 		adminScope.NewSubScope("executor").NewSubScope("flytepropeller"),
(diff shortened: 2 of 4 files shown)


EngHabu (approved):

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -45,6 +45,8 @@ type ApplicationConfig struct {
 	// This defines the nested path on the configured external storage provider where workflow closures are remotely
 	// offloaded.
 	MetadataStoragePrefix []string `json:"metadataStoragePrefix"`
+	// Event version to be used for Flyte workflows
+	EventVersion int `json:"eventVersion"`
```
katrogan, line 49: why is this set in flyteadmin config and not from the events source (the propeller deployment)?

katrogan:
can you link to a github issue in the PR description?

pkg/runtime/interfaces/application_configuration.go
anandswaminathan: @katrogan The propeller change is already done. The eventVersion is determined per workflow. This change is essentially for all the new workflows created. All the old workflows do not have events anymore, and are already stored in Admin.

katrogan: > All the old workflows do not have events anymore.

What? What does this mean?

anandswaminathan: Events are only sent when the workflows are not finished. So old workflows will not have events anymore sent from Propeller. This will not affect them.

katrogan: but we send events on workflow completion, right?

anandswaminathan: Yup. We will default to 1, from the next release. @katrogan The version is pinned to the execution. So running/completing executions does not matter. Only newly created executions will be affected. Also if you revert, you are still fine

katrogan (approved):

PR #138, opened 2020-11-04
Archive and unarchive a project

honnix:
# TL;DR
Make a project achievable. Depends on https://github.com/lyft/flyteidl/pull/88

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
This PR makes it possible to archive and unarchive a project. Archived project is not visible for `list` but visible for `get`, and it is treated as an invalid project. Archived project is possible to update so that it can be unarchived.

@kumare3 PTAL and see whether this is on the right track. Thanks a lot.

## Tracking Issue
https://github.com/lyft/flyte/issues/569

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/testutils/repository.go
+++ b/pkg/manager/impl/testutils/repository.go
@@ -14,7 +14,7 @@ func GetRepoWithDefaultProjectAndErr(err error) repositories.RepositoryInterface
 	repo := repositoryMocks.NewMockRepository()
 	repo.ProjectRepo().(*repositoryMocks.MockProjectRepo).GetFunction = func(
 		ctx context.Context, projectID string) (models.Project, error) {
-		activeState := int32(admin.ProjectState_PROJECT_ACTIVE)
+		activeState := int32(admin.Project_ACTIVE)
 		return models.Project{State: &activeState}, err
 	}
 	return repo
@@ -24,7 +24,7 @@ func GetRepoWithDefaultProject() repositories.RepositoryInterface {
 	repo := repositoryMocks.NewMockRepository()
 	repo.ProjectRepo().(*repositoryMocks.MockProjectRepo).GetFunction = func(
 		ctx context.Context, projectID string) (models.Project, error) {
-		activeState := int32(admin.ProjectState_PROJECT_ACTIVE)
+		activeState := int32(admin.Project_ACTIVE)
 		return models.Project{State: &activeState}, nil
 	}
 	return repo
--- a/pkg/manager/impl/testutils/repository.go
+++ b/pkg/manager/impl/testutils/repository.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/lyft/flyteadmin/pkg/repositories/models"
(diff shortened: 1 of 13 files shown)


pkg/manager/impl/testutils/repository.go
```diff
@@ -13,7 +14,8 @@ func GetRepoWithDefaultProjectAndErr(err error) repositories.RepositoryInterface
 	repo := repositoryMocks.NewMockRepository()
 	repo.ProjectRepo().(*repositoryMocks.MockProjectRepo).GetFunction = func(
 		ctx context.Context, projectID string) (models.Project, error) {
-		return models.Project{}, err
+		activeState := int32(admin.Project_ACTIVE)
+		return models.Project{State: &activeState}, err
```
honnix, line 18: I'm not so sure here. Do we usually assume migration will always happen so `state` will always exist? I have to do this because point deference `if *project.State != int32(admin.Project_ACTIVE)`

katrogan (commented):
sorry I should have caught this in the IDL PR - we probably want to update the ProjectListRequest endpoint to accept a filter param so we can select active only

pkg/manager/impl/testutils/repository.go
katrogan: we usually run the migrations script as part of the admin initcontainers so I think it is safe to assume you can dereference. this test data anyways, so it should use the latest model definition :)

honnix:
> sorry I should have caught this in the IDL PR - we probably want to update the ProjectListRequest endpoint to accept a filter param so we can select active only

No worries. I didn't plan to change API in this PR, to make it a bit smaller. :) I will do that in a follow-up PR. Thanks for pointing it out.

pkg/repositories/gormimpl/project_repo.go
```diff
@@ -50,7 +51,7 @@ func (r *ProjectRepo) Get(ctx context.Context, projectID string) (models.Project
 
 func (r *ProjectRepo) ListAll(ctx context.Context, sortParameter common.SortParameter) ([]models.Project, error) {
 	var projects []models.Project
-	var tx = r.db
+	var tx = r.db.Where("state != ?", int32(admin.Project_ACTIVE))
```
kumare3, line 54: do you mean the inverse of this condition?

honnix: :facepalm: Good catch. Copy pasta at its peak, I even "fixed" test case to make it work.

katrogan:
hey @honnix this looks good but seems like there are some build issues?

honnix:
@katrogan Yeah we need https://github.com/lyft/flyteidl/pull/88 to be released first.

katrogan:
@honnix oh right, do you have permission to draft a release? otherwise i can go ahead and create one

honnix:
@katrogan I have permission and I will do that. Just wanted to confirm it is OK I make a release.

katrogan (approved):

PR #140, opened 2020-11-13
Add single task execution create request validation

katrogan:
# TL;DR
Because single task executions don't use launch plans, some parameters that are optional overrides for conventional ExecutionCreateRequests are actually mandatory.

Note: the reference entity used to launch an execution is confusingly named LaunchPlan but accepts identifiers of task resouce_type.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/567

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/validation/execution_validator.go
+++ b/pkg/manager/impl/validation/execution_validator.go
@@ -59,6 +59,11 @@ func ValidateExecutionRequest(ctx context.Context, request admin.ExecutionCreate
 			"Invalid reference entity resource type [%v], only [%+v] allowed",
 			request.Spec.LaunchPlan.ResourceType, acceptedReferenceLaunchTypes)
 	}
+	if request.Spec.LaunchPlan.ResourceType == core.ResourceType_TASK {
+		if err := validateLaunchSingleTaskExecutionReq(request); err != nil {
+			return err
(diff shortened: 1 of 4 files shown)


pkg/manager/impl/validation/execution_validator.go
```diff
@@ -157,3 +162,12 @@ func ValidateWorkflowExecutionIdentifier(identifier *core.WorkflowExecutionIdent
 	}
 	return nil
 }
+
+// Because single task executions don't use launch plans, some parameters that are optional overrides for conventional
+// ExecutionCreateRequests are actually mandatory.
+func validateLaunchSingleTaskExecutionReq(request admin.ExecutionCreateRequest) error {
+	if request.Spec.AuthRole == nil {
```
EngHabu, line 169: Should this check inside the AuthRole as well? like `AuthRole.Method.....`

katrogan: excellent point, done

EngHabu (approved):

PR #142, opened 2020-11-24
Add filter and sort param when listing projects

honnix:
# TL;DR
Add filter and sort param when listing projects.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
This PR depends on https://github.com/lyft/flyteidl/pull/90

This revised API endpoint support filtering and sorting, as well as pagination. Everything is optional to keep backward compatibility.

* Offset defaults to `0` if no `token` provided
* Limit will not be set if not provided
* Filter defaults to `state != 1` to filter out archived projects
* Sort defaults to `identifier asc` if not provided
* ~The old list function is kept to be used by sync and we can of course switch to the new one~ Replaced

## Tracking Issue
https://github.com/lyft/flyte/issues/625

## Follow-up issue
_NA_

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -48,7 +48,7 @@ jobs:
           key: /tmp/docker-images-${{ github.event.after }}
       - name: Prime docker cache
         run: docker load -i /tmp/tmp/docker-images/snapshot.tar || true
-      - uses: engineerd/setup-kind@v0.4.0
+      - uses: engineerd/setup-kind@v0.5.0
       - name: End2End
         run: |
           kubectl cluster-info
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/jinzhu/gorm v1.9.12
 	github.com/kelseyhightower/envconfig v1.4.0 // indirect
 	github.com/lib/pq v1.3.0
-	github.com/lyft/flyteidl v0.18.10
+	github.com/lyft/flyteidl v0.18.11
 	github.com/lyft/flytepropeller v0.3.17
 	github.com/lyft/flytestdlib v0.3.9
 	github.com/magiconair/properties v1.8.1
@@ -58,5 +58,4 @@ replace (
 	k8s.io/api => github.com/lyft/api v0.0.0-20191031200350-b49a72c274e0
 	k8s.io/apimachinery => github.com/lyft/apimachinery v0.0.0-20191031200210-047e3ea32d7f
 	k8s.io/client-go => k8s.io/client-go v0.0.0-20191016111102-bec269661e48
-
 )
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/jinzhu/gorm v1.9.12
 	github.com/kelseyhightower/envconfig v1.4.0 // indirect
 	github.com/lib/pq v1.3.0
-	github.com/lyft/flyteidl v0.18.11
+	github.com/lyft/flyteidl v0.18.11-0.20201124224632-b7d8ea8378ab
 	github.com/lyft/flytepropeller v0.3.17
 	github.com/lyft/flytestdlib v0.3.9
 	github.com/magiconair/properties v1.8.1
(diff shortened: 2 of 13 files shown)


pkg/repositories/gormimpl/project_repo.go
```diff
@@ -62,6 +62,41 @@ func (r *ProjectRepo) ListAll(ctx context.Context, sortParameter common.SortPara
 	return projects, nil
```
katrogan, line 62: could we update the call to ListAll in cluster resource controller to use the new List and delete the ListAll method if it's not used anywhere else? and also add a filter for non-archived projects in that call too?

honnix: Yeah the last commit does that but I haven't deleted the function and mock. Now I see this makes sense so I can move on cleaning up. Thanks.

.github/workflows/master.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           key: /tmp/docker-images-${{ github.event.after }}
       - name: Prime docker cache
         run: docker load -i /tmp/tmp/docker-images/snapshot.tar || true
-      - uses: engineerd/setup-kind@v0.4.0
+      - uses: engineerd/setup-kind@v0.5.0
```
honnix, line 51: https://github.com/engineerd/setup-kind/releases/tag/v0.5.0

go.mod
```diff
@@ -25,7 +25,7 @@ require (
 	github.com/jinzhu/gorm v1.9.12
 	github.com/kelseyhightower/envconfig v1.4.0 // indirect
 	github.com/lib/pq v1.3.0
-	github.com/lyft/flyteidl v0.18.10
+	github.com/lyft/flyteidl v0.18.11-0.20201124224632-b7d8ea8378ab
```
honnix, line 28: Remove commit and point to v0.18.11 after flyteidl PR merged and released. @honnix

pkg/clusterresource/controller.go
```diff
@@ -357,7 +358,9 @@ func (c *controller) Sync(ctx context.Context) error {
 	logger.Debugf(ctx, "Running an invocation of ClusterResource Sync")
 
 	// Prefer to sync projects most newly created to ensure their resources get created first when other resources exist.
-	projects, err := c.db.ProjectRepo().ListAll(ctx, descCreatedAtSortParam)
+	projects, err := c.db.ProjectRepo().List(ctx, repositoriesInterfaces.ListResourceInput{
+		SortParameter: descCreatedAtSortParam,
```
katrogan, line 362: nit: as an optimization can we add a filter for active projects (now that we can filter!)

pkg/repositories/interfaces/project_repo.go
```diff
@@ -12,8 +11,8 @@ type ProjectRepoInterface interface {
 	Create(ctx context.Context, project models.Project) error
 	// Returns a matching project when it exists.
 	Get(ctx context.Context, projectID string) (models.Project, error)
-	// Lists unique projects registered as namespaces
-	ListAll(ctx context.Context, sortParameter common.SortParameter) ([]models.Project, error)
+	// Returns projects matching query parameters. A limit must be provided for the results page size.
```
katrogan, line 14: is this true?

> A limit must be provided

katrogan (commented):
looks good! mind updating the flyteidl version in go.mod?

pkg/repositories/interfaces/project_repo.go
honnix: Good catch. It is a copy paste mistake.

pkg/clusterresource/controller.go
honnix: Sure we can provider a filter here. Filtering out non-active projects is the default behavior when no filter provided, to keep backward compatibility.

honnix: This is addressed by https://github.com/lyft/flyteadmin/pull/142/commits/bf478e5936fdfebc23d530334ae03df9028c7664

katrogan: oh you're right, sorry i thought it was done in the manager layer for some reason. thank you for adding regardless, my bad!

katrogan (approved):

PR #145, opened 2020-12-01
Be more generous validating update project

honnix:
# TL;DR
https://github.com/lyft/flyteadmin/pull/132 added support for validating projects for both create and update. However when updating a project, it is reasonable to have some fields unset (empty string coming from gRPC). In case of archiving/activating a project, only project ID and state are required.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Executing for example ` flyte-cli archive-project -i -h <host:port> -p some-project`  can easily reproduce.

This PR skips validating empty string for `project.Name` when it is an update operation. Actually even for a create operation, it should be OK to leave `project.Name` unset, but to keep backward compatibility, this is still enforced.

## Tracking Issue
https://github.com/lyft/flyte/issues/336

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/validation/project_validator.go
+++ b/pkg/manager/impl/validation/project_validator.go
@@ -25,7 +25,11 @@ func ValidateProjectRegisterRequest(request admin.ProjectRegisterRequest) error 
 	if request.Project == nil {
 		return shared.GetMissingArgumentError(shared.Project)
 	}
-	return ValidateProject(*request.Project)
+	project := *request.Project
+	if err := ValidateEmptyStringField(project.Name, projectName); err != nil {
+		return err
+	}
+	return ValidateProject(project)
 }
 
 func ValidateProject(project admin.Project) error {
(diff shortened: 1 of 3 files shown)


pkg/manager/impl/validation/project_validator.go
```diff
@@ -25,7 +25,11 @@ func ValidateProjectRegisterRequest(request admin.ProjectRegisterRequest) error
 	if request.Project == nil {
 		return shared.GetMissingArgumentError(shared.Project)
 	}
-	return ValidateProject(*request.Project)
+	project := *request.Project
+	if err := ValidateEmptyStringField(project.Name, projectName); err != nil {
+		return err
+	}
+	return ValidateProject(project)
```
katrogan, line 32: do you want to move more of the validation (e.g. description, labels, etc) out of ValidateProject and into this method?

katrogan: oh perhaps labels is unnecessary, but description maybe

honnix: We are not validating emptiness of description in `VadlidateProject`. Do you mean we want to enforce non-empty for description as well?

katrogan: ah you're right again, sorry skimmed over that

katrogan (approved):

pkg/manager/impl/validation/project_validator.go
honnix: No worries. :)

katrogan (approved):

PR #151, opened 2021-02-02
Publish raw events

tnsetting:
# TL;DR
Send raw flyte events to a pubsub topic. This is the first step for flyte lineage events. These events will be consumed and enriched to external by another thread or process. If the `externalEvent` part is not configured, a noop publisher will be created.

Example of config:
```
notifications:
  type: gcp
  region: "my-region"
  gcp:
    projectId: "test"
  publisher:
    topicName: "topic"
  processor:
    queueName: "queue"
    accountId: "bar"
  externalEvent:
    eventPublisher:
      topicName: "event_topic"
      eventTypes: "all" // Or workflow, node, task. Or "*"
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/<number>

## Follow-up issue
_NA_
OR

### DIFF
--- a/pkg/async/notifications/factory.go
+++ b/pkg/async/notifications/factory.go
@@ -149,8 +149,8 @@ func NewNotificationsPublisher(config runtimeInterfaces.NotificationsConfig, sco
 	}
 }
 
-func NewEventsPublisher(config runtimeInterfaces.NotificationsConfig, scope promutils.Scope) interfaces.Publisher {
-	if config.ExternalEvent.EventPublisherConfig.TopicName == "" {
+func NewEventsPublisher(config runtimeInterfaces.ExternalEventsConfig, scope promutils.Scope) interfaces.Publisher {
+	if config.Enable == false {
 		return implementations.NewNoopPublish()
 	}
 	reconnectAttempts := config.ReconnectAttempts
@@ -158,13 +158,9 @@ func NewEventsPublisher(config runtimeInterfaces.NotificationsConfig, scope prom
 	switch config.Type {
 	case common.AWS:
 		snsConfig := gizmoAWS.SNSConfig{
-			Topic: config.ExternalEvent.EventPublisherConfig.TopicName,
+			Topic: config.EventsPublisherConfig.TopicName,
 		}
-		if config.AWSConfig.Region != "" {
-			snsConfig.Region = config.AWSConfig.Region
-		} else {
-			snsConfig.Region = config.Region
-		}
+		snsConfig.Region = config.AWSConfig.Region
 
 		var publisher pubsub.Publisher
 		var err error
@@ -177,10 +173,10 @@ func NewEventsPublisher(config runtimeInterfaces.NotificationsConfig, scope prom
 		if err != nil {
 			panic(err)
 		}
-		return implementations.NewEventsPublisher(publisher, scope, config.ExternalEvent.EventPublisherConfig.EventTypes)
+		return implementations.NewEventsPublisher(publisher, scope, config.EventsPublisherConfig.EventTypes)
 	case common.GCP:
 		pubsubConfig := gizmoGCP.Config{
-			Topic: config.ExternalEvent.EventPublisherConfig.TopicName,
+			Topic: config.EventsPublisherConfig.TopicName,
 		}
 		pubsubConfig.ProjectID = config.GCPConfig.ProjectID
 		var publisher pubsub.MultiPublisher
@@ -193,7 +189,7 @@ func NewEventsPublisher(config runtimeInterfaces.NotificationsConfig, scope prom
 		if err != nil {
 			panic(err)
 		}
-		return implementations.NewEventsPublisher(publisher, scope, config.ExternalEvent.EventPublisherConfig.EventTypes)
+		return implementations.NewEventsPublisher(publisher, scope, config.EventsPublisherConfig.EventTypes)
 	case common.Local:
 		fallthrough
 	default:
--- a/pkg/async/notifications/factory.go
+++ b/pkg/async/notifications/factory.go
@@ -148,3 +148,54 @@ func NewNotificationsPublisher(config runtimeInterfaces.NotificationsConfig, sco
 		return implementations.NewNoopPublish()
 	}
 }
+
+func NewEventsPublisher(config runtimeInterfaces.NotificationsConfig, scope promutils.Scope) interfaces.Publisher {
+	reconnectAttempts := config.ReconnectAttempts
+	reconnectDelay := time.Duration(config.ReconnectDelaySeconds) * time.Second
+	switch config.Type {
+	case common.AWS:
+		snsConfig := gizmoAWS.SNSConfig{
+			Topic: config.NotificationsPublisherConfig.TopicName,
+		}
+		if config.AWSConfig.Region != "" {
+			snsConfig.Region = config.AWSConfig.Region
+		} else {
+			snsConfig.Region = config.Region
+		}
+
+		var publisher pubsub.Publisher
+		var err error
+		err = async.Retry(reconnectAttempts, reconnectDelay, func() error {
+			publisher, err = gizmoAWS.NewPublisher(snsConfig)
+			return err
+		})
+
+		// Any persistent errors initiating Publisher with Amazon configurations results in a failed start up.
+		if err != nil {
+			panic(err)
+		}
+		return implementations.NewEventsPublisher(publisher, scope, config.EventPublisherConfig.EventTypes)
+	case common.GCP:
+		pubsubConfig := gizmoGCP.Config{
+			Topic: config.EventPublisherConfig.TopicName,
+		}
+		pubsubConfig.ProjectID = config.GCPConfig.ProjectID
+		var publisher pubsub.MultiPublisher
+		var err error
+		err = async.Retry(reconnectAttempts, reconnectDelay, func() error {
+			publisher, err = gizmoGCP.NewPublisher(context.TODO(), pubsubConfig)
+			return err
+		})
+
+		if err != nil {
+			panic(err)
+		}
+		return implementations.NewEventsPublisher(publisher, scope, config.EventPublisherConfig.EventTypes)
+	case common.Local:
+		fallthrough
+	default:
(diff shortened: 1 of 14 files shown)


tnsetting:
@kumare3 or @katrogan Any comments for the PR? Would it make more sense to have separate config section for the event? Thanks.

pkg/async/notifications/factory.go
```diff
@@ -148,3 +148,57 @@ func NewNotificationsPublisher(config runtimeInterfaces.NotificationsConfig, sco
 		return implementations.NewNoopPublish()
 	}
 }
+
+func NewEventsPublisher(config runtimeInterfaces.NotificationsConfig, scope promutils.Scope) interfaces.Publisher {
+	if config.ExternalEvent.EventPublisherConfig.TopicName == "" {
```
katrogan, line 153: can we explicitly just use the `Type` of the Notificationsconfig. and if it's not no-op, error when Topic Name is empty?

pkg/async/notifications/implementations/event_publisher.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// TODO: Add a counter that encompasses the publisher stats grouped by project and domain.
+type EventPublisher struct {
+	pub           pubsub.Publisher
+	systemMetrics eventPublisherSystemMetrics
+	events        []string
+}
+
+func getSupportedEvents() map[string]string {
```
katrogan, line 31: nit: instead of using a func you can actually initialize the map in golang as a var. see https://stackoverflow.com/questions/41078272/initializing-a-go-map-in-a-single-statement for an example

pkg/async/notifications/implementations/event_publisher.go
```diff
@@ -0,0 +1,98 @@
...
+		for _, e := range supportedEvents {
+			eventList = append(eventList, e)
+		}
+	} else {
+		events := strings.Split(eventTypes, ",")
+		for _, event := range events {
+			if e, found := supportedEvents[event]; found {
+				eventList = append(eventList, e)
+			}
+		}
```
katrogan, line 90: can we raise an error or log if a user passes in an unsupported event?

pkg/async/notifications/implementations/event_publisher.go
```diff
@@ -0,0 +1,98 @@
...
+		Scope:        scope,
+		PublishTotal: scope.MustNewCounter("event_publish_total", "overall count of publish messages"),
+		PublishError: scope.MustNewCounter("event_publish_errors", "count of publish errors"),
+	}
+}
+
+func NewEventsPublisher(pub pubsub.Publisher, scope promutils.Scope, eventTypes string) interfaces.Publisher {
+	supportedEvents := getSupportedEvents()
+
+	var eventList = make([]string, 0)
```
katrogan, line 79: do you want to use a[ String set](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String) here instead since you only ever check for membership/

pkg/async/notifications/implementations/event_publisher.go
```diff
@@ -0,0 +1,98 @@
...
+	supportedEvents["node"] = proto.MessageName(&nodeExecutionReq)
+	var workflowExecutionReq admin.WorkflowExecutionEventRequest
+	supportedEvents["workflow"] = proto.MessageName(&workflowExecutionReq)
+	return supportedEvents
+}
+
+// The key is the notification type as defined as an enum.
+func (p *EventPublisher) Publish(ctx context.Context, notificationType string, msg proto.Message) error {
+	p.systemMetrics.PublishTotal.Inc()
+	logger.Debugf(ctx, "Publishing the following message [%s]", msg.String())
```
katrogan, line 45: does String() return a human readable message? if not can we use `...[%+v]", msg)` instead?

pkg/async/notifications/implementations/event_publisher.go
```diff
@@ -0,0 +1,98 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func newEventPublisherSystemMetrics(scope promutils.Scope) eventPublisherSystemMetrics {
+	return eventPublisherSystemMetrics{
+		Scope:        scope,
+		PublishTotal: scope.MustNewCounter("event_publish_total", "overall count of publish messages"),
```
katrogan, line 71: i know it's all additive, but can we add a success counter too?

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -146,6 +146,18 @@ type NotificationsEmailerConfig struct {
 	Body string `json:"body"`
 }
 
+// This section handles configuration for the workflow notifications pipeline.
+type EventsPublisherConfig struct {
+	// The topic which events should be published, e.g. node, task, workflow
+	TopicName string `json:"topicName"`
+	// Event types: task, node, workflow executions
+	EventTypes string `json:"eventTypes"`
```
katrogan, line 154: why not an array type?

katrogan (commented):
looks good! just a few suggestions

katrogan:
> Would it make more sense to have separate config section for the event instead of inside the notification section? Thanks.

Yes let's please do that! I think we can worry about refactoring the existing email notifications implementation separately from the egress event work you're doing. Eventually notifications can be handled only by a separately configured subscriber so it make sense to separate out the events config now!

pkg/runtime/interfaces/application_configuration.go
tnsetting: Because I want to support the option to use `all` or `*` and using a single string makes it easier to filter on such cases.

pkg/async/notifications/implementations/event_publisher.go
tnsetting: Error loggings added.

katrogan:
hey @tnsetting is this ready for re-review?

tnsetting:
> hey @tnsetting is this ready for re-review?

Yes, it is ready.

pkg/runtime/interfaces/application_configuration.go
katrogan: but you can still do that with a single entry, no? I'd prefer to use conventional yaml list formatting vs our own string-split scheme

pkg/async/notifications/implementations/event_publisher.go
```diff
@@ -0,0 +1,98 @@
...
+	systemMetrics eventPublisherSystemMetrics
+	events        sets.String
+}
+
+var taskExecutionReq admin.TaskExecutionEventRequest
+var nodeExecutionReq admin.NodeExecutionEventRequest
+var workflowExecutionReq admin.WorkflowExecutionEventRequest
+
+var supportedEvents = map[string]string{
+	"task":     proto.MessageName(&taskExecutionReq),
```
katrogan, line 39: sorry should have brought this up in the last round, but can we use shared constants here as map keys? perhaps https://github.com/flyteorg/flyteadmin/blob/master/pkg/common/entity.go could work

pkg/async/notifications/implementations/event_publisher.go
```diff
@@ -0,0 +1,98 @@
...
+	"node":     proto.MessageName(&nodeExecutionReq),
+	"workflow": proto.MessageName(&workflowExecutionReq),
+}
+
+// The key is the notification type as defined as an enum.
+func (p *EventPublisher) Publish(ctx context.Context, notificationType string, msg proto.Message) error {
+	p.systemMetrics.PublishTotal.Inc()
+	logger.Debugf(ctx, "Publishing the following message [%+v]", msg)
```
katrogan, line 47: nit: do you want to move this log statement below the check to shouldPublishEvent?

pkg/async/notifications/implementations/event_publisher_test.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// This method should be invoked before every test around Publisher.
+func initializeEventPublisher() {
+	testEventPublisher.Published = nil
+	testEventPublisher.GivenError = nil
+	testEventPublisher.FoundError = nil
+}
+
+func TestNewEventsPublisher_EventTypes(t *testing.T) {
```
katrogan, line 90: Thank you for this very exhaustive test 👍

pkg/manager/impl/node_execution_manager.go
```diff
@@ -4,6 +4,9 @@ import (
 	"context"
 	"strconv"
 
+	"github.com/gogo/protobuf/proto"
```
katrogan, line 7: can we use `github.com/golang/protobuf/proto` instead? gogo behaves weirdly with oneofs

katrogan (commented):
sorry just a few more comments

pkg/manager/impl/node_execution_manager.go
tnsetting: Sure. Never paid attention for that as intellij did that for me.

pkg/async/notifications/implementations/event_publisher.go
```diff
@@ -0,0 +1,105 @@
...
+	systemMetrics eventPublisherSystemMetrics
+	events        sets.String
+}
+
+var taskExecutionReq admin.TaskExecutionEventRequest
+var nodeExecutionReq admin.NodeExecutionEventRequest
+var workflowExecutionReq admin.WorkflowExecutionEventRequest
+
+const (
+	Task          = "task"
```
katrogan, line 38: I'm still concerned this is a little brittle because here we define the strings statically but in the manager methods we use the proto message name as the event type. does that correspond to these raw strings like "task" or "node"?

katrogan (commented):
do you mind updating https://github.com/flyteorg/flyteadmin/blob/master/flyteadmin_config.yaml with an example of the config changes too?

pkg/async/notifications/implementations/event_publisher.go
katrogan: @tnsetting

katrogan (approved):
THANK YOU so much for this change

PR #153, opened 2021-03-02
Add project and domain to cluster resource templating

anandswaminathan:
Our current implementation only uses namespace as system templating value. The actual values should be project/domain as we iterate around that list.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

@katrogan @wild-endeavor

### DIFF
--- a/pkg/clusterresource/controller.go
+++ b/pkg/clusterresource/controller.go
@@ -42,8 +42,8 @@ import (
 )
 
 const namespaceVariable = "namespace"
-const projectVariable = "namespace"
-const domainVariable = "namespace"
+const projectVariable = "project"
+const domainVariable = "domain"
 const templateVariableFormat = "{{ %s }}"
 const replaceAllInstancesOfString = -1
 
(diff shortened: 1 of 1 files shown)


pkg/clusterresource/controller.go
```diff
@@ -40,6 +42,8 @@ import (
 )
 
 const namespaceVariable = "namespace"
+const projectVariable = "namespace"
```
katrogan, line 45:
```suggestion
const projectVariable = "project"
```

pkg/clusterresource/controller.go
```diff
@@ -40,6 +42,8 @@ import (
...
+const domainVariable = "namespace"
```
katrogan, line 46:
```suggestion
const domainVariable = "domain"
```

katrogan:
you'll need to update this to https://github.com/flyteorg/flyteadmin/blob/master/pkg/clusterresource/controller.go#L270 to do the actual substitution

katrogan (approved):

PR #159, opened 2021-03-17
Patch merge custom_info in task execution events

katrogan:
# TL;DR
Patch merge custom_info in task execution events

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/832

## Follow-up issue
_NA_


pkg/repositories/transformers/task_execution.go
```diff
@@ -152,6 +157,47 @@ func mergeLogs(existing, latest []*core.TaskLog) []*core.TaskLog {
...
+	}
+	if latest == nil {
+		return existing, nil
+	}
+
+	// To merge latest into existing we first create a patch object that consists of applying changes from latest to
+	// an empty struct. Then we apply this patch to existing so that the values changed in latest take precedence but
+	// barring conflicts/overwrites the values in existing stay the same.
+	var empty _struct.Struct
+	jsonEmpty, err := json.Marshal(&empty)
```
EngHabu, line 172: Should we use jsonpb to marshal proto messages?

katrogan: @EngHabu done

pkg/repositories/transformers/task_execution.go
```diff
@@ -152,6 +157,47 @@ func mergeLogs(existing, latest []*core.TaskLog) []*core.TaskLog {
...
+		return latest, nil
+	}
+	if latest == nil {
+		return existing, nil
+	}
+
+	// To merge latest into existing we first create a patch object that consists of applying changes from latest to
+	// an empty struct. Then we apply this patch to existing so that the values changed in latest take precedence but
+	// barring conflicts/overwrites the values in existing stay the same.
+	var empty _struct.Struct
```
kumare3, line 175: do you have to create this everytime? Can you not just create it once and re-use as a constant? Reading the docs makes me think its possible. This would reduce garbage

pkg/repositories/transformers/task_execution.go
katrogan: @EngHabu done

pkg/repositories/transformers/task_execution.go
kumare3: Also have you looked at - https://pkg.go.dev/k8s.io/apimachinery/pkg/util/strategicpatch#CreateTwoWayMergePatch

kumare3:
I think this mergepatching is awesome, it opens up great possibilities for tasktemplate/custom. I am thinking if tasktemplate custom can always be openAPISpec?

pkg/repositories/transformers/task_execution.go
katrogan: @kumare3 done

katrogan: @kumare3 re the link, this doesn't quite do what we need

> It will return a patch that yields the modified document when applied to the original document

pkg/repositories/transformers/task_execution.go
```diff
@@ -152,6 +164,42 @@ func mergeLogs(existing, latest []*core.TaskLog) []*core.TaskLog {
...
+	if err != nil {
+		return nil, err
+	}
+	custom, err := jsonpatch.MergePatch([]byte(jsonExisting), patch)
+	if err != nil {
+		return nil, err
+	}
+	var response _struct.Struct
+
+	err = json.Unmarshal(custom, &response)
```
EngHabu, line 196: Should we use jsonpb for this too :-D ? it's an honest question not a suggestion in form of a question... I'm not sure if the answer is yes... for both marshal and unmashal... If I recall correctly the _only_ difference is handling of oneofs? if there are no oneofs in pb.Struct then maybe it doesn't matter?

katrogan: my thinking for using json rather than jsonpb is that if we're not marshaling a proto into the custom struct to being with it makes sense to just use json but i suppose the answer is, it depends. from looking through plugins code i see existing customs as mostly just maps

EngHabu: So, at least in theory, even if the original message was "complex" enough that it needed jsonpb to be marshaled, it doesn't affect the decision whether we should use jsonpb or not on Struct because it has already been transformed to Struct at this point, right?

So the question is whether there are things you can set on a Struct that would make a difference... jsonpb (and protojson) both say you should use them to marshal/unmarshal proto types because they "produce different outputs" compared to the regular encoding/json... they don't seem to go into details as to what is different about the output..

So it's perhaps safer to just stick to the proto-recommended way of using json for both marshaling and unmarshaling?

btw this is the new jsonpb (jsonpb is deprecated) but I don't know if it works on our generated types or do we need to update protoc (which we will have to at some point).
https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson

katrogan: sg updated to use protojson! thanks for pointing that out btw, love the consistent interface

katrogan:
friendly ping

kumare3 (approved):

PR #172, opened 2021-03-30
Add exists check for workflow & node executions

katrogan:

# TL;DR
Add exists check for workflow & node executions to avoid fetching and passing around fully fledged executions in memory.

Additionally, only add the exists check for these parent executions when node and task executions, respectively, are not found and we create an entry in the db for them for the first time.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/880

## Follow-up issue
_NA_


kumare3 (approved):
This is great!

kumare3:
I tested this change, seems good to go, performance improvement is not great, but this is the right change nevertheless

PR #174, opened 2021-04-02
Write workflow and node execution events asynchronously

katrogan:
# TL;DR
Write workflow and node execution events asynchronously

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Should hopefully improve Create{Workflow|Node}ExecutionEvent performance

## Tracking Issue
https://github.com/flyteorg/flyte/issues/891

## Follow-up issue
_NA_


pkg/async/events/implementations/node_execution_event_writer.go
```diff
@@ -0,0 +1,41 @@
...
+		if err != nil {
+			logger.Warnf(context.TODO(), "Failed to write event [%+v] to database with err [%+v]", event, err)
+		}
+	}
+}
+
+func NewNodeExecutionEventWriter(db repositories.RepositoryInterface) interfaces.NodeExecutionEventWriter {
+	return &nodeExecutionEventWriter{
+		db:     db,
+		events: make(chan admin.NodeExecutionEventRequest),
```
kumare3, line 39: you should make a buffered channel?

kumare3: You should and the length of the channel should be pretty large, else you will block sender. Right now this is a sync send

pkg/async/events/implementations/workflow_execution_event_writer.go
```diff
@@ -0,0 +1,41 @@
...
+		if err != nil {
+			logger.Warnf(context.TODO(), "Failed to write event [%+v] to database with err [%+v]", event, err)
+		}
+	}
+}
+
+func NewWorkflowExecutionEventWriter(db repositories.RepositoryInterface) interfaces.WorkflowExecutionEventWriter {
+	return &workflowExecutionEventWriter{
+		db:     db,
+		events: make(chan admin.WorkflowExecutionEventRequest),
```
kumare3, line 39: ditto

pkg/async/events/implementations/node_execution_event_writer.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+
+	"github.com/flyteorg/flyteadmin/pkg/async/events/interfaces"
+	"github.com/flyteorg/flyteadmin/pkg/repositories"
+	"github.com/flyteorg/flyteadmin/pkg/repositories/transformers"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
+	"github.com/flyteorg/flytestdlib/logger"
+)
+
+type nodeExecutionEventWriter struct {
```
kumare3, line 13: Lets add a comment, as to why we are doing this

pkg/async/events/implementations/workflow_execution_event_writer.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+
+	"github.com/flyteorg/flyteadmin/pkg/async/events/interfaces"
+	"github.com/flyteorg/flyteadmin/pkg/repositories"
+	"github.com/flyteorg/flyteadmin/pkg/repositories/transformers"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
+	"github.com/flyteorg/flytestdlib/logger"
+)
+
+type workflowExecutionEventWriter struct {
```
kumare3, line 13: same here

pkg/async/events/implementations/node_execution_event_writer.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (w *nodeExecutionEventWriter) Run() {
+	for event := range w.events {
+		eventModel, err := transformers.CreateNodeExecutionEventModel(event)
+		if err != nil {
+			logger.Warnf(context.TODO(), "Failed to transform event [%+v] to database model with err [%+v]", event, err)
+			continue
+		}
+		err = w.db.NodeExecutionEventRepo().Create(context.TODO(), *eventModel)
```
kumare3, line 29: can we add a comment here explaining that we will drop the event on the floor and why it is ok to be lossy!

kumare3:
I do not think this is achieving the async comm, Also, I think  we should add enough documentation to explain the actual semantics of the API

pkg/async/events/implementations/node_execution_event_writer.go
katrogan: yes good catch, ty! not sure what a good buffer size would be, any suggestions @kumare3 ?

katrogan: done & added a test for this case

pkg/async/events/implementations/node_execution_event_writer.go
katrogan: done

pkg/async/events/implementations/workflow_execution_event_writer.go
katrogan: done

pkg/async/events/implementations/node_execution_event_writer.go
katrogan: done

pkg/async/events/implementations/workflow_execution_event_writer.go
katrogan: done

pkg/async/events/implementations/node_execution_event_writer.go
kumare3: Why do we not have to wait - because we create the coalesced node-execution event

pkg/async/events/implementations/node_execution_event_writer.go
```diff
@@ -0,0 +1,45 @@
...
+			// to replay and understand the event execution timeline.
+			logger.Warnf(context.TODO(), "Failed to write event [%+v] to database with err [%+v]", event, err)
+		}
+	}
+}
+
+func NewNodeExecutionEventWriter(db repositories.RepositoryInterface) interfaces.NodeExecutionEventWriter {
+	return &nodeExecutionEventWriter{
+		db:     db,
+		events: make(chan admin.NodeExecutionEventRequest, bufferSize),
```
kumare3, line 43: Let us take the buffer size as an input parameter?

kumare3:
Do we want to run a perf test before we commit this?

pkg/async/events/implementations/node_execution_event_writer.go
katrogan: done

katrogan:
@kumare3 no noticeable change, still coming at 15 mins although this time I see inserts into node execution events taking way less wait time comparatively

katrogan:
image
@kumare before and after comparison for create execution event endpoint

pkg/runtime/application_config_provider.go
```diff
@@ -56,7 +58,11 @@ func (p *ApplicationConfigurationProvider) GetDbConfig() interfaces.DbConfig {
 }
 
 func (p *ApplicationConfigurationProvider) GetTopLevelConfig() *interfaces.ApplicationConfig {
-	return flyteAdminConfig.GetConfig().(*interfaces.ApplicationConfig)
+	applicationConfig := flyteAdminConfig.GetConfig().(*interfaces.ApplicationConfig)
+	if applicationConfig.AsyncEventsBufferSize == 0 {
```
kumare3, line 62: no need to do this, just set the default in the config like so
https://github.com/flyteorg/flytepropeller/blob/master/pkg/controller/config/config.go#L47

katrogan: @kumare3 done, ty!

pkg/runtime/application_config_provider.go
```diff
@@ -19,7 +19,12 @@ const domains = "domains"
 const externalEvents = "externalEvents"
 
 var databaseConfig = config.MustRegisterSection(database, &interfaces.DbConfigSection{})
-var flyteAdminConfig = config.MustRegisterSection(flyteAdmin, &interfaces.ApplicationConfig{})
+
+var defaultFlyteAdminConfig = &interfaces.ApplicationConfig{
```
kumare3, line 23: this might be a time to create default configs for all the configs? Or if you want do it as a follow up PR?

kumare3:
looks good to me, one comment

pkg/runtime/application_config_provider.go
katrogan: considering how many configs we have... how about a follow-up pr :)

katrogan:
@kumare3 can you approve?

pkg/runtime/application_config_provider.go
katrogan: https://github.com/flyteorg/flyte/issues/904

kumare3 (approved):

PR #178, opened 2021-04-09
Always return execution data inline when no remote data provider set

katrogan:

# TL;DR
When no remote data provider is set or the local/noop option is specified, always return execution inputs and outputs inline for GetData requests

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/900

## Follow-up issue
_NA_


pkg/runtime/application_config_provider.go
```diff
@@ -21,7 +23,13 @@ const externalEvents = "externalEvents"
 var databaseConfig = config.MustRegisterSection(database, &interfaces.DbConfigSection{})
 var flyteAdminConfig = config.MustRegisterSection(flyteAdmin, &interfaces.ApplicationConfig{})
 var schedulerConfig = config.MustRegisterSection(scheduler, &interfaces.SchedulerConfig{})
```
EngHabu, line 25: Do you know why are we registering many sections in one package? it works... but not needed I guess....

pkg/runtime/application_config_provider.go
```diff
@@ -21,7 +23,13 @@ const externalEvents = "externalEvents"
...
-var remoteDataConfig = config.MustRegisterSection(remoteData, &interfaces.RemoteDataConfig{})
+
+var defaultRemoteDataConfig = &interfaces.RemoteDataConfig{
+	Scheme:         common.Local,
+	MaxSizeInBytes: 1048576, // 1 Mib
+}
+var remoteDataConfig = config.MustRegisterSection(remoteData, defaultRemoteDataConfig)
```
EngHabu, line 31: nit: Shouldn't we also generate pflags provider for these?

EngHabu (dismissed):

pkg/manager/impl/execution_manager.go
```diff
@@ -1111,15 +1111,16 @@ func (m *ExecutionManager) GetExecutionData(
 		Inputs:  &inputsURLBlob,
 	}
 	maxDataSize := m.config.ApplicationConfiguration().GetRemoteDataConfig().MaxSizeInBytes
-	if maxDataSize == 0 || inputsURLBlob.Bytes < maxDataSize {
+	remoteDataScheme := m.config.ApplicationConfiguration().GetRemoteDataConfig().Scheme
+	if remoteDataScheme == common.Local || inputsURLBlob.Bytes < maxDataSize {
```
kumare3, line 1115: what if we change it to always return inline and sometimes return the signed ulr?

pkg/runtime/application_config_provider.go
katrogan: oh just as a logical/organizational breakdown

pkg/manager/impl/execution_manager.go
katrogan: then we should just get rid of the signed url altogether, no?

kumare3: Thats what I want to do, but I dont know if that will work for all cases at Lyft? If we have a metric lets collect that?

katrogan: we have a metric for observing the bytes size - can follow up to find out what it trends towards

EngHabu (approved):

PR #182, opened 2021-04-13
Store dynamic workflows

katrogan:

# TL;DR
Store dynamic workflows

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/831

## Follow-up issue
_NA_


pkg/repositories/transformers/node_execution.go
```diff
@@ -173,12 +173,19 @@ func UpdateNodeExecutionModel(
...
 		st := request.Event.GetTaskNodeMetadata().GetCacheStatus().String()
-		nodeExecutionClosure.TargetMetadata = &admin.NodeExecutionClosure_TaskNodeMetadata{
+		targetMetadata := &admin.NodeExecutionClosure_TaskNodeMetadata{
 			TaskNodeMetadata: &admin.TaskNodeMetadata{
 				CacheStatus: request.Event.GetTaskNodeMetadata().GetCacheStatus(),
 				CatalogKey:  request.Event.GetTaskNodeMetadata().GetCatalogKey(),
 			},
 		}
+		if request.Event.GetTaskNodeMetadata().DynamicWorkflow != nil {
+			targetMetadata.TaskNodeMetadata.DynamicWorkflow = &admin.DynamicWorkflowNodeMetadata{
```
kumare3, line 183: are we storing it in the DB? Where do we store the current compiled workflow - i thought s3?

katrogan: oh yes good point! will revise to make this relational

katrogan: updated. also this follows the workflow model where a get will go and fetch the closure but the call to list node executions will not

katrogan:
PTAL @kumare3 @EngHabu

pkg/repositories/transformers/node_execution.go
kumare3: ok, but how does the lister know that the dynamic workflow has not been fetched?

kumare3: Also maybe I dont understand this, but where is the offloading code?

pkg/manager/impl/util/shared.go
```diff
@@ -42,6 +45,81 @@ func GetTask(ctx context.Context, repo repositories.RepositoryInterface, identif
...
+		identifier.Project,
+		identifier.Domain,
+		identifier.Name,
+		identifier.Version,
+	}
+	nestedKeys := append(storagePrefix, nestedSubKeys...)
+	return storageClient.ConstructReference(ctx, storageClient.GetBaseContainerFQN(ctx), nestedKeys...)
+}
+
+func WriteCompiledWorkflow(ctx context.Context, repo repositories.RepositoryInterface, storagePrefix []string,
```
kumare3, line 61: can we get some documentation on this public method

pkg/repositories/transformers/node_execution.go
katrogan: @kumare3 the lister will b/c a dynamic workflow id will exist but the corresponding closure will not. offloading is in this refactored shared method: https://github.com/flyteorg/flyteadmin/pull/182/files#diff-8826e8af1b89cbbc5038dae18de4989e16125b6148e236cd535954dc6cfb69c9R94

pkg/manager/impl/node_execution_manager.go
```diff
@@ -258,6 +272,21 @@ func (m *NodeExecutionManager) GetNodeExecution(
...
 		return nil, err
 	}
+	if nodeExecution.Closure.GetTaskNodeMetadata() != nil &&
+		nodeExecution.Closure.GetTaskNodeMetadata().DynamicWorkflow != nil &&
+		nodeExecution.Closure.GetTaskNodeMetadata().DynamicWorkflow.Id != nil {
+		dynamicWf, err := util.GetWorkflow(ctx, m.db, m.storageClient, *nodeExecution.Closure.GetTaskNodeMetadata().DynamicWorkflow.Id)
+		if err != nil {
+			logger.Warnf(ctx, "failed to fetch dynamic workflow definition for node [%+v] with err: %v", request.Id, err)
+		}
+		if dynamicWf.Closure != nil {
```
EngHabu, line 282: So GetNodeExecution is only used by console (and clis) to visualize an execution, right? it's never called by propeller in any scenario... I'm just worried about the size of these messages being passed around if that's the case... and then I started thinking about these: https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/FieldMask

pkg/manager/impl/node_execution_manager.go
```diff
@@ -225,6 +226,19 @@ func (m *NodeExecutionManager) CreateNodeEvent(ctx context.Context, request admi
 	}
 	m.dbEventWriter.Write(request)
 
+	// If the request references a dynamically compiled subworkflow in the node metadata, serialize that to the db.
+	if request.Event.GetTaskNodeMetadata() != nil && request.Event.GetTaskNodeMetadata().DynamicWorkflow != nil {
+		_, err := util.WriteCompiledWorkflow(ctx, m.db, m.storagePrefix, m.storageClient,
```
EngHabu, line 231: The prefix here is not specific to this execution, right? I think the path we store this workflow in should be sandboxed to the execution (e.g. `s3://myprefix/admin/project/domain/name/node1/sub-node/sub-node/<trial-num>/workflow_id`) because the same id can be returned multiple times (it's dynamically generated after all and the shape of the graph can depend on the inputs to the generator task)...
do you agree?

you don't have to have this path... just a suggestion... you can also hash the string and just keep it as a prefix of the workflow id (thought maybe maintaining the exec hierarchy is useful for debugging? what do you think?

pkg/manager/impl/node_execution_manager.go
katrogan: oh this is awesome! we should really integrate this throughout. would have been useful to have for the preliminary GetExecution performance hunt too

pkg/manager/impl/node_execution_manager.go
katrogan: oh good point - yes a human readable execution id in the closure path sounds like a good and readable guarantee of uniqueness! ty

katrogan: > The prefix here is not specific to this execution
No, but we do append the workflow keys to the prefix path to get the full closure loc

katrogan:
PTAL

pkg/manager/impl/util/shared.go
katrogan: done.

pkg/manager/impl/node_execution_manager.go
katrogan: and to answer your question, I don't see GetNodeExecution called anywhere in propeller

pkg/manager/impl/node_execution_manager.go
EngHabu: Do we register it? or just store it in storage?

EngHabu: I think we might actually change that rand to be the hash of the workflow closure for consistency...

pkg/manager/impl/node_execution_manager.go
EngHabu: I meant maybe propeller calls X that in turn calls GetNodeExecution? or something...

katrogan: well i checked plugins too, not sure what else there is to verify?

kumare3: we should never be calling GetNodeExecution in propeller. Question about the fieldmask, you cannot use that in a regular GetAPI IMO, as this changes the semantics of an API. The only way to do that is using a mask as an input, because then, this is the expectation by the user.

This is where GraphQL performs much better

pkg/manager/impl/node_execution_manager.go
kumare3: yes, why even worry about de-duping in S3? Does it really matter? We can simply write new keys right? And cache the key. The semantics are that, every write to workflow closure has a full copy of the closure, so does not matter, which one we consider as a copy?
Does the determinism help with anything?

pkg/manager/impl/validation/node_execution_validator.go
```diff
@@ -20,6 +21,35 @@ func ValidateNodeExecutionIdentifier(identifier *core.NodeExecutionIdentifier) e
 	return ValidateWorkflowExecutionIdentifier(identifier.ExecutionId)
 }
 
+func ValidateNodeExecutionEventRequest(request *admin.NodeExecutionEventRequest) error {
```
kumare3, line 24: i know the file does not have it, but some documentation please?

kumare3 (approved):

pkg/manager/impl/node_execution_manager.go
```diff
@@ -187,15 +190,52 @@ func (m *NodeExecutionManager) updateNodeExecutionWithEvent(
...
 
+func (m *NodeExecutionManager) uploadDynamicWorkflowClosure(
+	ctx context.Context, nodeID *core.NodeExecutionIdentifier, workflowID *core.Identifier,
+	compiledWorkflowClosure *core.CompiledWorkflowClosure) (storage.DataReference, error) {
+	nestedSubKeys := []string{
+		nodeID.ExecutionId.Project,
+		nodeID.ExecutionId.Domain,
+		nodeID.ExecutionId.Name,
+		nodeID.NodeId,
+		fmt.Sprintf("%s_%s_%s_%s", workflowID.Project, workflowID.Domain, workflowID.Name, workflowID.Version),
```
EngHabu, line 201: nit: Move to a helper function?

pkg/manager/impl/validation/node_execution_validator.go
katrogan: done

pkg/manager/impl/node_execution_manager.go
```diff
@@ -142,7 +144,8 @@ func (m *NodeExecutionManager) createNodeExecutionWithEvent(
 }
 
 func (m *NodeExecutionManager) updateNodeExecutionWithEvent(
-	ctx context.Context, request *admin.NodeExecutionEventRequest, nodeExecutionModel *models.NodeExecution) (updateNodeExecutionStatus, error) {
+	ctx context.Context, request *admin.NodeExecutionEventRequest, nodeExecutionModel *models.NodeExecution,
+	dynamicWorkflowRemoteClosure string) (updateNodeExecutionStatus, error) {
```
EngHabu, line 148: nit:
```suggestion
	dynamicWorkflowRemoteClosureReference storage.Reference) (updateNodeExecutionStatus, error) {
```

pkg/manager/impl/node_execution_manager.go
```diff
@@ -427,14 +467,27 @@ func (m *NodeExecutionManager) GetNodeExecutionData(
 		response.FullOutputs = &fullOutputs
 	}
 
+	if len(nodeExecutionModel.DynamicWorkflowRemoteClosure) > 0 {
```
EngHabu, line 470: nit:
```suggestion
	if len(nodeExecutionModel.DynamicWorkflowRemoteClosureReference) > 0 {
```

EngHabu (commented):
a couple of nits

pkg/manager/impl/node_execution_manager.go
katrogan: done

pkg/manager/impl/node_execution_manager.go
katrogan: done

pkg/manager/impl/node_execution_manager.go
katrogan: done

katrogan:
PTAL @EngHabu

EngHabu (approved):

PR #196, opened 2021-05-26
wip : Fix workflow list request to return closure data aswell

pmahindrakar-oss:

# TL;DR
Return the closure in the workflow list response aswell.
This might have been omitted for a purpose may be due to size of the closure data.
If not this then can we have the client get the latest version using one call and then call get workflow with it.

But i do see that we return the entire objects in case of launch plan list request so fixing this in the admin would be ideal.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

Unit tests  pending

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1053

## Follow-up issue
_NA_
OR

### DIFF
--- a/pkg/manager/impl/util/shared.go
+++ b/pkg/manager/impl/util/shared.go
@@ -56,6 +56,17 @@ func GetWorkflowModel(
 	return workflowModel, nil
 }
 
+// ListWorkflowModel lists the workflow model using the repo and passing in requested listing criteria.
+func ListWorkflowModel(
+	ctx context.Context, repo repositories.RepositoryInterface,
+	listResourceInput repoInterfaces.ListResourceInput) ([]models.Workflow, error) {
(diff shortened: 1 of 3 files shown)


katrogan:
@pmahindrakar-oss is there a use-case for this or are you just fixing for consistency's sake? the closure can indeed be very big so it would be ideal if flytectl can call a get in the background transparently

pmahindrakar-oss:
@katrogan its more to make it consistent with launch plan behavior , but i agree if the payload is going to be huge then flytectl can call the get transparently . Will send PR to make launchplan consistent with this

katrogan:
@pmahindrakar-oss what do you mean by making launch plan consistent with this?

PR #199, opened 2021-05-27
Allow external emailers

wild-endeavor:
Signed-off-by: wild-endeavor <wild-endeavor@users.noreply.github.com>

# TL;DR
Add ability to use sendgrid

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
* Accidentally ran `go mod tidy` and `make generate` so a bunch of other changes got brought in.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/<number>


pkg/async/notifications/factory.go
```diff
@@ -41,6 +43,17 @@ type EmailerConfig struct {
 }
 
 func GetEmailer(config runtimeInterfaces.NotificationsConfig, scope promutils.Scope) interfaces.Emailer {
+	// If an external email service is specified use that first.
+
+	if config.NotificationsEmailerConfig.EmailerConfig.ServiceName != ""{
```
katrogan, line 48: might be worth refactoring to to make the EmailerConfig accept 'aws'  and 'local' as a service name so we don't have to do two types of checks

pkg/async/notifications/implementations/sendgrid_emailer.go
```diff
@@ -0,0 +1,45 @@
...
+type SendgridEmailer struct {
+	client        *sendgrid.Client
+	systemMetrics emailMetrics
+}
+
+func (s SendgridEmailer) SendEmail(ctx context.Context, email admin.EmailMessage) error {
+	from := mail.NewEmail("Flyte Notifications", "workflow-notifications@uniondemo.run")
+	subject := email.SubjectLine
+	to := mail.NewEmail("Sou", "souffle@example.com") // Change to your recipient
+	message := mail.NewSingleEmail(from, subject, to, "", email.Body)
```
katrogan, line 25: what's the fourth arg?

pkg/async/notifications/implementations/sendgrid_emailer.go
```diff
@@ -0,0 +1,45 @@
...
+
+type SendgridEmailer struct {
+	client        *sendgrid.Client
+	systemMetrics emailMetrics
+}
+
+func (s SendgridEmailer) SendEmail(ctx context.Context, email admin.EmailMessage) error {
+	from := mail.NewEmail("Flyte Notifications", "workflow-notifications@uniondemo.run")
+	subject := email.SubjectLine
+	to := mail.NewEmail("Sou", "souffle@example.com") // Change to your recipient
```
katrogan, line 24: update?

pkg/async/notifications/implementations/sendgrid_emailer.go
```diff
@@ -0,0 +1,45 @@
...
+	message := mail.NewSingleEmail(from, subject, to, "", email.Body)
+	s.systemMetrics.SendTotal.Inc()
+	response, err := s.client.Send(message)
+	if err != nil {
+		logger.Errorf(ctx, "Sendgrid error sending %s", err)
+		s.systemMetrics.SendError.Inc()
+		return err
+	} else {
+		s.systemMetrics.SendSuccess.Inc()
```
katrogan, line 33: nit, since you return in the if block above you don't need this in an else block

pkg/async/notifications/factory.go
wild-endeavor: I can add that, but I don't want to remove the existing check - if I do, it'll be a breaking config change.  or you think it's worth it?

katrogan: sure sure let's avoid a breaking change in this PR. maybe we can file an issue to track cleaning up the emailer configs

wild-endeavor: https://github.com/flyteorg/flyte/issues/1063

pkg/async/notifications/factory.go
```diff
@@ -41,6 +42,16 @@ type EmailerConfig struct {
 }
 
 func GetEmailer(config runtimeInterfaces.NotificationsConfig, scope promutils.Scope) interfaces.Emailer {
+	// If an external email service is specified use that first.
```
EngHabu, line 45: nit
```suggestion
	// If an external email service is specified use that instead.
```

pkg/async/notifications/implementations/sendgrid_emailer.go
```diff
@@ -0,0 +1,67 @@
...
+
+type SendgridEmailer struct {
+	client        *sendgrid.Client
+	systemMetrics emailMetrics
+}
+
+func getEmailAddresses(addresses []string) []*mail.Email {
+	sendgridAddresses := make([]*mail.Email, len(addresses))
+	for idx, email := range addresses {
+		sendgridAddresses[idx] = mail.NewEmail("", email)
```
katrogan, line 25: comment or named variable for the "" please?

pkg/async/notifications/implementations/sendgrid_emailer.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	return sendgridAddresses
+}
+
+func getSendgridEmail(adminEmail admin.EmailMessage) *mail.SGMailV3 {
+	m := mail.NewV3Mail()
+
+	from := mail.NewEmail("Flyte Notifications", adminEmail.SenderEmail)
```
katrogan, line 33: what is "Flyte Notifications" used for? should this be configurable?

pkg/async/notifications/implementations/sendgrid_emailer.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	s.systemMetrics.SendSuccess.Inc()
+	logger.Debugf(ctx, "Sendgrid sent email %s", response.Body)
+
+	return nil
+}
+
+func NewSendGridEmailer(config runtimeInterfaces.NotificationsConfig, scope promutils.Scope) interfaces.Emailer {
+	return &SendgridEmailer{
+		client:        sendgrid.NewSendClient(os.Getenv(config.NotificationsEmailerConfig.EmailerConfig.APIKeyEnvVar)),
```
katrogan, line 64: would it be worthwhile supporting different ways of passing the api key beyond just as an env var? for example, specifying a file path to the secret too? maybe something to worry about if/when someone requests so

pkg/async/notifications/implementations/sendgrid_emailer.go
```diff
@@ -0,0 +1,67 @@
...
+		systemMetrics: newEmailMetrics(scope.NewSubScope("sendgrid")),
```
katrogan, line 65: can we move newEmailMetrics out of the SES impl and into somewhere common?

pkg/async/notifications/implementations/sendgrid_emailer.go
wild-endeavor: you know how when you send an email, gmail shows you something like "Alice <alice@umbrella.com>".  I think it's just the first part of it, the "name".  I don't think it's worth it to make it configurable, since at least for sendgrid, it doesn't actually matter (should probably add a comment though). The sender is determined by the API key.

pkg/async/notifications/implementations/sendgrid_emailer.go
wild-endeavor: sure.

pkg/async/notifications/implementations/sendgrid_emailer_test.go
```diff
@@ -41,3 +45,54 @@ func TestCreateEmailer(t *testing.T) {
 	emailer := NewSendGridEmailer(cfg, promutils.NewTestScope())
 	assert.NotNil(t, emailer)
 }
+
+func TestReadAPIFromEnv(t *testing.T) {
+	envVar := "sendgrid_api_test_key"
```
katrogan, line 50: not sure if it's worth storing what env var (if any) was here originally and resetting after the test runs

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -139,8 +139,9 @@ type NotificationsProcessorConfig struct {
 }
 
 type EmailServerConfig struct {
-	ServiceName  string `json:"serviceName"`
-	APIKeyEnvVar string `json:"apiKeyEnvVar"`
+	ServiceName    string `json:"serviceName"`
+	APIKeyEnvVar   string `json:"apiKeyEnvVar"`
```
katrogan, line 143: Mind adding a comment that one of APIKeyEnvVar or APIKeyFilePath should be specified?

katrogan (dismissed):

katrogan (approved):

PR #202, opened 2021-05-31
Connection pooling for flyteadmin db

evalsocket:
# TL;DR
1. Added new db option in config like ` maxOpenConnection,maxIdleConnection,connectionMaxLifetime `
2. Set db option for connection pooling  

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Added new config and set value in Database connection 
```
database:
  maxOpenConnection: 5
  maxIdleConnection: 25
  connectionMaxLifetime: 1000
```

## Tracking Issue
https://github.com/flyteorg/flyte/issues/881

## Follow-up issue
_NA_
OR

### DIFF
--- a/pkg/repositories/config/postgres.go
+++ b/pkg/repositories/config/postgres.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/flyteorg/flytestdlib/promutils"
 
@@ -9,7 +10,12 @@ import (
 	_ "github.com/jinzhu/gorm/dialects/postgres" // Required to import database driver.
 )
 
-const Postgres = "postgres"
+const (
+	Postgres                     = "postgres"
+	DefaultMaxIdleConnection     = 25
+	DefaultMaxOpenConnection     = 1
+	DefaultConnectionMaxLifetime = time.Duration(1000)
+)
 
 // Generic interface for providing a config necessary to open a database connection.
 type DbConnectionConfigProvider interface {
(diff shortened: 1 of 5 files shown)


katrogan (commented):
It's a bit confusing but can you update https://github.com/flyteorg/flyteadmin/blob/master/pkg/runtime/interfaces/application_configuration.go#L6? The db config predates the runtime structure and contains several changes

Also were you able to test this locally?

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -20,6 +22,12 @@ type DbConfigSection struct {
 	ExtraOptions string `json:"options"`
 	// Whether or not to start the database connection with debug mode enabled.
 	Debug bool `json:"debug"`
+	// Max open connection in postgres
+	MaxOpenConnection     int           `json:"maxOpenConnection"`
```
katrogan, line 26: you need to copy these over to the db configuration struct, right? https://github.com/flyteorg/flyteadmin/blob/a4abd4cbb0552a905b95b5d1e58e95baef9f7a7e/pkg/rpc/adminservice/base.go#L83 here and elsewhere

pkg/repositories/config/postgres.go
```diff
@@ -2,21 +2,33 @@ package config
...
 	"github.com/flyteorg/flytestdlib/promutils"
 
 	"github.com/jinzhu/gorm"
 	_ "github.com/jinzhu/gorm/dialects/postgres" // Required to import database driver.
 )
 
-const Postgres = "postgres"
+const (
+	Postgres                     = "postgres"
+	DefaultMaxIdleConnection     = 25
```
katrogan, line 15: instead of setting these defaults here, can we set them in the runtime config when calling MustRegisterSection like so: https://github.com/flyteorg/flyteadmin/blob/master/pkg/runtime/application_config_provider.go#L25,L30

pkg/runtime/interfaces/application_configuration.go
evalsocket: @kanterov why do we need to pass these config in extraOption. I used gorm approach https://gorm.io/docs/generic_interface.html

pkg/repositories/config/postgres.go
evalsocket: ok

PR #218, opened 2021-07-07
Add pod spec validation

katrogan:

# TL;DR
Now that pod specs are first class task template targets we should add some preliminary validation upon registration since we can reason about the expected object structure.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/998

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/validation/task_validator.go
+++ b/pkg/manager/impl/validation/task_validator.go
@@ -4,6 +4,10 @@ package validation
 import (
 	"context"
 
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/apimachinery/pkg/util/validation"
+
 	"github.com/flyteorg/flyteadmin/pkg/repositories"
 
 	"github.com/flyteorg/flyteadmin/pkg/common"
@@ -20,11 +24,6 @@ import (
 
 var whitelistedTaskErr = errors.NewFlyteAdminErrorf(codes.InvalidArgument, "task type must be whitelisted before use")
 
-// Sidecar tasks do not necessarily define a primary container for execution and are excluded from container validation.
-var containerlessTaskTypes = map[string]bool{
-	"sidecar": true,
-}
-
 // This is called for a task with a non-nil container.
 func validateContainer(task core.TaskTemplate, taskConfig runtime.TaskResourceConfiguration) error {
 	if err := ValidateEmptyStringField(task.GetContainer().Image, shared.Image); err != nil {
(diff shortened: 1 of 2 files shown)


kumare3:
ohh this is awesome! - just that we will have to potentially have different ones for v1.Pod, v2.Pod (if that ever happens).

pkg/manager/impl/validation/task_validator.go
```diff
@@ -43,6 +42,26 @@ func validateContainer(task core.TaskTemplate, taskConfig runtime.TaskResourceCo
...
+		return errors.NewFlyteAdminErrorf(codes.InvalidArgument, "K8sPod task type targets must specify a non-empty pod spec")
+	}
+	var podSpec v1.PodSpec
+	jsonObj, err := json.Marshal(task.GetK8SPod().PodSpec)
+	if err != nil {
+		return errors.NewFlyteAdminErrorf(codes.InvalidArgument, "Failed to json marshal K8sPod task type target with err: %v", err)
+	}
+	if err = json.Unmarshal(jsonObj, &podSpec); err != nil {
+		return errors.NewFlyteAdminErrorf(codes.InvalidArgument, "Failed to json unmarshal K8sPod task type target with err: %v", err)
+	}
```
EngHabu, line 56: Should we move this pattern somewhere? json marshal json unmarshal... maybe here: https://github.com/flyteorg/flytestdlib/blob/master/utils/marshal_utils.go but it's ok if it's within flyteadmin utils somewhere?

pkg/manager/impl/validation/task_validator.go
```diff
@@ -43,6 +42,26 @@ func validateContainer(task core.TaskTemplate, taskConfig runtime.TaskResourceCo
...
+	for _, container := range podSpec.Containers {
+		if errs := validation.IsDNS1123Label(container.Name); len(errs) > 0 {
```
EngHabu, line 58: Why only validate the name? do we have a dependency on that later? or just trying to prevent issues as early as possible? 
Should that entire function move into the compiler now that Pods are first class citizens?
Also should we use a more elaborate [validation function](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/validation/objectmeta.go#L143) if we are trying to prevent runtime errors as much as possible?

pkg/manager/impl/validation/task_validator.go
katrogan: @EngHabu done https://github.com/flyteorg/flytestdlib/pull/90, please review

pkg/manager/impl/validation/task_validator.go
katrogan: @EngHabu just trying to prevent issues as soon as possible 
good point re: the compiler, will close out this PR in favor of https://github.com/flyteorg/flytepropeller/pull/286

i looked into k8s pod spec validation functions but there wasn't anything servicable since the spec can be incomplete at the time of registration so I'm not sure if the above linked validation function is necessarily useful at the compilation stage

katrogan:
closing in favor of https://github.com/flyteorg/flytepropeller/pull/286

PR #223, opened 2021-07-20
Customizable user namespace templating

katrogan:

# TL;DR
This change allows a user to explicitly specify the templating used for workflow execution namespaces (which is then further customizable by the cluster resource controller). Because it was previously undocumented I don't think it's worth carrying over backwards compatibility with the existing enum. I'll work with L5 to make sure they're aware of the changes when they're released since it will affect them.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1265

## Follow-up issue
_NA_


wild-endeavor (approved):

flyteadmin_config.yaml
```diff
@@ -200,3 +200,5 @@ qualityOfService:
     development: LOW
     staging: MEDIUM
     # by default production has an UNDEFINED tier when it is omitted from the configuration
+namespace_mapping:
+   template: "{{ project }}-{{ domain }}" # Default namespace mapping template.
```
erowan, line 204: How about adding the prefix to the default template which defaults to an empty string?
"{{prefix}}{{ project }}-{{ domain }}"

& add project-prefix as an additional cmd line option to the project & file registration

wild-endeavor: if the prefix isn't dynamic, why can't we statically define it in here?
```
tempate: "flyteatjpm-{{ project }}-{{ domain }}"
```

katrogan: resolved offline - we'll use static prefixes

PR #228, opened 2021-07-30
Flyte native scheduler

pmahindrakar-oss:

# TL;DR

## Fixed rate scheduler

*  1 min cadence.

```
 ---------------------- ----------------------------------- ------------- ---------------------- 
| NAME (9)             | LAUNCH PLAN NAME                  | TYPE        | SCHEDULED TIME       |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fbffab7495702ed09000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:44:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f31d31318686f42a9000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:44:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fd5c888bd200ef680000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:44:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fa8d9b4c9ef425152000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:43:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| ff4f2f8bbe32ceab2000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:43:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f12bcd85edc5ff625000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:43:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fc38270d37fa972cb000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:42:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f5f22b3efca3c33b5000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:42:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fb502602c92445df2000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:42:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
```

* crashed the scheduler
* After catchup
```
  ---------------------- ----------------------------------- ------------- ---------------------- 
| NAME (18)            | LAUNCH PLAN NAME                  | TYPE        | SCHEDULED TIME       |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fb456594fc8cb4fbd000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:47:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f1f861383afdb98bf000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:46:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f1ea0fd62f564dab4000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:45:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fe8996f330ba48444000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:47:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f03f52c3d9b0aa7bd000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:46:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fb285e664ffd80864000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:45:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f3e431986c8d076fb000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:47:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f51ccdfd59077ee4a000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:46:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fd4cb9908dd0f3676000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:45:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fbffab7495702ed09000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:44:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f31d31318686f42a9000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:44:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fd5c888bd200ef680000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:44:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fa8d9b4c9ef425152000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:43:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| ff4f2f8bbe32ceab2000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:43:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f12bcd85edc5ff625000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:43:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fc38270d37fa972cb000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:42:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f5f22b3efca3c33b5000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:42:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fb502602c92445df2000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:42:24Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
```

## Cron expression scheduler

* Ran the scheduler for cron expression every 25 sec schedule ( */25 * * * * *)

```
 ---------------------- ----------------------------------- ------------- ---------------------- 
| NAME (6)             | LAUNCH PLAN NAME                  | TYPE        | SCHEDULED TIME       |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fed7a632a5218e834000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:51:00Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f4b189b7cca73a02d000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:51:00Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fd97fc141db51a03f000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:51:00Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fc96581f09367d187000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:50:50Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fd553da3f135c7fa3000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:50:50Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f0bf89f19b7d96c21000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:50:50Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
```

* Brought down the scheduler 

* catch up and follow

```
 ---------------------- ----------------------------------- ------------- ---------------------- 
| NAME (24)            | LAUNCH PLAN NAME                  | TYPE        | SCHEDULED TIME       |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fe59d6b2c1977372c000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:53:00Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f13b9b20d06d4e7d5000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:53:00Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f4fa8c4eebfa8cac9000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:53:00Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f272dc5db62f2c2e9000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:52:50Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fdb3d4afddb96a53c000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:52:25Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fdc9942786eea01b6000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:52:00Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f9638a644dc67a04f000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:51:50Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f5a30de5ec35f491e000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:51:25Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fd399b7aaf239b4c9000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:52:50Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f87669715acaa5d9d000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:52:25Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fa05028b18a6b3d18000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:52:00Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fc2a8f851b812a279000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:51:50Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fb66df7eddfd7ffa9000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:51:25Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f2d276382bae9baf8000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:52:50Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f5d2eb40e9e8e5515000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:52:25Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f02cbd9ac20e4d940000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:52:00Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fd82d15216759282e000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:51:50Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f680ed4f8f7d750e3000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:51:25Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fed7a632a5218e834000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:51:00Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f4b189b7cca73a02d000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:51:00Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fd97fc141db51a03f000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:51:00Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fc96581f09367d187000 | my_cron_scheduled_lp_cloned       | LAUNCH_PLAN | 2021-08-09T13:50:50Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| fd553da3f135c7fa3000 | my_cron_scheduled_lp              | LAUNCH_PLAN | 2021-08-09T13:50:50Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
| f0bf89f19b7d96c21000 | my_cron_scheduled_lp_clone_cloned | LAUNCH_PLAN | 2021-08-09T13:50:50Z |
 ---------------------- ----------------------------------- ------------- ---------------------- 
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/1227

## Follow-up issue
https://github.com/flyteorg/flyte/issues/1415
https://github.com/flyteorg/flyte/issues/1422


pkg/async/schedule/flytescheduler/workflow_executor.go
```diff
@@ -0,0 +1,234 @@
...
+	var err error
+	catchUpTimes, err = w.goGfInterface.GetCatchUpTimes(s, fromTime, toTime)
+	if err != nil {
+		return err
+	}
+	var catchupTime time.Time
+	for _, catchupTime = range catchUpTimes {
+		err := fire(ctx, w.executionManager, catchupTime, s)
+		if err != nil {
+			logger.Errorf(ctx, "unable to fire the schedule %v at %v time due to %v", s, catchupTime, err)
```
pmahindrakar-oss, line 103: This needs to be abort instead of continuing further

pkg/async/schedule/flytescheduler/workflow_executor.go
```diff
@@ -0,0 +1,234 @@
...
+func (w *workflowExecutor) Run() {
+	ctx := context.Background()
+	c, err := w.db.SchedulableEntityRepo().GetAll(ctx)
+	if err != nil {
+		panic(fmt.Errorf("unable to run the workflow executor after reading the schedules due to %v", err))
+	}
+	// Start the go routine to write the snapshot map to the configured snapshot file
+	go w.CheckPointState(ctx)
+	catchUpTill := time.Now()
+	err = w.CatchUpAllSchedules(ctx, c.Entities, catchUpTill)
```
pmahindrakar-oss, line 171: This can spawn go routines later for each schedule .Right now this is synchronous

pkg/async/schedule/flytescheduler/go_gf.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		scheduledTimes = append(scheduledTimes, scheduledTime)
+		currFrom = scheduledTime
+	}
+	return scheduledTimes, nil
+}
+
+func addCronJob(cronExpression string, job func(), nameOfSchedule string) error {
+	_, err := gcron.AddSingleton(cronExpression, job, nameOfSchedule)
+	if err != nil && strings.HasSuffix(err.Error(), "already exists"){
```
katrogan, line 80: this seems brittle, is there no code or anything exposed by the library?

pkg/async/schedule/flytescheduler/workflow_executor.go
katrogan: also would be good to add a sleep between calls to fire to avoid overwhelming admin/k8s

pkg/async/schedule/flytescheduler/workflow_executor.go
```diff
@@ -0,0 +1,234 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	executionRequest := admin.ExecutionCreateRequest{
+		Project: s.Project,
+		Domain:  s.Domain,
+		Name:    "f" + strings.ReplaceAll(uuid.New().String(), "-", "")[:19],
```
katrogan, line 133: don't we want this name to be deterministic based on the kickoff/scheduled time?

pkg/async/schedule/flytescheduler/workflow_executor.go
```diff
@@ -0,0 +1,234 @@
...
+	snapshot             flyteSchedulerInterfaces.Snapshot
+	snapShotReaderWriter flyteSchedulerInterfaces.SnapshotReaderWriter
+	goGfInterface        flyteSchedulerInterfaces.GoGFWrapper
+}
+
+func (w *workflowExecutor) CheckPointState(ctx context.Context) {
+	workflowExecConfig := w.config.ApplicationConfiguration().GetSchedulerConfig().WorkflowExecutorConfig.FlyteWorkflowExecutorConfig
+	fileName := workflowExecConfig.SnapshotFileName
+	for true {
+		f, err := os.OpenFile(fileName, os.O_WRONLY | os.O_CREATE, 0644)
```
katrogan, line 40: why are we using a file to store state? this won't be recoverable if we bring up a new instance of a scheduler pod, right? can we use the database to store our marker?

pkg/async/schedule/flytescheduler/workflow_executor.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	defer logger.Infof(ctx, "Exiting Workflow executor")
+	for true {
+		for _, s := range c.Entities {
+			funcRef := func() {
+				nameOfSchedule := w.goGfInterface.GetScheduleName(s)
+				lastT := w.snapshot.GetLastExecutionTime(nameOfSchedule)
+				n := time.Now()
+				err = w.CatchUpSingleSchedule(ctx, s, n)
```
katrogan, line 183: I'm confused, doesn't the call above to w.CatchUpAllSchedules (line 171) call CatchUpSingleSchedule for each schedule entity?

pkg/async/schedule/flytescheduler/workflow_executor.go
```diff
@@ -0,0 +1,234 @@
...
+				if err != nil {
+					logger.Errorf(ctx, "unable to catch on schedule %v from lastT %v to %v with jitter configured due to %v", s.Name, lastT, n, err)
+				}
+			}
+			err := w.goGfInterface.Register(ctx, s, funcRef)
```
katrogan, line 188: why are we registering the catchup job as a gogf cron?

pkg/async/schedule/flytescheduler/workflow_executor.go
```diff
@@ -0,0 +1,234 @@
...
+		snapShotReaderWriter: &snapShotReaderWriter,
+		goGfInterface:        GoGF{}}
+}
+
+func readSnapShot(ctx context.Context, fileName string, version int) flyteSchedulerInterfaces.Snapshot {
+	var snapshot flyteSchedulerInterfaces.Snapshot
+	f, err := os.Open(fileName)
+	// Just log the error but dont interrupt the startup of the scheduler
+	if err != nil {
+		logger.Errorf(ctx, "unable to read the snapshot file %v", fileName)
```
katrogan, line 222: nit: log the error too?

pkg/repositories/models/schedulable_entity.go
```diff
@@ -0,0 +1,31 @@
...
+// Database model to encapsulate metadata associated with a SchedulableEntity
+type SchedulableEntity struct {
+	BaseModel
+	SchedulableEntityKey
+	CronExpression      string
+	FixedRateValue		uint32
+	Unit                admin.FixedRateUnit
+	LastExecutionTime   *time.Time
+	KickoffTimeInputArg string
+	Active              bool
```
katrogan, line 17: i think you want this to be a pointer so gorm actually updates the model when you deactivate a schedule, see https://gorm.io/docs/create.html#Default-Values

pkg/repositories/interfaces/schedulable_entity_repo.go
```diff
@@ -0,0 +1,25 @@
+package interfaces
+
+import (
+	"context"
+	"github.com/flyteorg/flyteadmin/pkg/repositories/models"
+)
+
+//go:generate mockery -name=SchedulableEntityRepoInterface -output=../mocks -case=underscore
+
+type SchedulableEntityRepoInterface interface {
```
katrogan, line 10: should there be a deactivate here too?

pkg/repositories/interfaces/schedulable_entity_repo.go
```diff
@@ -0,0 +1,25 @@
...
+	Create(ctx context.Context, input models.SchedulableEntity) error
+
+	// Get a schedulable entity from the database store using the schedulable entity id.
+	Get(ctx context.Context, ID models.SchedulableEntityKey) (models.SchedulableEntity, error)
+
+	// GetAll Gets all the active schedulable entities from the db
+	GetAll(ctx context.Context) (models.SchedulableEntityCollectionOutput, error)
+
+	// Update a schedulable entity with
+	UpdateLastExecution(ctx context.Context, input models.SchedulableEntity) error
```
katrogan, line 21: how is this used?

pkg/async/schedule/flytescheduler/workflow_executor.go
pmahindrakar-oss: I have added a rate limiter to admin which defines how many requests per sec should be sent to the admin

pkg/async/schedule/flytescheduler/workflow_executor.go
pmahindrakar-oss: Makes sense. I was assuming we would have access to persistent volumes to our pods. If thats not the case then yes storing it the DB would be right thing to do for now.
I will update on this

pkg/async/schedule/flytescheduler/workflow_executor.go
pmahindrakar-oss: done. Thanks

pkg/repositories/interfaces/schedulable_entity_repo.go
pmahindrakar-oss: Its not used now. Removing it.  I removed from the model but forgot to update the interface

pkg/repositories/interfaces/schedulable_entity_repo.go
pmahindrakar-oss: Yes i have named it as delete , But deactivate seems right name to use since we are technically just toggling the flag

pkg/repositories/models/schedulable_entity.go
pmahindrakar-oss: Thanks for pointing that out

pkg/async/schedule/aws/cloud_watch_scheduler.go
```diff
@@ -170,6 +172,33 @@ func (s *cloudWatchScheduler) AddSchedule(ctx context.Context, input scheduleInt
...
+		schedule.GetKickoffTimeInputArg(),
+		identifier)
+	if err != nil {
+		logger.Errorf(ctx, "failed to serialize schedule workflow payload for launch plan: %v with err: %v",
+			identifier, err)
+		return scheduleInterfaces.AddScheduleInput{}, err
+	}
+
+	// Backward compatible with old EvenSchedulerConfig structure
+	scheduleNamePrefix := appConfig.EventSchedulerConfig.ScheduleNamePrefix
```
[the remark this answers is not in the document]
pmahindrakar-oss: Agreed. Adding those

pkg/async/schedule/flytescheduler/workflow_executor.go
pmahindrakar-oss: https://github.com/uber-go/ratelimit

pkg/async/schedule/aws/cloud_watch_scheduler.go
```diff
@@ -68,15 +70,15 @@ type cloudWatchScheduler struct {
 	metrics cloudWatchSchedulerMetrics
 }
 
-func getScheduleName(scheduleNamePrefix string, identifier admin.NamedEntityIdentifier) string {
-	hashedIdentifier := hashIdentifier(identifier)
+func getScheduleName(scheduleNamePrefix string, identifier core.Identifier) string {
+	hashedIdentifier := HashIdentifier(identifier)
```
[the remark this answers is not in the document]
pmahindrakar-oss: Inorder to avoid the complexity of backward compatibility , should we continue to support AWS schedule identifier using project,domain, name

And only for the new native scheduler user project,domain,name and version
That way even though the interface is passing core.Identifier 
only the native scheduler will use the version fields whereas aws will continue to use it without the version.
Let me know what you think

pkg/async/schedule/flytescheduler/flyte_scheduler.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+	logger.Debug(ctx, "Created a scheduled entity for %v ", input)
+	return nil
+}
+
+func (s *FlyteScheduler) RemoveSchedule(ctx context.Context, input interfaces.RemoveScheduleInput) error {
+	logger.Debugf(ctx, "Received call to remove schedule [%+v]", input.Identifier)
+
+	//schedulableEntity, err := s.db.SchedulableEntityRepo().Get(ctx, models.SchedulableEntityKey{
```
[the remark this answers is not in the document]
pmahindrakar-oss: This is untested. This would be called deactivate

pkg/async/schedule/flytescheduler/flyte_scheduler.go
```diff
@@ -0,0 +1,84 @@
+package flytescheduler
```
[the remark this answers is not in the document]
pmahindrakar-oss: Sure. Will update this

pkg/async/schedule/flytescheduler/go_gf.go
```diff
@@ -0,0 +1,137 @@
...
+	"strconv"
+	"strings"
+	"time"
+)
+
+type GoGF struct {
+}
+
+func (g GoGF) GetScheduleName(s models.SchedulableEntity) string {
+	return strconv.FormatUint(aws.HashIdentifier(core.Identifier{
```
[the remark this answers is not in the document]
pmahindrakar-oss: Ok as per the previous comment here https://github.com/flyteorg/flyteadmin/pull/228#discussion_r682547294
i would prefer to use a new function for generation hash with version to avoid backward compat issues with aws scheduler. Let me know if you agree

pkg/async/schedule/flytescheduler/go_gf.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+		} else {
+			err := addFixedIntervalJob(s.Unit, s.FixedRateValue, funcRef)
+			if err != nil {
+				logger.Errorf(ctx, "failed to add fixed rate schedule %v due to %v", s, err)
+			}
+		}
+	} else {
+		if len(s.CronExpression) > 0 {
+			removeCronJob(ctx, nameOfSchedule)
```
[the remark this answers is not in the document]
pmahindrakar-oss: Register from GoGF perspective is any new schedule that it receives which can either be an activation in which case it will add or a deactivation in which case it will remove.

Do we want to name this something else ?

pkg/async/schedule/flytescheduler/go_gf.go
pmahindrakar-oss: Sure. will use gerror.CodeInvalidOperation https://github.com/gogf/gf/blob/e9ea58df64c297f89eec633e3ee4eda4b290c8f1/os/gcron/gcron_cron.go#L65

pkg/async/schedule/flytescheduler/workflow_executor.go
pmahindrakar-oss: So the captupallschedules happens in the main scheduler routine during bootup
and uses the time.Now when this is executed.
```
catchUpTill := time.Now()
err = w.CatchUpAllSchedules(ctx, c.Entities, catchUpTill)
``
So here all the schedules are caught up to catchUpTill


But when this function is invoked because of cron / regular cadence
```
funcRef := func() {
				nameOfSchedule := w.goGfInterface.GetScheduleName(s)
				lastT := w.snapshot.GetLastExecutionTime(nameOfSchedule)
				n := time.Now()
				err = w.CatchUpSingleSchedule(ctx, s, n)
				if err != nil {
					logger.Errorf(ctx, "unable to catch on schedule %v from lastT %v to %v with jitter configured due to %v", s.Name, lastT, n, err)
				}
			}
```

At that time `n := time.Now()` would be different and we will catch the schedule from `lastT to n`

Does that answer your question?

File: pkg/async/schedule/flytescheduler/workflow_executor.go (hunk 7 above)
Comment (pmahindrakar-oss) [reply pkg/async/schedule/flytescheduler/workflow_executor.go:188]: This can either be a cron or regular cadence work .
We are registering here every time we read new schedules from the DB.
IF the schedule already exists in the memory , then based on isActive it will either try to  add the schedule which will fail(already exists) or remove if isActive is false.

Does that answer your question?

File: pkg/async/schedule/flytescheduler/workflow_executor.go (hunk 4 above)
Comment (pmahindrakar-oss) [reply pkg/async/schedule/flytescheduler/workflow_executor.go:133]: I was trying to keep this consistent with other places where we are generating the UUID for the executions .
The UUID on the details does give us the scheduleTime in its metadata field .

Do we want something 

like instead of this  `f09f1395a2d824cf9800` to have  this `2021-08-04T17:07:37+05:30-f09f1395a2d824cf9800`
or some other form.

File: pkg/async/schedule/flytescheduler/go_gf.go (hunk 3 above)
Comment (pmahindrakar-oss) [reply pkg/async/schedule/flytescheduler/go_gf.go:80]: This change just got  added in latest version 15 days ago so using the latest to use error code now

File: pkg/async/schedule/flytescheduler/workflow_executor.go (hunk 5 above)
Comment (pmahindrakar-oss) [reply pkg/async/schedule/flytescheduler/workflow_executor.go]: Done

File: pkg/async/schedule/flytescheduler/go_gf.go (hunk 3 above)
Comment (pmahindrakar-oss) [reply pkg/async/schedule/flytescheduler/go_gf.go]: Done

File: pkg/async/schedule/flytescheduler/interfaces/go_gf_wrapper.go [hunk 18]
```diff
@@ -0,0 +1,15 @@
+package interfaces
+
+import (
+	"context"
+	"github.com/flyteorg/flyteadmin/pkg/repositories/models"
+	"time"
+)
+
+type GoGFWrapper interface {
```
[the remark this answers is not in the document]
pmahindrakar-oss: Done

pkg/async/schedule/flytescheduler/go_gf.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/flyteorg/flytestdlib/logger"
+	"github.com/gogf/gf/os/gcron"
+	"github.com/gogf/gf/os/gtimer"
+	"github.com/robfig/cron"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type GoGF struct {
```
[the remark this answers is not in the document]
pmahindrakar-oss: Done.

pkg/async/schedule/flytescheduler/flyte_scheduler.go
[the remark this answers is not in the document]
pmahindrakar-oss: Done

pkg/async/schedule/flytescheduler/flyte_scheduler.go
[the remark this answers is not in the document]
pmahindrakar-oss: Done

pkg/async/schedule/aws/cloud_watch_scheduler.go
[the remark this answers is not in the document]
pmahindrakar-oss: Done

pkg/async/schedule/flytescheduler/workflow_executor.go
pmahindrakar-oss: Done

pkg/async/schedule/aws/cloud_watch_scheduler.go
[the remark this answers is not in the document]
katrogan: i like this proposal!

pkg/async/schedule/flytescheduler/go_gf.go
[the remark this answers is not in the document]
katrogan: got it, maybe a comment to explain?

pkg/async/schedule/flytescheduler/workflow_executor.go
katrogan: a hash of the input time is fine. if we generate a new uuid every time though we don't get idempotency from repeated calls to trigger a scheduled execution

pkg/async/schedule/flytescheduler/workflow_executor.go
katrogan: i think so, thanks

pkg/async/schedule/flytescheduler/workflow_executor.go
katrogan: yes, thanks

flyteadmin_config.yaml
```diff
@@ -69,12 +69,17 @@ database:
...
+    local:
     region: "my-region"
     scheduleRole: "arn:aws:iam::abc123:role/my-iam-role"
     targetName: "arn:aws:sqs:my-region:abc123:my-queue"
     scheduleNamePrefix: "flyte"
   workflowExecutor:
     scheme: local
+    local:
+      snapshotVersion: 1 # version of the snapshot
+      jitterValue: 30 # jitter or delta from the current time to use for finding the schedule
```
kumare3, line 81: can we sane default values for these

pkg/async/schedule/aws/cloud_watch_scheduler.go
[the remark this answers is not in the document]
kumare3: yes please, I think the unique id creation logic should reside with the actual scheduler

flyteadmin_config.yaml
kumare3: and by default i mean, we should not need to add these values to our config at all!

pkg/async/schedule/factory.go
```diff
@@ -42,21 +44,36 @@ func (w *workflowScheduler) GetEventScheduler() interfaces.EventScheduler {
 }
 
 func (w *workflowScheduler) GetWorkflowExecutor(
+	db repositories.RepositoryInterface,
```
kumare3, line 47: this seems like a very broad interface that we are passing here? Can we reduce it? If you want you can create a local interface too

pkg/async/schedule/factory.go
```diff
@@ -81,26 +98,15 @@ func NewWorkflowScheduler(cfg WorkflowSchedulerConfig) WorkflowScheduler {
 			cfg.SchedulerConfig.EventSchedulerConfig.ScheduleRole, cfg.SchedulerConfig.EventSchedulerConfig.TargetName, sess, awsConfig,
 			cfg.Scope.NewSubScope("cloudwatch_scheduler"))
 	case common.Local:
-		fallthrough
-	default:
 		logger.Infof(context.Background(),
-			"Using default noop event scheduler implementation for cloud provider type [%s]",
+			"Using default flyte scheduler implementation for cloud provider type [%s]",
```
kumare3, line 102:
```suggestion
			"Using default flyte scheduler implementation ",
```

pkg/async/schedule/flytescheduler/go_gf.go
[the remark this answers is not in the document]
kumare3: My preference is to move the entire scheduler (the actual daemon) to be under
flyteadmin/scheduler/....

the way of doing flyteadmin/pkg/... is not encouraged by go conventions

Can we do that? and then refer to it here

pkg/async/schedule/flytescheduler/flyte_scheduler.go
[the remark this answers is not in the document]
kumare3: I think we should move flytescheduler to be a top level golang pkg under flyteadmin/flytescheduler. This is a nicer way to separate things than club everything under pkg (this was an old way in golang)

pkg/async/schedule/flytescheduler/go_gf.go
```diff
@@ -0,0 +1,139 @@
...
+func addCronJob(cronExpression string, job func(), nameOfSchedule string) error {
+	_, err := gcron.AddSingleton(cronExpression, job, nameOfSchedule)
+	if err != nil && gerror.Code(err) == gerror.CodeInvalidOperation {
+		return nil
+	}
+	return err
+}
+
+func removeCronJob(ctx context.Context, nameOfSchedule string) {
+	if e := gcron.Search(nameOfSchedule); e != nil {
```
kumare3, line 88: are these thread safe methods? and thus are they re-entrant or idempotent

pkg/async/schedule/flytescheduler/go_gf.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return err
+	}
+	gtimer.AddSingleton(d, job)
+	return nil
+}
+
+func removeFixedIntervalJob(ctx context.Context, nameOfSchedule string) {
+	// TODO : find the right way to remove the fixed interval job
+	logger.Infof(ctx, "successfully remove the schedule %v from scheduler", nameOfSchedule)
```
kumare3, line 105:
```suggestion
	logger.Infof(ctx, "successfully removed the schedule %s from scheduler", nameOfSchedule)
```

pkg/async/schedule/flytescheduler/interfaces/go_gf_wrapper.go
```diff
@@ -0,0 +1,19 @@
+package interfaces
+
+import (
+	"context"
+	"github.com/flyteorg/flyteadmin/pkg/repositories/models"
+	"time"
+)
+
+// GoGFWrapper Wrapper interface to the gogf framework which is used for scheduler
```
kumare3, line 9: this comment does not explain much, why do we need a wrappeR?

kumare3: For example, we should write that this wrapper handles scheduler common logic like jitter subtraction, time locking etc

pkg/async/schedule/flytescheduler/interfaces/go_gf_wrapper.go
```diff
@@ -0,0 +1,19 @@
...
+type GoGFWrapper interface {
+	// Register add and remove cron or fixed rate schedules from gogf
+	Register(ctx context.Context, s models.SchedulableEntity, funcRef func()) error
```
kumare3, line 12: Shouldnt the funcRef be of the format
```
funcRef func(ctx context.Context, scheduleTime time.Time) error
```
This makes it the most future proof. Context is always a good thing to add, this way we can allow graceful termination. Also schedule_time should be handled in our wrapper, Ideally the cron system should have done it, but sadly we have to. And we can have a common way of handling errors.

pkg/async/schedule/flytescheduler/snapshot_v1.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"bytes"
+	"encoding/gob"
+	"github.com/flyteorg/flyteadmin/pkg/async/schedule/flytescheduler/interfaces"
+	"reflect"
+	"time"
+)
+
+type SnapshotV1 struct {
```
kumare3, line 11: comments, lets explain why v1

pkg/async/schedule/flytescheduler/workflow_executor.go
kumare3: we should not use UUID, to make it idempotent, it has to be deterministic

pkg/async/schedule/flytescheduler/workflow_executor.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+func (w *workflowExecutor) Run() {
+	ctx := context.Background()
+	c, err := w.db.SchedulableEntityRepo().GetAll(ctx)
+	if err != nil {
+		panic(fmt.Errorf("unable to run the workflow executor after reading the schedules due to %v", err))
+	}
+	// Start the go routine to write the snapshot map to the configured snapshot file
+	go w.CheckPointState(ctx)
```
kumare3, line 169: can we handle, error and panic here. this go routine may error out or have a panic this should be handled. This can be done with some helper methods from k8s apiachinery too

kumare3: I do not think we should start the checkpointer, till we catchup?

pkg/async/schedule/flytescheduler/workflow_executor.go
```diff
@@ -0,0 +1,251 @@
...
+	catchUpTill := time.Now()
+	err = w.CatchUpAllSchedules(ctx, c.Entities, catchUpTill)
+	if err != nil {
+		panic(fmt.Errorf("unable to catch up on the schedules due to %v", err))
+	}
+
+	defer logger.Infof(ctx, "Exiting Workflow executor")
+	for true {
+		for _, s := range c.Entities {
+			funcRef := func() {
```
kumare3, line 179: lets pass in a context here

kumare3: and schedule time

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -93,6 +191,48 @@ type WorkflowExecutorConfig struct {
...
+func (a *AWSWorkflowExecutorConfig) GetScheduleScheduleQueueName() string {
+	return a.ScheduleQueueName
+}
+
+func (a *AWSWorkflowExecutorConfig) GetAccountID() string {
+	return a.AccountID
+}
+
+// FlyteWorkflowExecutorConfig specifies the workflow executor configuration for the native flyte scheduler
+type FlyteWorkflowExecutorConfig struct {
```
kumare3, line 207: please add default values for all these values.

cc @katrogan / @EngHabu any new config should have a default value, else we have to go through a justification

kumare3:
A few high level comments
1. Move the actual scheduler to a top level package
2. This package should have a doc.go that explains the workings of the scheduler, with the algorithm, catchup etc. A few things to note - 
         Various assumptions. (min schedule frequency)
         The scheduler is a singleton. The deployment should have one replica. If multiple runs, it just duplicates the work.
         Catchup system
         Snapshotting system, with support for versioning
3. The GoGFWrapper should lock the time and calculate the jitter
4. The actual Job function should get a context and schedule time
5. the go-routine for the job function should handle a panic (maybe crash, but log and set a metric)
6. The go routine for snapshotter should handle panic (maybe crash but log / set a metric)

pkg/async/schedule/flytescheduler/interfaces/go_gf_wrapper.go
pmahindrakar-oss: Can we discuss more over this on chat.
the semantic of the func is that it would run a catch up between last execution time and the current triggered time of the job func.
So it can be not just one schedule time but a list of schedule times that this one func can be doing.

pkg/async/schedule/flytescheduler/workflow_executor.go
pmahindrakar-oss: I have moved the checkpointer after catchup, the reason i kept it before is so that everytime catch up runs it makes some progress in case we have continuous crashes or stop/restart. 

Eg : If the catch was to be done for a 1 week , then checkpointing before can be useful. Let me know what you think. 1 week doesn't sounds reasonable so i am ok with having the checkpointer after the catchup

pkg/async/schedule/flytescheduler/workflow_executor.go
pmahindrakar-oss: Done. Instead of schedule time i am passing the schedule and also the semantics of this func as explained in my previous comment is not to send execution for a single schedule time but over all the times between last execution time and current time when this func gets triggered.


That way we also are able to handle cases where some goroutine are running slow for some reason and hence we just use catch everytime

pkg/async/schedule/flytescheduler/snapshot_v1.go
pmahindrakar-oss: Done

flyteadmin_config.yaml
pmahindrakar-oss: Moved jitter as a const in the scheduler

scheduler/workflow_executor.go
```diff
@@ -0,0 +1,279 @@
...
+		return err
+	}
+	return nil
+}
+
+func (w *workflowExecutor) Run() {
+	ctx := context.Background()
+	schedules, err := w.db.SchedulableEntityRepo().GetAll(ctx)
+	if err != nil {
+		panic(fmt.Errorf("unable to run the workflow executor after reading the schedules due to %v", err))
```
kumare3, line 194: no panic, return an error.

pkg/runtime/interfaces/application_configuration.go
kumare3: did you add the default values?

scheduler/executor/doc.go
```diff
@@ -0,0 +1,97 @@
...
+// Flyte scheduler implementation that allows to schedule fixed rate and cron schedules on sandbox deployment
+// Scheduler has two components
+// 1] Schedule management
+//    This component is part of the pkg/async/schedule/flytescheduler package
+//    Role of this component is to create / activate / deactivate schedules
+//    The above actions are exposed through launchplan activation/deactivation api's and donot have separate controls.
+//    Whenever a launchplan with a schedule is activated, a new schedule entry is created in the datastore
+//    On deactivation the created scheduled and launchplan is deactivated through a flag
+//    Atmost one launchplan is active at any moment across its various versions and same semantics apply for the
+//    schedules aswell.
```
kumare3, line 11:
```suggestion
//    schedules as well.
```

scheduler/executor/doc.go
```diff
@@ -0,0 +1,97 @@
...
+// 2] Scheduler
+//    This component is a singleton and has its source in the current folder and is responsible for reading the schedules
+//    from the DB and running them at the cadence defined by there schedule
```
kumare3, line 14:
```suggestion
//    from the DB and running them at the cadence defined by the schedule
```

scheduler/executor/doc.go
```diff
@@ -0,0 +1,97 @@
...
+//    The lowest granularity supported is minutes for scheduling through cron and fixed rate scheduler
+// 	  The scheduler should be running in one replica , two at the most during redeployment. Multiple replicas will just
+// 	  duplicate the work since each execution for a scheduleTime will have unique identifier derived from schedule name
+//	  and time of the schedule. The idempotency aspect of the admin for same identifier prevents duplication on the admin
+//	  side.
+//    The scheduler runs continously in a loop reading the updated schedule entries in the data store and adding or removing
+//    the schedules. Removing a schedule doesn't gurantee about inflight routines launched by the scheduler.
```
kumare3, line 21:
```suggestion
//    the schedules. Removing a schedule will not alter in-flight go-routines launched by the scheduler. Thus the behavior of these executions is undefined (most probably will get executed).
```

scheduler/executor/doc.go
```diff
@@ -0,0 +1,97 @@
...
+//		a) Snapshoter
+// 			This component is responsible for writing the snapshot state of all the schedules at a regular cadence to a
+//			persistent store. The current implementation uses DB to store the GOB format of the snapshot which is versioned.
+//			The snapshot is map[string]time.Time which stores a map of schedules names to there last execution times
+// 			During bootup the snapshot is bootstraped from the data store and loaded in memory
+//			The Scheduler use this snapshot to schedule any missed schedules.
+//
+//			We cannot use global snapshot time since each time snapshot doesn't contain information on how many schedules
+//			were executed till that point in time. And hence the need to maintain map[string]time.Time of schedules to there
+//			lastExectimes
```
kumare3, line 32:
```suggestion
//			lastExectimes
//   In the future we may support global snapshots, such that we can record the last successfully considered time for each schedule and select the lowest as the watermark. currently since the underlying scheduler does not expose the last considered time, we just calculate our own watermark per schedule.
```

scheduler/executor/doc.go
```diff
@@ -0,0 +1,97 @@
...
+// 		b) Catchuper :
```
kumare3, line 33:
```suggestion
// 		b) Catchup-System :
```

scheduler/executor/versioned_snapshot.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"encoding/gob"
+	"fmt"
+	"github.com/flyteorg/flyteadmin/scheduler/executor/interfaces"
+	"io"
+	"time"
+)
+
+type VersionedSnapshot struct {
```
kumare3, line 11: docs please. But this is awesome!

scheduler/executor/workflow_executor.go
```diff
@@ -0,0 +1,349 @@
...
+		logger.Debugf(ctx, "caught up successfully on the schedule %+v from %v to %v", s, fromTime, toTime)
+	}
+	return nil
+}
+
+func (w *workflowExecutor) CatchUpSingleSchedule(ctx context.Context, s models.SchedulableEntity, fromTime time.Time, toTime time.Time) error {
+	nameOfSchedule := GetScheduleName(s)
+	var catchUpTimes []time.Time
+	var err error
+	catchUpTimes, err = w.goGfInterface.GetCatchUpTimes(s, fromTime, toTime)
```
kumare3, line 125: this is not goGCInterface anymore right?

scheduler/executor/workflow_executor.go
```diff
@@ -0,0 +1,349 @@
...
+	snapshot             interfaces2.Snapshoter
+	snapShotReaderWriter interfaces2.SnapshotReaderWriter
+	goGfInterface        interfaces2.GoCronWrapper
+	rateLimiter          ratelimit.Limiter
+	metrics              schedulerMetrics
+	adminServiceClient service.AdminServiceClient
+}
+
+func (w *workflowExecutor) CheckPointState(ctx context.Context) {
+	defer func() {
```
kumare3, line 60: a better pattern would be put this differ at the call-site. Like in the previous method. then the code here will raise panics and you can test that

scheduler/executor/workflow_executor.go
```diff
@@ -0,0 +1,349 @@
...
+		panic(fmt.Errorf("unable to run the workflow executor after reading the schedules due to %v", err))
+	}
+	catchUpTill := time.Now()
+	err = w.CatchUpAllSchedules(ctx, schedules, catchUpTill)
+	if err != nil {
+		panic(fmt.Errorf("unable to catch up on the schedules due to %v", err))
+	}
+
+	// Start the go routine to write the snapshot map to the configured snapshot file
+	go w.CheckPointState(ctx)
```
kumare3, line 246: handle panic here

scheduler/executor/workflow_executor.go
```diff
@@ -0,0 +1,349 @@
...
+			} else {
+				err := w.goGfInterface.Register(ctx, s, funcRef)
+				if err != nil {
+					w.metrics.ScheduleRegistrationFailure.Inc()
+					logger.Errorf(ctx, "unable to register the schedule %+v due to %v", s, err)
+				}
+			}
+		}
+		time.Sleep(executorSleepTime * time.Second)
+		schedules, err = w.db.SchedulableEntityRepo().GetAll(ctx)
```
kumare3, line 279: this looks a little weird?
Maybe refactor the code to have the second part in a separate method? 
RunBackgroundSync?

scheduler/executor/workflow_executor.go
```diff
@@ -0,0 +1,349 @@
...
+		if err != nil {
+			w.metrics.ScheduleReadFailure.Inc()
+			logger.Errorf(ctx, "going to sleep additional %v backoff time due to DB error %v", backOffSleepTime, err)
+			time.Sleep(backOffSleepTime * time.Second)
+		}
+	}
+}
+
+func (w *workflowExecutor) Stop() error {
```
kumare3, line 288: this is important. Today your exit is not graceful. You can do that by having a bunch of this background threads using a 

for {
  select {
  }
}
loop. And you can use a Done channel in context to stop the Run command.
So instead of having a Stop, just change run to be

Run(ctx: context.Context)

And delete Stop. I can share examples here

kumare3:
This looks pretty good to me. Couple things the Run methods needs to be better with graceful termination.

scheduler/executor/workflow_executor.go
pmahindrakar-oss: renamed from goGfInterface  -> goCronInterface
We are still using the interface to get a slice of catch up times.
Do you have any concerns here?

scheduler/workflow_executor.go
pmahindrakar-oss: Done

scheduler/executor/workflow_executor.go
```diff
@@ -0,0 +1,431 @@
...
+				w.metrics.ScheduleRegistrationFailure.Inc()
+				logger.Errorf(ctx, "unable to register the schedule %+v due to %v", s, err)
+			}
+		}
+	} // Done iterating over all the read schedules
+	return nil
+}
+
+func (w *workflowExecutor) RunCatchupSystem(ctx context.Context) chan error {
+	catchupCtx, cancelCatchUp := context.WithCancel(ctx)
```
kumare3, line 283: What's the use of the cancel
Catchup method?

scheduler/executor/workflow_executor.go
```diff
@@ -0,0 +1,431 @@
...
+				logger.Errorf(ctx, "Failed to iterated over all schedules in the current run due to %v", err)
+			}
+		}, registrationSleepTime * time.Second)
+	}(scheduleRegisterCtxWithLabel)
+}
+
+func (w *workflowExecutor) Run(ctx context.Context) error {
+	// Start the go routine to catchup all the schedules
+	// Create catchup function label to be used for creating the child context
+	catchUpErrChan := w.RunCatchupSystem(ctx)
```
kumare3, line 341: You should create the cancel func here and defer it so that in case any other goroutine returns you can cancel others

scheduler/executor/workflow_executor.go
```diff
@@ -0,0 +1,431 @@
...
+
+	// Start the go routine to write the snapshot map to the configured snapshot file
+	// Create checkpointer  label to be used for creating the child context
+	w.RunCheckpointerSystem(ctx)
+
+	// Start the go routine to write the snapshot map to the configured snapshot file
+	// Create checkpointer  label to be used for creating the child context
```
kumare3, line 348: Repeated comment

pkg/runtime/interfaces/application_configuration.go
kumare3: @pmahindrakar-oss default values? Where did you add them. you can see an example [here](https://github.com/flyteorg/flyteadmin/blob/master/auth/config/config.go#L52)

scheduler/executor/go_cron.go
```diff
@@ -0,0 +1,149 @@
...
+type goCronMetrics struct {
+	Scope                  promutils.Scope
+	JobFuncPanicCounter prometheus.Counter
+}
+
+// GoCron Struct implementing the GoCronWrapper which is used by the scheduler for adding and removing schedules
+// Each scheduled job accepts scheduled time parameter which helps to know what the actual invocation time and use
+// that to send an execution to admin
+type GoCron struct {
+	jobsMap map[string]interfaces.GoCronJobWrapper
```
kumare3, line 29: this needs to be thread-safe. this is a data-race?
The Register and Deregister is called from goroutines right?

scheduler/executor/go_cron.go
```diff
@@ -0,0 +1,149 @@
...
+	c       *cron.Cron
+	metrics              goCronMetrics
+}
+
+func (g GoCron) DeRegister(ctx context.Context, s models.SchedulableEntity) {
```
kumare3, line 34: also needs to be a pointer? To work on the same object, instead of a copy.

scheduler/executor/go_cron.go
```diff
@@ -0,0 +1,149 @@
...
+		logger.Debugf(ctx, "Job doesn't exists in the map for name %v with schedule %+v  and hence already removed", nameOfSchedule, s)
+		return
+	}
+	g.jobsMap[nameOfSchedule].DeScheduleJob()
+
+	// Delete it from the jobs map
+	delete(g.jobsMap, nameOfSchedule)
+}
+
+func (g GoCron) Register(ctx context.Context, s models.SchedulableEntity, registerFuncRef interfaces.RegisterFuncRef) error {
```
kumare3, line 47: same thing

scheduler/executor/go_cron_job_wrapper.go
```diff
@@ -0,0 +1,77 @@
+package executor
+
+import (
+	"context"
+	"github.com/flyteorg/flyteadmin/scheduler/repositories/models"
+	"github.com/flyteorg/flytestdlib/logger"
+	"github.com/robfig/cron"
+)
+
+type GoCronJobWrapper struct {
```
kumare3, line 10: docs

scheduler/executor/interfaces/go_cron_wrapper.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"context"
+	"github.com/flyteorg/flyteadmin/scheduler/repositories/models"
+	"time"
+)
+
+
+type RegisterFuncRef func(ctx context.Context, schedule models.SchedulableEntity, scheduledTime time.Time)
+
+// GoCronWrapper Wrapper interface to the gogf framework which is used for scheduler. It also locks in schedule time.
```
kumare3, line 12: wrong docs. Also a little confused, what are the 2 interfaces?
GoCronJobWrapper and GoCronWrapper

scheduler/executor/schedule_checkpointer.go
```diff
@@ -0,0 +1,115 @@
...
+const snapShotVersion = 1
+
+type checkPointerMetrics struct {
+	Scope                        promutils.Scope
+	CheckPointPanicCounter       prometheus.Counter
+	CheckPointSaveErrCounter     prometheus.Counter
+	CheckPointCreationErrCounter prometheus.Counter
+}
+
+type ScheduleCheckPointer struct {
```
kumare3, line 31: it is very confusing to see the interface and the struct name the same - ScheduleCheckPointer
Can we instead call it - DBSchedulerCheckpointer

kumare3: and add docs that say, that this checkpoints to a DB.

Also i would move the interfaces to a separate folder and this implementation to a separate folder. Everything in one folder seems very messy

scheduler/executor/interfaces/schedule_checkpointer.go
```diff
@@ -0,0 +1,9 @@
+package interfaces
+
+import "context"
+
+type ScheduleCheckPointer interface {
```
kumare3, line 5:
```suggestion
type CheckPointer interface {
```
rename the folder to schedule or core

scheduler/executor/interfaces/schedule_execution_firer.go
```diff
@@ -0,0 +1,11 @@
+package interfaces
+
+import (
+	"context"
+	"github.com/flyteorg/flyteadmin/scheduler/repositories/models"
+	"time"
+)
+
+type ScheduleExecutionFirer interface {
```
kumare3, line 9: Weird name
```suggestion
type Executor interface {
```

scheduler/executor/interfaces/schedule_execution_firer.go
```diff
@@ -0,0 +1,11 @@
...
+	Fire(ctx context.Context, scheduledTime time.Time, s models.SchedulableEntity) error
```
kumare3, line 10:
```suggestion
	Execute(ctx context.Context, scheduledTime time.Time, s models.SchedulableEntity) error
```

scheduler/executor/interfaces/schedule_registerer.go
```diff
@@ -0,0 +1,7 @@
+package interfaces
+
+import "context"
+
+type ScheduleRegisterer interface {
```
kumare3, line 5:
```suggestion
type Registrar interface {
```

scheduler/executor/interfaces/snapshot_readerwriter.go
```diff
@@ -0,0 +1,11 @@
+package interfaces
+
+import "io"
+
+// SnapshotReaderWriter provides an interface to read and write the snapshot
+type SnapshotReaderWriter interface {
```
kumare3, line 6: Lets create a new folder called `snapshot`
and 2 interfaces
`Reader`
`Writer`.

so usage should be snapshot.Reader
snapshot.Writer

If you want to use both you can always right. This is like io.Writer, io.Reader. single method interfaces are beautiful

scheduler/executor/interfaces/snapshoter.go
```diff
@@ -0,0 +1,21 @@
+package interfaces
+
+import (
+	"time"
+)
+
+// Snapshoter used by the scheduler for creating, updating and reading snapshots of the schedules.
+type Snapshoter interface {
```
kumare3, line 8:
```suggestion
type Executor interface {
```
The folder `snapshot` you dont have to put everything in `interfaces`

scheduler/executor/interfaces/workflow_executor.go
```diff
@@ -0,0 +1,8 @@
+package interfaces
+
+import "context"
+
+// WorkflowExecutor Handles reading the new workflow schedules and creating executions for them.
+type WorkflowExecutor interface {
```
kumare3, line 6: whats the difference between this and Firer!!!!!!!

scheduler/executor/schedule_catchuper.go
```diff
@@ -0,0 +1,145 @@
...
+	"time"
+)
+
+type catchuperMetrics struct {
+	Scope               promutils.Scope
+	CatchupPanicCounter prometheus.Counter
+	CatchupErrCounter   prometheus.Counter
+}
+
+type ScheduleCatchuper struct {
```
kumare3, line 25: dude your names are funny - everything is an er haha

how about  the package called `catchup`
and the interface called `Do` or `Executor` or `Runner`

scheduler/executor/schedule_catchuper.go
```diff
@@ -0,0 +1,145 @@
...
+	metrics         catchuperMetrics
+	snapshoter      interfaces.Snapshoter
+	goCronInterface interfaces.GoCronWrapper
+	rateLimiter     ratelimit.Limiter
+	db              repositories.SchedulerRepoInterface
+	executionFirer  interfaces.ScheduleExecutionFirer
+}
+
+func (c *ScheduleCatchuper) RunCatchuper(ctx context.Context) chan error {
```
kumare3, line 34:
```suggestion
func (c *ScheduleCatchuper) Run(ctx context.Context) chan error {
```

scheduler/executor/schedule_catchuper.go
```diff
@@ -0,0 +1,145 @@
...
+			return err
+		} else {
+			w.snapshoter.UpdateLastExecutionTime(nameOfSchedule, catchupTime)
+		}
+	}
+
+	return nil
+}
+
+func NewScheduleCatchuper(scope promutils.Scope, snapshoter interfaces.Snapshoter,
```
kumare3, line 122: rename
```suggestion
func New...(scope promutils.Scope, snapshoter interfaces.Snapshoter,
```

scheduler/crud/flyte_scheduler.go
```diff
@@ -0,0 +1,89 @@
+package crud
```
kumare3, line 1: what does crud mean?
also 
`scheduler/crud/flyte_scheduler.go` is weird
how about
`scheduler/api.go` -> just the interface
then `scheduler/db_api/api_impl.go`

scheduler/crud/flyte_scheduler.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/flyteorg/flyteadmin/scheduler/repositories/models"
+
+	scheduleInterfaces "github.com/flyteorg/flyteadmin/pkg/async/schedule/interfaces"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
+	"github.com/flyteorg/flytestdlib/logger"
+)
+
+// FlyteScheduler used for saving the scheduler entries after launch plans are enabled or disabled.
+type FlyteScheduler struct {
```
kumare3, line 18: rename all structs

scheduler/entrypoints/root.go
```diff
@@ -0,0 +1,74 @@
+package entrypoints
```
kumare3, line 1: this code should be under
`flyteadmin/cmd/scheduler/root.go`

scheduler/entrypoints/scheduler.go
```diff
@@ -0,0 +1,86 @@
+package entrypoints
```
kumare3, line 1: also this should be under that further

kumare3:
the code needs beautification and refactor to make the api more pleasant.

I recommend move all cmd code to cmd
move interfaces to their own packages and these package can also have the implementations alongside. package is a folder in go. this improves how you import. think about the other side of the usage

names are important and hard. `er` to everything is a little confusing. Can we revisit please?

pkg/runtime/interfaces/application_configuration.go
kumare3: Default values have to be done on this change, as it will break others

pmahindrakar-oss: Done

EngHabu (commented):
Looking good! Haven't finished reviewing yet but a few passerby comments

flyteadmin_config.yaml
```diff
@@ -69,12 +69,15 @@ database:
...
     scheme: local
+    local:
     region: "my-region"
     scheduleRole: "arn:aws:iam::abc123:role/my-iam-role"
     targetName: "arn:aws:sqs:my-region:abc123:my-queue"
     scheduleNamePrefix: "flyte"
   workflowExecutor:
     scheme: local
+    local:
+      adminFireReqRateLimit: 100 #per sec how many requests to send to admin
```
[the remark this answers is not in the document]
pmahindrakar-oss: Done

cmd/scheduler/entrypoints/scheduler.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/flyteorg/flytestdlib/promutils/labeled"
+
+	_ "github.com/jinzhu/gorm/dialects/postgres" // Required to import database driver.
+	"github.com/spf13/cobra"
+)
+
+var schedulerRunCmd = &cobra.Command{
+	Use:   "run",
+	Short: "This command will start the flyte native scheduler and periodically get new schedules from the db for scheduling",
+	Run: func(cmd *cobra.Command, args []string) {
```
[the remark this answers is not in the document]
pmahindrakar-oss: Done

cmd/scheduler/entrypoints/scheduler.go
```diff
@@ -0,0 +1,86 @@
...
+			BaseConfig: repositoryCommonConfig.BaseConfig{
+				IsDebug: dbConfigValues.Debug,
+			},
+			Host:         dbConfigValues.Host,
+			Port:         dbConfigValues.Port,
+			DbName:       dbConfigValues.DbName,
+			User:         dbConfigValues.User,
+			Password:     dbConfigValues.Password,
+			ExtraOptions: dbConfigValues.ExtraOptions,
+		}
```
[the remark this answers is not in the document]
pmahindrakar-oss: Done

scheduler/schedule_executor.go
```diff
@@ -0,0 +1,111 @@
...
+	adminServiceClient     service.AdminServiceClient
+	workflowExecutorConfig *runtimeInterfaces.FlyteWorkflowExecutorConfig
+}
+
+func (w *ScheduledExecutor) Run(ctx context.Context) error {
+	logger.Infof(ctx, "Flyte native scheduler started successfully")
+
+	defer logger.Infof(ctx, "Flyte native scheduler shutdown")
+
+	// Read snapshot from the DB.
```
kumare3, line 41: lets add a couple lines here, about Snapshot reading, version of snapshot read, last time of the snapshot - etc?

scheduler/schedule_executor.go
```diff
@@ -0,0 +1,111 @@
...
+		return err
+	}
+
+	// Read all the schedules from the DB
+	schedules, err := w.db.SchedulableEntityRepo().GetAll(ctx)
+	if err != nil {
+		logger.Errorf(ctx, "unable to read the schedules from the db due to %v. Aborting", err)
+		return err
+	}
+	adminRateLimit := w.workflowExecutorConfig.GetAdminRateLimit()
```
kumare3, line 56: lets write how many schedules did we retrieve

scheduler/schedule_executor.go
```diff
@@ -0,0 +1,111 @@
...
+	executor := executor.New(w.scope, w.adminServiceClient)
+
+	// Create the scheduler using GoCronScheduler implementation
+	gcronScheduler := core.NewGoCronScheduler(w.scope, snapshot, rateLimiter, executor)
+	w.scheduler = gcronScheduler
+
+	// Bootstrap the schedules from the snapshot
+	bootStrapCtx, bootStrapCancel := context.WithCancel(ctx)
+	defer bootStrapCancel()
+	gcronScheduler.BootStrapSchedulesFromSnapShot(bootStrapCtx, schedules, snapshot)
```
kumare3, line 71: Instead of bootstrap can we make this part of the constructor?

kumare3: as in part of the NewGoCron... method

scheduler/schedule_executor.go
```diff
@@ -0,0 +1,111 @@
...
+	af := futures.NewAsyncFuture(ctx, func(ctx context.Context) (interface{}, error) {
+		return gcronScheduler.CatchupAll(ctx, currTime), nil
+	})
+	isCatchupSuccess, err := af.Get(ctx)
+	if err != nil {
+		logger.Errorf(ctx, "failed to get future value for catchup due to %v", err)
+		return err
+	}
+
+	if isCatchupSuccess.(bool) {
```
kumare3, line 90: if !success,
log and return nil?

and then put the logic out of this block (nit)

kumare3 (commented):
Couple small nits then +1. This is awesome. Thank you!

kumare3:
Can we also create an issue as a follow up that all the `schedule` related examples can be moved into the basic examples. Like these ones - https://docs.flyte.org/projects/cookbook/en/latest/auto/deployment/lp_schedules.html#sphx-glr-auto-deployment-lp-schedules-py

Move to
https://docs.flyte.org/projects/cookbook/en/latest/auto/remote_access/index.html

scheduler/schedule_executor.go
pmahindrakar-oss: Done

scheduler/schedule_executor.go
pmahindrakar-oss: Done

scheduler/schedule_executor.go
pmahindrakar-oss: Done

scheduler/schedule_executor.go
pmahindrakar-oss: Done

pmahindrakar-oss:
@kumare3 created this followup issue https://github.com/flyteorg/flyte/issues/1415.

scheduler/schedule_executor.go
```diff
@@ -0,0 +1,114 @@
...
+		return gcronScheduler.CatchupAll(ctx, currTime), nil
+	})
+	isCatchupSuccess, err := af.Get(ctx)
+	if err != nil {
+		logger.Errorf(ctx, "failed to get future value for catchup due to %v", err)
+		return err
+	}
+
+	if !isCatchupSuccess.(bool) {
+		logger.Errorf(ctx, "failed to catch up on all the schedules. Aborting")
```
kumare3, line 91: you need to return here

pmahindrakar-oss: Oops. Thanks for catching this.

kumare3 (approved):

kumare3:
Hurrah!!

PR #232, opened 2021-08-17
Fix Subworkflow auth role

milton0825:
If we are launching SubWorkflows defined in a different project and domain via dynamic task,
We want to make sure SubWorkflows are using the auth role defined in their LaunchPlan
instead of the auth role defined in the request (parent's auth role).

### DIFF
--- a/pkg/manager/impl/execution_manager.go
+++ b/pkg/manager/impl/execution_manager.go
@@ -610,7 +610,7 @@ func (m *ExecutionManager) launchSingleTaskExecution(
 
 }
 
-func resolvePermissions(ctx context.Context, request *admin.ExecutionCreateRequest, launchPlan *admin.LaunchPlan) *admin.AuthRole {
+func resolvePermissions(ctx context.Context, launchPlan *admin.LaunchPlan) *admin.AuthRole {
 	authRole := admin.AuthRole{}
 
 	// If we are launching SubWorkflows defined in a different project and domain via dynamic task,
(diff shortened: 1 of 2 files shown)


akhurana001:
👍  Let's test on staging . Also tagging @katrogan to see if this is broken in master or a change in expected behavior

pkg/manager/impl/execution_manager.go
```diff
@@ -624,9 +626,7 @@ func resolvePermissions(ctx context.Context, request *admin.ExecutionCreateReque
 		logger.Warning(ctx, "AssumableIamRole not set.")
 	}
 
-	if request.Spec.AuthRole != nil && len(request.Spec.AuthRole.KubernetesServiceAccount) > 0 {
```
katrogan, line 627: this is overridable in the request so we prefer to read the value there if it's set

milton0825:
This is how the workflow looks like:
```
from flytekit.common.launch_plan import SdkLaunchPlan
from flytekit.sdk.tasks import dynamic_task
from flytekit.sdk.workflow import workflow_class


@dynamic_task
def launch_simple_workflow(wf_params):
    launch_plan = SdkLaunchPlan.fetch(
        project='flytefunctionaltests',
        domain='production',
        name='tests.common.workflows.simple.SimpleWorkflow',
    )

    launch_plan.assign_name('ctsai')

    execution = launch_plan(input_1=1)
    yield execution


@workflow_class
class DynamicTaskWorkflow(object):
    task = launch_simple_workflow()
```

pkg/manager/impl/execution_manager.go
milton0825: This is the workflow that I am testing with:
```
from flytekit.common.launch_plan import SdkLaunchPlan
from flytekit.sdk.tasks import dynamic_task
from flytekit.sdk.workflow import workflow_class


@dynamic_task
def launch_simple_workflow(wf_params):
    launch_plan = SdkLaunchPlan.fetch(
        project='flytefunctionaltests',
        domain='production',
        name='tests.common.workflows.simple.SimpleWorkflow',
    )

    launch_plan.assign_name('ctsai')

    execution = launch_plan(input_1=1)
    yield execution


@workflow_class
class DynamicTaskWorkflow(object):
    task = launch_simple_workflow()
```

milton0825: It seems like the parent task sets the auth role in the request so it overrides the auth role defined in the launch plan

milton0825: nvm

PR #245, opened 2021-08-31
Change all to svc

milton0825:
At Lyft, we require all applications to request for `scope: svc` so changing the required scope from all to svc to satisfy lyft's setup.

### DIFF
--- a/cmd/entrypoints/serve.go
+++ b/cmd/entrypoints/serve.go
@@ -87,7 +87,7 @@ func blanketAuthorization(ctx context.Context, req interface{}, _ *grpc.UnarySer
 		return handler(ctx, req)
 	}
 
-	if !identityContext.Scopes().Has(auth.ScopeAll) {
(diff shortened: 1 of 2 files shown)


cmd/entrypoints/serve.go
```diff
@@ -87,7 +87,7 @@ func blanketAuthorization(ctx context.Context, req interface{}, _ *grpc.UnarySer
 		return handler(ctx, req)
 	}
 
-	if !identityContext.Scopes().Has(auth.ScopeAll) {
+	if !identityContext.Scopes().Has(auth.ScopeAll) || !identityContext.Scopes().Has("all") {
```
akhurana001, line 90: can we keep the const as such and add another one ? Probably easier to upstream with that

akhurana001: this will actually not work .. it should be && instead of || ?

PR #250, opened 2021-09-03
Handling empty inputUri for start-node and any node which doesn't have inputs

pmahindrakar-oss:

## _Read then delete this section_

# TL;DR
Handling the empty InputUri for any node event while doing get
We will fetch it from the storage only if its non-empty . 

```
[
	{
		"node_exec": {
			"id": {
				"nodeId": "start-node",
				"executionId": {
					"project": "flytesnacks",
					"domain": "development",
					"name": "fdbbaa49c298c4bb48ad"
				}
			},
			"closure": {
				"outputUri": "s3://my-s3-bucket/metadata/propeller/sandbox/flytesnacks-development-fdbbaa49c298c4bb48ad/start-node/data/0/outputs.pb",
				"phase": "SUCCEEDED",
				"createdAt": "2021-09-14T11:51:04.355307Z",
				"updatedAt": "2021-09-14T11:51:04.355307Z"
			},
			"metadata": {
				"specNodeId": "start-node"
			}
		},
		"outputs": {
			"am": true,
			"day_of_week": "Sunday",
			"number": 5
		}
	}
]
```

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1221
## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/node_execution_manager.go
+++ b/pkg/manager/impl/node_execution_manager.go
@@ -437,9 +437,12 @@ func (m *NodeExecutionManager) GetNodeExecutionData(
 		logger.Debugf(ctx, "failed to transform node execution model [%+v] when fetching data: %v", request.Id, err)
 		return nil, err
 	}
-	signedInputsURLBlob, err := m.urlData.Get(ctx, nodeExecution.InputUri)
-	if err != nil {
-		return nil, err
+	signedInputsURLBlob := admin.UrlBlob{}
+	if nodeExecution.InputUri != "" {
+		signedInputsURLBlob, err = m.urlData.Get(ctx, nodeExecution.InputUri)
+		if err != nil {
+			return nil, err
+		}
 	}
 	signedOutputsURLBlob := admin.UrlBlob{}
 	if nodeExecution.Closure.GetOutputUri() != "" {
--- a/pkg/manager/impl/util/data.go
+++ b/pkg/manager/impl/util/data.go
@@ -8,7 +8,7 @@ import (
 
 func ShouldFetchData(config *interfaces.RemoteDataConfig, urlBlob admin.UrlBlob) bool {
 	return config.Scheme == common.Local || config.Scheme == common.None || config.MaxSizeInBytes == 0 ||
-		urlBlob.Bytes < config.MaxSizeInBytes
+		(len(urlBlob.Url) > 0 && urlBlob.Bytes < config.MaxSizeInBytes)
 }
 
 func ShouldFetchOutputData(config *interfaces.RemoteDataConfig, urlBlob admin.UrlBlob, outputURI string) bool {
--- a/pkg/manager/impl/util/data_test.go
+++ b/pkg/manager/impl/util/data_test.go
@@ -15,6 +15,7 @@ func TestShouldFetchData(t *testing.T) {
 			Scheme:         common.Local,
 			MaxSizeInBytes: 100,
 		}, admin.UrlBlob{
+			Url: "s3://data",
 			Bytes: 200,
 		}))
 	})
@@ -23,6 +24,7 @@ func TestShouldFetchData(t *testing.T) {
 			Scheme:         common.None,
 			MaxSizeInBytes: 100,
 		}, admin.UrlBlob{
+			Url: "s3://data",
 			Bytes: 200,
 		}))
 	})
@@ -30,6 +32,7 @@ func TestShouldFetchData(t *testing.T) {
 		assert.True(t, ShouldFetchData(&interfaces.RemoteDataConfig{
 			Scheme: common.None,
 		}, admin.UrlBlob{
+			Url: "s3://data",
 			Bytes: 200,
 		}))
 	})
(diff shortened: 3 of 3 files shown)


pmahindrakar-oss:
Not sure why the codecov reduced even though it says 100 % diff coverage

pingsutw (approved):

kumare3:
@pmahindrakar-oss  what should the URI be? Have you actually at the bucket around this prefix.

Also why make this fix at all, everything has been ok without this fix. Let's fix it correctly in propeller

pmahindrakar-oss:
@kumare3  Checked the minio data for start-node and it only has output info 

Screenshot 2021-09-06 at 10 11 12 AM

So propeller is sending incorrect link in the event. I agree fixing in the propeller should be right thing to do here.

kumare3:
@pmahindrakar-oss so the error
is only for inputs right- so
In here we should understand that startnode is special- the input and output of start node is the same. And thus we could
Simply replicate them or even assume it in admin

pmahindrakar-oss:
@kumare3 the problem then becomes that propeller is not sending the output result in the event.

Screenshot 2021-09-06 at 6 18 17 PM

The closure doesn't have the output  set so we would have to make the propeller change to make closure output result and input uri the same in case of the start node. Let me know if thats ok with you

pmahindrakar-oss:
@kumare3 fix in propeller https://github.com/flyteorg/flytepropeller/pull/317 . Please have a look

katrogan:
can you update the PR title?

pkg/repositories/transformers/node_execution.go
```diff
@@ -88,6 +87,14 @@ func CreateNodeExecutionModel(input ToNodeExecutionModelInput) (*models.NodeExec
 		InputURI: input.Request.Event.InputUri,
 	}
 
+	// Start node is special case where the Inputs and Outputs are the same and hence here we copy the Output uri
```
katrogan, line 90: thank you for adding this! just to be clear, do we ever send multiple events for a start node? if so, should we update `UpdateNodeExecutionModel` too?

pkg/repositories/transformers/node_execution_test.go
```diff
@@ -167,6 +166,61 @@ func TestCreateNodeExecutionModel(t *testing.T) {
...
+						RetryAttempt: 1,
+					},
+				},
+			},
+		},
+		ParentTaskExecutionID: 8,
+	})
+	assert.Nil(t, err)
+
+	var closure = &admin.NodeExecutionClosure{
```
katrogan, line 197: shouldn't the output result from the event also be captured in this closure?

pkg/repositories/transformers/node_execution.go
```diff
@@ -88,6 +87,14 @@ func CreateNodeExecutionModel(input ToNodeExecutionModelInput) (*models.NodeExec
...
+	// into input
+	if input.Request.Event.Id.NodeId == v1alpha1.StartNodeID {
+		// We dont need to check the result of type assertion here since for Startnode the putput result
+		// would always be OutputUri
+		nodeExecution.InputURI = input.Request.Event.OutputResult.(*event.NodeExecutionEvent_OutputUri).OutputUri
```
katrogan, line 95: what if the workflow has no inputs, and therefore the start node has no outputs? will the output uri be set? can we verify the output result exists before dereferencing it?

pmahindrakar-oss: I have removed this logic for now as we can have nodes with no Inputs.

pkg/repositories/transformers/node_execution_test.go
pmahindrakar-oss: Removed this test as falling back to not fetching the data on start node input if the InputUri is not set

pkg/repositories/transformers/node_execution.go
pmahindrakar-oss: Same here. Please review the updated logic which avoids fetching the data from storage if the InputUri is not set 
At propeller side we set the InputUri on start-node to be empty

pkg/manager/impl/node_execution_manager.go
```diff
@@ -437,9 +437,12 @@ func (m *NodeExecutionManager) GetNodeExecutionData(
 		logger.Debugf(ctx, "failed to transform node execution model [%+v] when fetching data: %v", request.Id, err)
 		return nil, err
 	}
-	signedInputsURLBlob, err := m.urlData.Get(ctx, nodeExecution.InputUri)
-	if err != nil {
-		return nil, err
+	signedInputsURLBlob := admin.UrlBlob{}
+	if nodeExecution.InputUri != "" {
```
katrogan, line 441: should we also update https://github.com/flyteorg/flyteadmin/blob/0387a06511b633b485dbc2d8f216569bd2c22239/pkg/manager/impl/util/data.go#L9 to check if the blob is not empty?

pkg/repositories/transformers/node_execution.go
```diff
@@ -3,20 +3,17 @@ package transformers
 import (
 	"context"
 
-	"github.com/flyteorg/flytestdlib/logger"
-
 	"github.com/flyteorg/flyteadmin/pkg/common"
+	"github.com/flyteorg/flyteadmin/pkg/errors"
```
katrogan, line 7: mind reverting these transformer files since they're functionally unchanged and it mucks with git blame?

pkg/manager/impl/node_execution_manager.go
pmahindrakar-oss: Done

pkg/repositories/transformers/node_execution.go
pmahindrakar-oss: Done

katrogan (approved):

pkg/manager/impl/node_execution_manager.go
```diff
@@ -437,9 +437,12 @@ func (m *NodeExecutionManager) GetNodeExecutionData(
...
+		signedInputsURLBlob, err = m.urlData.Get(ctx, nodeExecution.InputUri)
```
EngHabu, line 442: nit:
```suggestion
	signedInputsURLBlob := admin.UrlBlob{}
	if len(nodeExecution.InputUri) != 0 {
		signedInputsURLBlob, err = m.urlData.Get(ctx, nodeExecution.InputUri)
```

pmahindrakar-oss: Thanks . Will take this as an immediate followup in the scheduler pr.

PR #262, opened 2021-10-01
Added default admin config for MaxParallelism

pmahindrakar-oss:

# TL;DR

Workflow executions can be gated by the max number of parallel nodes that can be processed. Specific override support for this MaxParallelism value already exists but not a generic default in the FlyteAdmin config.

This PR adds support in the admin config and provides a default value of 25 
Following is the order in which property override is checked

- launchplan spec override
- matchable resource attribute override
- admin application config override (defaults to 25 or else uses the configured value)

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1532

## Follow-up issue
_NA_

### DIFF
--- a/pkg/runtime/application_config_provider.go
+++ b/pkg/runtime/application_config_provider.go
@@ -37,7 +37,9 @@ var flyteAdminConfig = config.MustRegisterSection(flyteAdmin, &interfaces.Applic
 	MetadataStoragePrefix: []string{"metadata", "admin"},
 	EventVersion:          1,
 	AsyncEventsBufferSize: 100,
+	MaxParallelism:        25,
 })
+
 var schedulerConfig = config.MustRegisterSection(scheduler, &interfaces.SchedulerConfig{
 	EventSchedulerConfig: interfaces.EventSchedulerConfig{
 		Scheme:               common.Local,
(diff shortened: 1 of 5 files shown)


EngHabu (approved):
LGTM. Maybe get one more from Katrina?

katrogan (commented):
can we set a sensible default here: https://github.com/flyteorg/flyteadmin/blob/master/pkg/runtime/application_config_provider.go#L39

@kumare3 what sounds good? 100?

pmahindrakar-oss:
@katrogan its  already set to 25 which was discussed in the slack chat

pkg/runtime/application_config_provider.go
```diff
@@ -37,7 +37,9 @@ var flyteAdminConfig = config.MustRegisterSection(flyteAdmin, &interfaces.Applic
 	MetadataStoragePrefix: []string{"metadata", "admin"},
 	EventVersion:          1,
 	AsyncEventsBufferSize: 100,
+	MaxParallelism:        25,
```
pmahindrakar-oss, line 40: Have kept this as default for now .
cc : @katrogan @kumare3

katrogan: oh thanks! i missed this earlier

katrogan (approved):

PR #264, opened 2021-10-04
Prefer to set task limits from requests when only the latter is set

katrogan:
# TL;DR
Users report that 

> we're noticing that setting resource requests, but not limits, for tasks is causing the requests to not be honored. it appears that the platform defaults are getting set for the limits, and copied into requests. is this intended?

The source of this is that an earlier change erroneously preferred to assign unset limits from the admin application config rather than set request values (when they were set).

Tested on flytesandbox

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1450

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/execution_manager.go
+++ b/pkg/manager/impl/execution_manager.go
@@ -435,9 +435,6 @@ func (m *ExecutionManager) setCompiledTaskDefaults(ctx context.Context, task *co
 	if resource != nil && resource.Attributes != nil && resource.Attributes.GetTaskResourceAttributes() != nil {
 		taskResourceSpec = resource.Attributes.GetTaskResourceAttributes().Defaults
 	}
-	logger.Warnf(ctx, "setting task resources with defaults [%+v], task container [%+v], and taskResourceSpec [%+v]",
-		m.config.TaskResourceConfiguration().GetDefaults().CPU	,
-		task.Template.GetContainer().Resources.Requests, taskResourceSpec)
 	task.Template.GetContainer().Resources.Requests = assignResourcesIfUnset(
 		ctx, task.Template.Id, m.config.TaskResourceConfiguration().GetDefaults(), task.Template.GetContainer().Resources.Requests,
 		taskResourceSpec)
@@ -450,9 +447,7 @@ func (m *ExecutionManager) setCompiledTaskDefaults(ctx context.Context, task *co
 	task.Template.GetContainer().Resources.Limits = assignResourcesIfUnset(
 		ctx, task.Template.Id, createTaskDefaultLimits(ctx, task, m.config.TaskResourceConfiguration().GetDefaults()), task.Template.GetContainer().Resources.Limits,
 		taskResourceSpec)
-	logger.Warnf(ctx, "computed task limits [%+v]", task.Template.GetContainer().Resources.Limits)
 	checkTaskRequestsLessThanLimits(ctx, task.Template.Id, task.Template.GetContainer().Resources)
-	logger.Warnf(ctx, "and adjusted requests downwards [%+v]", task.Template.Id, task.Template.GetContainer().Resources)
 }
 
 func fromAdminProtoTaskResourceSpec(ctx context.Context, spec *admin.TaskResourceSpec) runtimeInterfaces.TaskResourceSet {
--- a/pkg/manager/impl/execution_manager.go
+++ b/pkg/manager/impl/execution_manager.go
@@ -3,6 +3,7 @@ package impl
 import (
 	"context"
 	"fmt"
+	v1 "k8s.io/api/core/v1"
 	"strconv"
 	"time"
 
@@ -294,6 +295,7 @@ func assignResourcesIfUnset(ctx context.Context, identifier *core.Identifier,
 			ephemeralStorageindex = idx
 		}
 	}
+	logger.Warnf(ctx, "cpu index [%+v]", cpuIndex)
 	if cpuIndex > 0 && memoryIndex > 0 && ephemeralStorageindex > 0 {
 		// nothing to do
 		return resourceEntries
@@ -342,6 +344,7 @@ func assignResourcesIfUnset(ctx context.Context, identifier *core.Identifier,
 			resourceEntries = append(resourceEntries, ephemeralStorageResource)
 		}
 	}
+	logger.Warnf(ctx, "returning [%+v]", resourceEntries)
 	return resourceEntries
 }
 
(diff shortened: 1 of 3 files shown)


pkg/manager/impl/execution_manager.go
```diff
@@ -234,26 +234,36 @@ func createTaskDefaultLimits(ctx context.Context, task *core.CompiledTask,
 	}
 	taskResourceLimits := runtimeInterfaces.TaskResourceSet{}
 
-	// For resource values, we prefer to use the limits set in the application config over the set resource values.
-	if !configResourceLimits.CPU.IsZero() {
+	// For resource values, we prefer to set unset limits to the request values over the config values.
+	if cpuIndex >= 0 {
+		cpuLimit = resourceRequestEntries[cpuIndex].Value
```
kumare3, line 239: why do we parse this multiple times? Why not parse it once at load time and then use? this will cause failuers much later, incase of a bug in the config?

pkg/manager/impl/execution_manager.go
```diff
@@ -234,26 +234,36 @@ func createTaskDefaultLimits(ctx context.Context, task *core.CompiledTask,
...
+		cpu, err := resource.ParseQuantity(cpuLimit)
+		if err != nil {
+			logger.Errorf(ctx, "Failed to parse user cpu limit from task spec [%s] with err [%+v]", cpuLimit, err)
```
kumare3, line 242:
```suggestion
			logger.Errorf(ctx, "In Task [%s] Failed to parse specified CPU limits [%s]. Resources should be valid Kubernetes quantities. Reason [%s]", task.Name, cpuLimit, err.String())
```

maybe `task.Name` is incorrect, but please provide the actual name of the task. Errors are extremely important. Obtuse errors just increase confusion

pkg/manager/impl/execution_manager.go
```diff
@@ -234,26 +234,36 @@ func createTaskDefaultLimits(ctx context.Context, task *core.CompiledTask,
...
 		} else {
 			taskResourceLimits.CPU = cpu
 		}
 	}
-	if !configResourceLimits.Memory.IsZero() {
+	if memoryIndex >= 0 {
+		memoryLimit = resourceRequestEntries[memoryIndex].Value
+		memory, err := resource.ParseQuantity(memoryLimit)
+		if err != nil {
+			logger.Errorf(ctx, "Failed to parse user memory limit from task spec [%s] with err [%+v]", memoryLimit, err)
```
kumare3, line 260: same as above, better errors please

pkg/manager/impl/execution_manager.go
```diff
@@ -234,26 +234,36 @@ func createTaskDefaultLimits(ctx context.Context, task *core.CompiledTask,
...
+		}
+	} else if !configResourceLimits.Memory.IsZero() {
 		taskResourceLimits.Memory = configResourceLimits.Memory
 	} else {
-		if memoryIndex >= 0 {
-			memoryLimit = resourceRequestEntries[memoryIndex].Value
-		}
 		memory, err := resource.ParseQuantity(memoryLimit)
 		if err != nil {
 			logger.Errorf(ctx, "Failed to parse user memory limit from task spec [%s] with err [%+v]", memoryLimit, err)
```
kumare3, line 269: same as above

pkg/manager/impl/execution_manager.go
```diff
@@ -265,18 +275,19 @@ func createTaskDefaultLimits(ctx context.Context, task *core.CompiledTask,
...
 		taskResourceLimits.GPU = configResourceLimits.GPU
 	}
-	if !configResourceLimits.EphemeralStorage.IsZero() {
-		taskResourceLimits.EphemeralStorage = configResourceLimits.EphemeralStorage
-	} else if ephemeralStorageIndex >= 0 {
+
+	if ephemeralStorageIndex >= 0 {
 		ephemeralStorage, err := resource.ParseQuantity(resourceRequestEntries[ephemeralStorageIndex].Value)
 		if err != nil {
 			logger.Errorf(ctx, "Failed to parse user ephemeral storage limit from task spec [%s] with err [%+v]",
```
kumare3, line 282: same as above

pkg/manager/impl/execution_manager.go
```diff
@@ -234,26 +234,36 @@ func createTaskDefaultLimits(ctx context.Context, task *core.CompiledTask,
...
+		}
+	} else if !configResourceLimits.CPU.IsZero() {
 		taskResourceLimits.CPU = configResourceLimits.CPU
 	} else {
-		if cpuIndex >= 0 {
-			cpuLimit = resourceRequestEntries[cpuIndex].Value
-		}
 		cpu, err := resource.ParseQuantity(cpuLimit)
 		if err != nil {
 			logger.Errorf(ctx, "Failed to parse user cpu limit from task spec [%s] with err [%+v]", cpuLimit, err)
```
pmahindrakar-oss, line 251: Wont this value be `cpuLimit := "500m"` and if yes do we need to handle the parsing and do error handling . If its const then we can probably have a right quantity set . 

Also the error message might need to change if thats the case

pkg/manager/impl/execution_manager.go
```diff
@@ -234,26 +234,36 @@ func createTaskDefaultLimits(ctx context.Context, task *core.CompiledTask,
...
+		} else {
+			taskResourceLimits.Memory = memory
+		}
+	} else if !configResourceLimits.Memory.IsZero() {
 		taskResourceLimits.Memory = configResourceLimits.Memory
 	} else {
-		if memoryIndex >= 0 {
-			memoryLimit = resourceRequestEntries[memoryIndex].Value
-		}
 		memory, err := resource.ParseQuantity(memoryLimit)
```
pmahindrakar-oss, line 267: Same here .
`memoryLimit := "500Mi"`

pkg/manager/impl/execution_manager.go
katrogan: talked with Haytham - since we never expect to use these defaults (represents a configuration error if so) we'll log an error when the task limits end up resolving to unset.

pkg/manager/impl/execution_manager.go
katrogan: see above

pkg/manager/impl/execution_manager.go
katrogan: updated

pkg/manager/impl/execution_manager.go
katrogan: code has been refactored but the full task id is in the error msg

pkg/manager/impl/execution_manager.go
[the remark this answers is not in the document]
katrogan: yes, the index is from `resourceRequestEntries`

PR #283, opened 2021-11-03
Make returning execution data via signed URL optional

katrogan:

# TL;DR
Signed URLs in get execution data calls are deprecated, like [here](https://github.com/flyteorg/flyteidl/blob/6eff9b0be4f41b717b2d89697c7187a5216a0547/protos/flyteidl/admin/execution.proto#L307,L313). This PR introduces a config option to disable computing them for get execution data responses and defaults to omitting them altogether.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1796

## Follow-up issue
_NA_

### DIFF
--- a/pkg/runtime/interfaces/application_configuration.go
+++ b/pkg/runtime/interfaces/application_configuration.go
@@ -295,6 +295,9 @@ func (s *SchedulerConfig) GetReconnectDelaySeconds() int {
 
 // Configuration specific to setting up signed urls.
 type SignedURL struct {
+	// Whether signed urls should even be returned with GetExecutionData, GetNodeExecutionData and GetTaskExecutionData
+	// response objects.
+	Enabled bool `json:"enabled"`
 	// The amount of time for which a signed URL is valid.
 	DurationMinutes int `json:"durationMinutes"`
 	// The principal that signs the URL. This is only applicable to GCS URL.
--- a/pkg/runtime/interfaces/application_configuration.go
+++ b/pkg/runtime/interfaces/application_configuration.go
@@ -297,7 +297,7 @@ func (s *SchedulerConfig) GetReconnectDelaySeconds() int {
 type SignedURL struct {
 	// Whether signed urls should even be returned with GetExecutionData, GetNodeExecutionData and GetTaskExecutionData
 	// response objects.
-	Enabled bool `json:"enabled"`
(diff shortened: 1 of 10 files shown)


pmahindrakar-oss (dismissed):

script/integration/k8s/integration.yaml
```diff
@@ -296,6 +296,7 @@ data:
       region: "my-region"
       scheme: local
       signedUrls:
+        enabled: true
```
pmahindrakar-oss, line 299: If the default behavior is going to be false as per this https://github.com/flyteorg/flyteadmin/pull/283/files#diff-4149c9b64083f7b9e6680f7294d71f6ba58bc8da28a1189309eeea69d8ed07f3R63

Should we instead have integration test for false since that will be run without any config changes.
If possible we can have both the vairations in integration tests

katrogan: since changing configs in integration tests is a bit complicated I'll just remove this and adjust the tests accordingly to use the default behavior.

katrogan:
PTAL @pmahindrakar-oss

EngHabu (dismissed):
One more +1... because why not

pmahindrakar-oss (approved):

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -295,6 +295,9 @@ func (s *SchedulerConfig) GetReconnectDelaySeconds() int {
 
 // Configuration specific to setting up signed urls.
 type SignedURL struct {
+	// Whether signed urls should even be returned with GetExecutionData, GetNodeExecutionData and GetTaskExecutionData
+	// response objects.
+	Enabled bool `json:"enabled"`
```
EngHabu, line 300:
```suggestion
	Enabled bool `json:"enabled" pflag:",Whether signed urls should even be returned with GetExecutionData, GetNodeExecutionData and GetTaskExecutionData response objects."`
```

EngHabu (approved):

EngHabu (approved):

PR #284, opened 2021-11-08
Registerable k8s workflow executor

katrogan:
# TL;DR
Provide additional abstraction layer for execution creation to enhance runtime flexibility.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Follow-up issue
_NA_


pkg/workflowengine/impl/propeller_executor.go
```diff
@@ -3,10 +3,12 @@ package impl
 import (
```
EngHabu, line 3: Something is a bit confusing about the structure of the code here... 
why is this called propeller_executor?

It feels as if we have sort of two abstractions, plz correct me if this is wrong,
1. workflow engine abstracts the execution of workflows... with "propeller" as "one" possible executor.. but no way to select another
2. Within propeller, it abstracts the execution of workflows on k8s and we have one default implementation that creates the WF CRD in the target cluster directly and we are introducing a way to register separate k8s executors... right?

If that's the case, I would like to understand why? What does the propeller abstraction add that can't be moved up one level to the workflow engine abstraction (metrics?) ? and why are we only opening a way to registering separate k8s executors but not replace propeller executor?

pkg/manager/impl/execution_manager.go
```diff
@@ -55,19 +55,22 @@ const childContainerQueueKey = "child_queue"
...
+	PublishNotificationError  prometheus.Counter
+	TransformerError          prometheus.Counter
+	UnexpectedDataError       prometheus.Counter
+	SpecSizeBytes             prometheus.Summary
+	ClosureSizeBytes          prometheus.Summary
+	AcceptanceDelay           prometheus.Summary
+	PublishEventError         prometheus.Counter
+	WorkflowBuildSuccess      prometheus.Counter
+	WorkflowBuildFailure      prometheus.Counter
+	TerminateExecutionFailure prometheus.Counter
```
pmahindrakar-oss, line 73: Since these are counters, should we use plurals similar to PropellerFailures

pkg/manager/impl/execution_manager.go
```diff
@@ -140,40 +144,23 @@ func validateMapSize(maxEntries int, candidate map[string]string, candidateName
...
-	partiallyPopulatedInputs *workflowengineInterfaces.ExecuteWorkflowInput) error {
-
-	var labels map[string]string
-	if requestSpec.Labels != nil && requestSpec.Labels.Values != nil {
-		labels = requestSpec.Labels.Values
-	} else if partiallyPopulatedInputs.Reference.Spec.Labels != nil &&
-		partiallyPopulatedInputs.Reference.Spec.Labels.Values != nil {
-		labels = partiallyPopulatedInputs.Reference.Spec.Labels.Values
-	}
+type mapLikeThing interface {
```
pmahindrakar-oss, line 147: nit : mapWrapper may be

pkg/manager/impl/execution_manager.go
```diff
@@ -1497,11 +1528,19 @@ func newExecutionSystemMetrics(scope promutils.Scope) executionSystemMetrics {
 			"delay in seconds from when an execution was requested to be created and when it actually was"),
 		PublishEventError: scope.MustNewCounter("publish_event_error",
 			"overall count of publish event errors when invoking publish()"),
+
+		WorkflowBuildSuccess: scope.MustNewCounter("build_success",
```
pmahindrakar-oss, line 1532: workflow_build_success , as build_success could be too generic a name to understand on the dashboard.
Similar for workflow_build_failure

pkg/manager/impl/execution_manager.go
```diff
@@ -140,40 +144,23 @@ func validateMapSize(maxEntries int, candidate map[string]string, candidateName
...
+	GetValues() map[string]string
+}
 
-	var annotations map[string]string
-	if requestSpec.Annotations != nil && requestSpec.Annotations.Values != nil {
-		annotations = requestSpec.Annotations.Values
-	} else if partiallyPopulatedInputs.Reference.Spec.Annotations != nil &&
-		partiallyPopulatedInputs.Reference.Spec.Annotations.Values != nil {
-		annotations = partiallyPopulatedInputs.Reference.Spec.Annotations.Values
+func (m *ExecutionManager) resolveStringMap(preferredValues, defaultValues mapLikeThing, valueName string, maxEntries int) (map[string]string, error) {
```
pmahindrakar-oss, line 151: Should this  now be member function of ExecutionManager.

Earlier since it was using m.config.RegistrationValidationConfiguration().GetMaxLabelEntries() , but this seems to be standalone function now.

pkg/workflowengine/impl/workflow_builder.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+type flyteWorkflowBuilder struct {
+	metrics builderMetrics
+}
+
+func (b *flyteWorkflowBuilder) Build(
+	wfClosure *core.CompiledWorkflowClosure, inputs *core.LiteralMap, executionID *core.WorkflowExecutionIdentifier,
+	namespace string) (*v1alpha1.FlyteWorkflow, error) {
+	return k8s.BuildFlyteWorkflow(wfClosure, inputs, executionID, namespace)
```
pmahindrakar-oss, line 28: Should this be called out as an interface BuildFlyteWorkflow and then be able to subsitute mock or other implmentations.
The interface can be part flyteWorkflowBuilder

pkg/workflowengine/impl/workflow_builder.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func newBuilderMetrics(scope promutils.Scope) builderMetrics {
+	return builderMetrics{
+		Scope: scope,
+		WorkflowBuildSuccess: scope.MustNewCounter("build_success",
+			"count of workflows built by propeller without error"),
+		WorkflowBuildFailure: scope.MustNewCounter("build_failure",
+			"count of workflows built by propeller with errors"),
```
pmahindrakar-oss, line 37: Same here . may be we can add the workflow_ prefix to the counter and make them plural.

pkg/workflowengine/impl/workflow_builder.go
pmahindrakar-oss: I think we already have an interface, can we use that here .

pkg/workflowengine/interfaces/registry.go
```diff
@@ -0,0 +1,12 @@
+package interfaces
+
+// FlyteK8sWorkflowExecutorRegistry is a singleton provider of a K8sWorkflowExecutor implementation to use for
+// creating and deleting Flyte workflow CRD objects.
+type FlyteK8sWorkflowExecutorRegistry interface {
+	// Registers a new K8sWorkflowExecutor to handle creating and aborting Flyte workflow executions.
+	Register(executor K8sWorkflowExecutor)
+	// Registers the default K8sWorkflowExecutor to handle creating and aborting Flyte workflow executions.
+	RegisterDefault(executor K8sWorkflowExecutor)
```
pmahindrakar-oss, line 9: Will there be at a time default and non-default workflow executors in the system. Because i dont see reference to Register()

katrogan: mind checking the offline thread for an explanation? :)

pkg/workflowengine/impl/workflow_builder.go
katrogan: done, thanks

pkg/manager/impl/execution_manager.go
katrogan: good point, done

pkg/manager/impl/execution_manager.go
katrogan: done

pkg/manager/impl/execution_manager.go
katrogan: updated

pkg/manager/impl/execution_manager.go
katrogan: okay, done

pkg/workflowengine/impl/workflow_builder.go
katrogan: Sorry, I'm not following - this is already an implementation of the Build interface

katrogan:
updated, PTAL @EngHabu

pkg/workflowengine/interfaces/registry.go
pmahindrakar-oss: Thanks for clarifying this offline

katrogan:
PTAL @EngHabu @pmahindrakar-oss

EngHabu (approved):
Awesome! It's a large change... Let's do some manual verification before merging

pkg/manager/impl/execution_manager.go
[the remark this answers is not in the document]
katrogan: mind refreshing? the latest should already have the ExecutionManager receiver removed

this handles reconciling Labels and Annotations from the CreateExecutionRequest spec and the launch plan defaults

these are both maps but the messages themselves define the map values in a field called `values`

pkg/workflowengine/impl/workflow_builder.go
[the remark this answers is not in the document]
katrogan: if you refresh, the counter should be called :)

katrogan:
verified with a live deployment that
- inputs get passed
- annotations & labels get passed correctly
- execution config params are passed
  - Recovery Execution
  - Max Parallelism
- permissions (k8s service account and iam role) get resolved
- task resources get set
- AcceptedAt set
- RawOutputDataConfig set
@EngHabu

PR #295, opened 2021-11-19
Add ability to create any k8s resource from Flyte's resource controller

pradithya:
# TL;DR
Allow Flyte's resource controller to provision any k8s resource

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Flyte is able to create kubernetes resource for each of Flyte's project via Flyte's resource controller. Flyte admin can customise what resource to be created by specifying k8s template file(s) such as the [example](https://github.com/flyteorg/flyteadmin/tree/master/sampleresourcetemplates). 

Currently, this feature is limited to provisioning k8s resource recognised by the controller's  decoder (and currently it is able to recognize standard k8s resources). It's the case because current implementation requires decoding the kubernetes manifest (yaml) into kubernetes object, thus the schema must be registered first. 
The PR remove the necessity of decoding the k8s manifest hence allowing any resource to be created (as long as it's recognized by the k8s api) 
In my case, this feature is useful for automating Flyte's project creation when it need to create a separate [Google Service Account](https://cloud.google.com/config-connector/docs/reference/resource-docs/iam/iamserviceaccount) via [config connector](https://cloud.google.com/config-connector/docs/overview) for each Flyte's project to be used by task execution pod when using any Google's offering (BQ, GCS, etc)

```
apiVersion: iam.cnrm.cloud.google.com/v1beta1
kind: IAMServiceAccount
metadata:
  name: {{ namespace }}-gsa
  namespace: {{ namespace }}
spec:
  displayName: Service account for {{ namespace }}
```

### DIFF
--- a/pkg/clusterresource/controller.go
+++ b/pkg/clusterresource/controller.go
@@ -3,10 +3,6 @@ package clusterresource
 import (
 	"context"
 	"encoding/json"
-
-	"github.com/flyteorg/flyteadmin/pkg/executioncluster"
-	"k8s.io/apimachinery/pkg/types"
-
 	//"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -17,7 +13,12 @@ import (
 	"strings"
 	"time"
 
+	"github.com/flyteorg/flyteadmin/pkg/executioncluster"
+	"github.com/flyteorg/flyteadmin/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+
 	"k8s.io/apimachinery/pkg/api/meta"
+	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/discovery/cached/memory"
@@ -40,6 +41,8 @@ import (
 	"github.com/flyteorg/flyteadmin/pkg/common"
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
 
+	ccscheme "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/client/clientset/versioned/scheme"
+
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -52,7 +55,6 @@ import (
 	"github.com/flyteorg/flyteadmin/pkg/errors"
 	"github.com/flyteorg/flyteadmin/pkg/repositories"
 	repositoriesInterfaces "github.com/flyteorg/flyteadmin/pkg/repositories/interfaces"
-	"github.com/flyteorg/flyteadmin/pkg/runtime"
 	runtimeInterfaces "github.com/flyteorg/flyteadmin/pkg/runtime/interfaces"
 	"github.com/flyteorg/flytestdlib/promutils"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -102,6 +104,8 @@ type controller struct {
 	lastAppliedTemplateDir string
 	// Map of [namespace -> [templateFileName -> last modified time]]
 	appliedTemplates NamespaceCache
+	// K8S manifest template decoder
+	decoder k8sruntime.Decoder
 }
 
 var descCreatedAtSortParam, _ = common.NewSortParameter(admin.Sort{
@@ -284,10 +288,11 @@ func prepareDynamicCreate(target executioncluster.ExecutionTarget, config string
 
 // This function loops through the kubernetes resource template files in the configured template directory.
 // For each unapplied template file (wrt the namespace) this func attempts to
-//   1) read the template file
-//   2) substitute templatized variables with their resolved values
-//   3) decode the output of the above into a kubernetes resource
-//   4) create the resource on the kubernetes cluster and cache successful outcomes
+//   1) create k8s object resource from template by performing:
+//      a) read template file
+//      b) substitute templatized variables with their resolved values
+//      c) decode the output of the above into a kubernetes resource
+//   2) create the resource on the kubernetes cluster and cache successful outcomes
 func (c *controller) syncNamespace(ctx context.Context, project models.Project, domain runtimeInterfaces.Domain, namespace NamespaceName,
 	templateValues, customTemplateValues templateValuesType) error {
 	templateDir := c.config.ClusterResourceConfiguration().GetTemplatePath()
@@ -319,52 +324,14 @@ func (c *controller) syncNamespace(ctx context.Context, project models.Project, 
 			continue
 		}
 
-		// 1) read the template file
-		template, err := ioutil.ReadFile(path.Join(templateDir, templateFileName))
+		// 1) create resource from template:
+		k8sObj, k8sManifest, err := c.createResourceFromTemplate(ctx, templateDir, templateFileName, project, domain, namespace, templateValues, customTemplateValues)
 		if err != nil {
-			logger.Warningf(ctx,
-				"failed to read config template from path [%s] for namespace [%s] with err: %v",
-				templateFileName, namespace, err)
-			err := errors.NewFlyteAdminErrorf(
-				codes.Internal, "failed to read config template from path [%s] for namespace [%s] with err: %v",
-				templateFileName, namespace, err)
 			collectedErrs = append(collectedErrs, err)
-			c.metrics.TemplateReadErrors.Inc()
-			continue
-		}
-		logger.Debugf(ctx, "successfully read template config file [%s]", templateFileName)
-
-		// 2) substitute templatized variables with their resolved values
-		// First, add the special case namespace template which is always substituted by the system
-		// rather than fetched via a user-specified source.
-		templateValues[fmt.Sprintf(templateVariableFormat, namespaceVariable)] = namespace
-		templateValues[fmt.Sprintf(templateVariableFormat, projectVariable)] = project.Identifier
-		templateValues[fmt.Sprintf(templateVariableFormat, domainVariable)] = domain.ID
-
-		var config = string(template)
-		for templateKey, templateValue := range templateValues {
-			config = strings.Replace(config, templateKey, templateValue, replaceAllInstancesOfString)
-		}
-		// Replace remaining template variables from domain specific defaults.
-		for templateKey, templateValue := range customTemplateValues {
-			config = strings.Replace(config, templateKey, templateValue, replaceAllInstancesOfString)
-		}
-
-		// 3) decode the kubernetes resource template file into an actual resource object
-		decode := scheme.Codecs.UniversalDeserializer().Decode
-		k8sObj, _, err := decode([]byte(config), nil, nil)
-		if err != nil {
-			logger.Warningf(ctx, "Failed to decode config template [%s] for namespace [%s] into a kubernetes object with err: %v",
-				templateFileName, namespace, err)
-			err := errors.NewFlyteAdminErrorf(codes.InvalidArgument,
-				"Failed to decode namespace config template [%s] for namespace [%s] into a kubernetes object with err: %v",
-				templateFileName, namespace, err)
-			collectedErrs = append(collectedErrs, err)
-			c.metrics.TemplateDecodeErrors.Inc()
 			continue
 		}
 
-		// 4) create the resource on the kubernetes cluster and cache successful outcomes
+		// 2) create the resource on the kubernetes cluster and cache successful outcomes
 		if _, ok := c.appliedTemplates[namespace]; !ok {
 			c.appliedTemplates[namespace] = make(LastModTimeCache)
 		}
@@ -373,7 +340,7 @@ func (c *controller) syncNamespace(ctx context.Context, project models.Project, 
 			logger.Debugf(ctx, "Attempting to create resource [%+v] in cluster [%v] for namespace [%s]",
 				k8sObj.GetObjectKind().GroupVersionKind().Kind, target.ID, namespace)
 
-			dynamicObj, err := prepareDynamicCreate(target, config)
+			dynamicObj, err := prepareDynamicCreate(target, k8sManifest)
 			if err != nil {
 				logger.Warningf(ctx, "Failed to transform kubernetes template file for [%+v] for namespace [%s] "+
 					"into a dynamic unstructured mapping with err: %v", k8sObj.GetObjectKind().GroupVersionKind().Kind, namespace, err)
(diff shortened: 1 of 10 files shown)


kumare3:
is there a way to just create the raw json, without decoding the object at all? IIUC, the problem is that the scheme is not standard and hence unknown.
Is it possible to use something like [k8s unstructured client](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured) ?

pradithya:
> is there a way to just create the raw json, without decoding the object at all? IIUC, the problem is that the scheme is not standard and hence unknown. Is it possible to use something like [k8s unstructured client](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured) ?

I think it is possible to make it similar to `kubectl`. Let me get better understanding why current approach requires decoding.

pradithya:
While smoke testing, I managed to create `IAMServiceAccount` resource

```
apiVersion: iam.cnrm.cloud.google.com/v1beta1
kind: IAMServiceAccount
metadata:
  annotations:
    cnrm.cloud.google.com/management-conflict-prevention-policy: none
    cnrm.cloud.google.com/project-id: flyteexamples
    cnrm.cloud.google.com/state-into-spec: merge
  creationTimestamp: "2021-11-19T06:06:16Z"
  finalizers:
  - cnrm.cloud.google.com/finalizer
  - cnrm.cloud.google.com/deletion-defender
  generation: 1
  name: flyteexamples-gsa
  namespace: flyteexamples
  resourceVersion: "51426499"
  selfLink: /apis/iam.cnrm.cloud.google.com/v1beta1/namespaces/flyteexamples/iamserviceaccounts/flyteexamples-gsa
  uid: cf2d53a1-08d9-495f-b75d-4d323e36aaa5
```

But in subsequent sync it causes the resource controller to crashes with following errors. Any idea?

```
{"json":{"src":"controller.go:522"},"level":"warning","msg":"Failed to create cluster resources for namespace [flyteexamples] with err: iamserviceaccounts.iam.cnrm.cloud.google.com \"flyteexamples-gsa\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update","ts":"2021-11-19T06:44:12Z"}
{"json":{"src":"clusterresource.go:112"},"level":"fatal","msg":"Failed to sync cluster resources [iamserviceaccounts.iam.cnrm.cloud.google.com \"my-project-gsa\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update, iamserviceaccounts.iam.cnrm.cloud.google.com \"flytetester-gsa\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update, iamserviceaccounts.iam.cnrm.cloud.google.com \"flytesnacks-gsa\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update, iamserviceaccounts.iam.cnrm.cloud.google.com \"flyteexamples-gsa\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update]","ts":"2021-11-19T06:44:12Z"}
```

katrogan:
does updating a Google Service Account require additional parameters not in the create? are you able to test updating without the cluster resource controller just by applying yaml?

pradithya:
> does updating a Google Service Account require additional parameters not in the create? are you able to test updating without the cluster resource controller just by applying yaml?

Yes,  applying them directly using kubectl works fine. 
I tried to ensure the update happens using patch (by adding the resource kind into [this mapping](https://github.com/flyteorg/flyteadmin/blob/1373fa4e8eb83893dbd78e6e5de132e5991a524e/pkg/clusterresource/controller.go#L120), but it doesn't seem to work too, likely due to https://github.com/kubernetes/website/pull/10994

kumare3:
@pradithya we should never call and update. We should simply call create. Update needs the resource version of the current object. So perform an update you have to fetch, merge and then update. 

https://github.com/kubernetes/apimachinery/blob/7149480564b7d40046f19040efa20171b49059ae/pkg/apis/meta/v1/types.go#L78

kumare3:
@katrogan why do we update? Is
It for secrets

pradithya:
Updated the PR to allow any unknown resource to be created by cluster resource controller.

The issue with update still persists though, I'll need your advise on that @kumare3 @katrogan

katrogan:
@kumare3 we actually use update by default, k8s handles reconciling with the existing resource. we only use strategic merge patch for serviceaccounts because update was actually resulting in creating a new one each time, which quickly overwhelms k8s api server when the number of resources grows large

@pradithya sorry this is extra work, but it looks JSONPatch and JSON merge patch [work](https://github.com/kubernetes/kubernetes/issues/58414#issuecomment-383148632) perhaps you could implement that and we could similarly special-case handle JSONPatch update-types (e.g. CRDs). since i believe the current default behavior to call Update() actually ends up calling http [PUT](https://github.com/kubernetes/client-go/blob/master/dynamic/simple.go#L161)

pradithya:
Latest implementation uses Patch for all resources with following strategy:
1. Strategic merge patch for known resource
2. JSON merge patch for custom resource
The implementation heavily follows `kubectl apply` behaviour to ensure more use cases.
It will also skip the patch if the controller doesn't detect any modification.

Have been testing it in my dev cluster and so far so good, it managed to create new resource (no change in this code path) and sync existing resource for both known and custom resource.

```
{"json":{"src":"controller.go:343"},"level":"debug","msg":"Attempting to create resource [ServiceAccount] in cluster [] for namespace [flyteexamples]","ts":"2021-11-23T16:07:10Z"}
{"json":{"src":"controller.go:351"},"level":"debug","msg":"Type [ServiceAccount] in namespace [flyteexamples] already exists - attempting update instead","ts":"2021-11-23T16:07:10Z"}
{"json":{"src":"controller.go:383"},"level":"info","msg":"Resource [ServiceAccount] in namespace [flyteexamples] is not modified","ts":"2021-11-23T16:07:10Z"}
{"json":{"src":"controller.go:481"},"level":"debug","msg":"successfully read template config file [ab_project_resource_quota.yaml]","ts":"2021-11-23T16:07:10Z"}
{"json":{"src":"controller.go:343"},"level":"debug","msg":"Attempting to create resource [ResourceQuota] in cluster [] for namespace [flyteexamples]","ts":"2021-11-23T16:07:10Z"}
{"json":{"src":"controller.go:351"},"level":"debug","msg":"Type [ResourceQuota] in namespace [flyteexamples] already exists - attempting update instead","ts":"2021-11-23T16:07:11Z"}
{"json":{"src":"controller.go:398"},"level":"debug","msg":"Successfully updated resource [ResourceQuota] in namespace [flyteexamples]","ts":"2021-11-23T16:07:11Z"}
{"json":{"src":"controller.go:481"},"level":"debug","msg":"successfully read template config file [abb_google_service_account.yaml]","ts":"2021-11-23T16:07:11Z"}
{"json":{"src":"controller.go:343"},"level":"debug","msg":"Attempting to create resource [IAMServiceAccount] in cluster [] for namespace [flyteexamples]","ts":"2021-11-23T16:07:11Z"}
{"json":{"src":"controller.go:351"},"level":"debug","msg":"Type [IAMServiceAccount] in namespace [flyteexamples] already exists - attempting update instead","ts":"2021-11-23T16:07:11Z"}
{"json":{"src":"controller.go:383"},"level":"info","msg":"Resource [IAMServiceAccount] in namespace [flyteexamples] is not modified","ts":"2021-11-23T16:07:11Z"}


```

pradithya:
@katrogan @kumare3 It can be reviewed again

pkg/clusterresource/controller.go
```diff
@@ -447,6 +499,69 @@ func (c *controller) createResourceFromTemplate(ctx context.Context, templateDir
...
+// for custom resource, json merge patch is used
+// heavily inspired by kubectl's patcher
+func (c *controller) createPatch(mapping *meta.RESTMapping, currentObj *unstructured.Unstructured, modified []byte, namespace string) ([]byte, types.PatchType, error) {
+	current, err := k8sruntime.Encode(unstructured.UnstructuredJSONScheme, currentObj)
+	if err != nil {
+		return nil, "", fmt.Errorf("Failed to encode [%+v] in namespace [%s] to json with err: %v",
+			currentObj.GetKind(), namespace, err)
+	}
+
+	original, err := getLastApplied(currentObj)
```
katrogan, line 501: do you ever check that err != nil?

pkg/clusterresource/controller.go
```diff
@@ -447,6 +499,69 @@ func (c *controller) createResourceFromTemplate(ctx context.Context, templateDir
...
+
+	var patch []byte
+	patchType := types.StrategicMergePatchType
+	obj, err := scheme.Scheme.New(mapping.GroupVersionKind)
+	switch {
+	case err == nil:
+		patchType = types.StrategicMergePatchType
```
katrogan, line 512: nit: this is what you initialize it to already in line 516

pkg/clusterresource/controller.go
```diff
@@ -447,6 +499,69 @@ func (c *controller) createResourceFromTemplate(ctx context.Context, templateDir
...
+		if patch == nil {
```
katrogan, line 513: isn't patch always nil at this point?

pkg/clusterresource/controller.go
```diff
@@ -447,6 +499,69 @@ func (c *controller) createResourceFromTemplate(ctx context.Context, templateDir
...
+					currentObj.GetKind(), namespace, err, original, modified, current)
+			}
+		}
+	case k8sruntime.IsNotRegisteredError(err):
+		patchType = types.MergePatchType
+		preconditions := []mergepatch.PreconditionFunc{mergepatch.RequireKeyUnchanged("apiVersion"),
+			mergepatch.RequireKeyUnchanged("kind"), mergepatch.RequireMetadataKeyUnchanged("name")}
+		patch, err = jsonmergepatch.CreateThreeWayJSONMergePatch(original, modified, current, preconditions...)
+		if err != nil {
+			return nil, "", fmt.Errorf("Failed to create 3 way merge patch for resource [%+v] in namespace [%s] with err: %v",
```
katrogan: is this case still 3 way merge patch?

pkg/clusterresource/controller.go
```diff
@@ -447,6 +499,69 @@ func (c *controller) createResourceFromTemplate(ctx context.Context, templateDir
...
+					currentObj.GetKind(), namespace, err)
+			}
+
+			patch, err = strategicpatch.CreateThreeWayMergePatch(original, modified, current, lookupPatchMeta, true)
+			if err != nil {
+				return nil, "", fmt.Errorf("Failed to create 3 way merge patch for resource [%+v] in namespace [%s] with err: %v\noriginal:\n%s\nmodified:\n%s\ncurrent:\n%s",
+					currentObj.GetKind(), namespace, err, original, modified, current)
+			}
+		}
+	case k8sruntime.IsNotRegisteredError(err):
```
katrogan, line 526: what does not registered in this case mean?

pkg/clusterresource/controller.go
```diff
@@ -447,6 +499,69 @@ func (c *controller) createResourceFromTemplate(ctx context.Context, templateDir
...
+		patchType = types.MergePatchType
+		preconditions := []mergepatch.PreconditionFunc{mergepatch.RequireKeyUnchanged("apiVersion"),
+			mergepatch.RequireKeyUnchanged("kind"), mergepatch.RequireMetadataKeyUnchanged("name")}
+		patch, err = jsonmergepatch.CreateThreeWayJSONMergePatch(original, modified, current, preconditions...)
+		if err != nil {
+			return nil, "", fmt.Errorf("Failed to create 3 way merge patch for resource [%+v] in namespace [%s] with err: %v",
+				currentObj.GetKind(), namespace, err)
+		}
+	case err != nil:
```
katrogan, line 535: nit: just `default` here

pkg/clusterresource/controller.go
```diff
@@ -447,6 +499,69 @@ func (c *controller) createResourceFromTemplate(ctx context.Context, templateDir
...
+		return nil, "", fmt.Errorf("Failed to create get instance of versioned object [%+v] in namespace [%s] with err: %v",
+			currentObj.GetKind(), namespace, err)
+
+	}
+
+	if string(patch) == "{}" {
```
katrogan, line 541: nit, can we make "{}" a named constant?

pkg/clusterresource/controller.go
```diff
@@ -447,6 +499,69 @@ func (c *controller) createResourceFromTemplate(ctx context.Context, templateDir
...
+		// not modified
+		return patch, patchType, nil
+	}
+
+	if currentObj.GetResourceVersion() != "" {
```
katrogan, line 546: nit: `if len(currentObj.GetResourceVersion()) > 0 {`

pkg/clusterresource/controller.go
```diff
@@ -346,38 +352,47 @@ func (c *controller) syncNamespace(ctx context.Context, project models.Project,
...
+					patch, patchType, err := c.createPatch(dynamicObj.mapping, currentObj, modified, namespace)
+					if err != nil {
+						c.metrics.TemplateUpdateErrors.Inc()
+						logger.Warningf(ctx, "Failed to create patch for resource [%+v] in namespace [%s] err: %v",
+							dynamicObj.obj.GetKind(), namespace, err)
+						collectedErrs = append(collectedErrs, err)
+						continue
+					}
+
+					if string(patch) == "{}" {
```
katrogan, line 370: can we use a named constant for "{}"?

katrogan (commented):
this is awesome! thank you for cleaning up this code and adding such thorough unit tests

just a few nits and questions but looks great

pkg/clusterresource/controller.go
pradithya: Yep, missed this, updated now

pkg/clusterresource/controller.go
pradithya: it should be json merge patch, updated the error message

pkg/clusterresource/controller.go
pradithya: It's resources that's not recognized by client-go's schema by default. It's usually CRDs

pkg/clusterresource/controller.go
pradithya: Updated

pkg/clusterresource/controller.go
pradithya: Updated

pkg/clusterresource/controller.go
pradithya: Updated

pkg/clusterresource/controller.go
pradithya: Updated

pkg/clusterresource/controller.go
```diff
@@ -452,6 +411,147 @@ func (c *controller) syncNamespace(ctx context.Context, project models.Project,
...
+	err := json.Unmarshal(patch, &patchMap)
+	if err != nil {
+		return nil, err
+	}
+	u := unstructured.Unstructured{Object: patchMap}
+	a, err := meta.Accessor(&u)
+	if err != nil {
+		return nil, err
+	}
+	a.SetResourceVersion(rv)
```
kumare3, line 446: wow this is a little dangerous right?

kumare3 (dismissed):

kumare3:
@pradithya i think you are missing a DCO

pkg/clusterresource/controller.go
pradithya: Based on the docs it seems ok as long as we don't modify it.

```
	// String that identifies the server's internal version of this object that
	// can be used by clients to determine when objects have changed.
	// Value must be treated as opaque by clients and passed unmodified back to the server.
	// Populated by the system.
	// Read-only.
	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency
	// +optional
	ResourceVersion string `json:"resourceVersion,omitempty" protobuf:"bytes,2,opt,name=resourceVersion"`

```

pradithya:
Updated the DCO

katrogan (approved):

PR #297, opened 2021-11-23
Add nil checks and avoid nil returns

EngHabu:
Signed-off-by: Haytham Abuelfutuh <haytham@afutuh.com>

# TL;DR
_Please replace this text with a description of what this PR accomplishes._

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added
 - [X] Any pending items have an associated Issue

## Complete description
There are at least a couple of places where it's possible to end up with nil objects returned from GetInputs and GetOutputs functions. If that happens, the rest of the code doesn't check for nils and causes nil refs.

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1858

### DIFF
--- a/pkg/manager/impl/util/data.go
+++ b/pkg/manager/impl/util/data.go
@@ -22,14 +22,17 @@ func shouldFetchOutputData(config *runtimeInterfaces.RemoteDataConfig, urlBlob a
 	return len(outputURI) > 0 && shouldFetchData(config, urlBlob)
 }
 
-// Returns an inputs URL blob and if config settings permit, inline inputs data for an execution.
+// GetInputs returns an inputs URL blob and if config settings permit, inline inputs data for an execution.
 func GetInputs(ctx context.Context, urlData dataInterfaces.RemoteURLInterface,
 	remoteDataConfig *runtimeInterfaces.RemoteDataConfig, storageClient *storage.DataStore, inputURI string) (
 	*core.LiteralMap, *admin.UrlBlob, error) {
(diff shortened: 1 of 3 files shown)


wild-endeavor (approved):

pkg/manager/impl/util/data.go
```diff
@@ -22,14 +22,17 @@ func shouldFetchOutputData(config *runtimeInterfaces.RemoteDataConfig, urlBlob a
...
+// GetInputs returns an inputs URL blob and if config settings permit, inline inputs data for an execution.
 func GetInputs(ctx context.Context, urlData dataInterfaces.RemoteURLInterface,
 	remoteDataConfig *runtimeInterfaces.RemoteDataConfig, storageClient *storage.DataStore, inputURI string) (
 	*core.LiteralMap, *admin.UrlBlob, error) {
+	var inputsURLBlob admin.UrlBlob
+	var fullInputs core.LiteralMap
+
 	if len(inputURI) == 0 {
-		return nil, nil, nil
+		return &fullInputs, &inputsURLBlob, nil
```
ckiosidis, line 33: aren't these values nil here?

EngHabu: They are not nil.. they are default values of the struct types... which are valid struct objects with default values for their fields... I think that more accurately represent the intended behavior, if there is no inputURI, inputs are empty...

PR #305, opened 2021-12-03
Migrate auth to security context

pmahindrakar-oss:

# TL;DR
Added usage of security context if it passed in the request or launch spec which overrides the auth request or launch spec values.

Also modified the flyteadmin_config.yaml to use the right minio port and s3 bucket when running in local dev mode.
This makes it in sync with sandbox

Testing Done :
Created execution request from flytectl which passes in the security context.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/1722

## Follow-up issue
_NA_


pkg/workflowengine/impl/prepare_execution.go
```diff
@@ -40,6 +40,21 @@ func addPermissions(auth *admin.AuthRole, roleNameKey string, flyteWf *v1alpha1.
 	}
 }
 
+func addPermissionsFromSecurtiyCtx(securityCtx *core.SecurityContext, roleNameKey string, flyteWf *v1alpha1.FlyteWorkflow) {
```
katrogan, line 43: what if you just pass the security context to addPermissions and have it prefer values from non-empty securitycontexts so that permissions handling is done in one place?

pkg/manager/impl/execution_manager.go
```diff
@@ -672,6 +673,20 @@ func resolvePermissions(request *admin.ExecutionCreateRequest, launchPlan *admin
 	return &admin.AuthRole{}
 }
 
+func resolvePermissionsFromSecurityCtx(ctx context.Context, request admin.ExecutionCreateRequest, launchPlan *admin.LaunchPlan) *core.SecurityContext {
```
katrogan, line 676:
```suggestion
func resolveSecurityCtx(ctx context.Context, request admin.ExecutionCreateRequest, launchPlan *admin.LaunchPlan) *core.SecurityContext {
```

pkg/manager/impl/execution_manager.go
```diff
@@ -672,6 +673,20 @@ func resolvePermissions(request *admin.ExecutionCreateRequest, launchPlan *admin
...
+	// Use security context from the request if its set
+	if request.Spec.SecurityContext != nil {
+		return request.Spec.SecurityContext
+	}
+
+	// Use launchplans security context if its set
+	if launchPlan.Spec.SecurityContext != nil {
+		return launchPlan.Spec.SecurityContext
```
katrogan, line 684: nit: you can always just return this value when `request.Spec.SecurityContext == nil`, no need to wrap it in an if block

flyteadmin_config.yaml
```diff
@@ -64,7 +64,7 @@ database:
   port: 5432
   username: postgres
   host: localhost
-  dbname: postgres
+  dbname: flyteadmin
```
EngHabu, line 67: Until we fix this: https://github.com/flyteorg/flyte/issues/1882, I don't think we should attempt to use anything but the default DB...

pkg/manager/impl/execution_manager.go
```diff
@@ -672,6 +673,20 @@ func resolvePermissions(request *admin.ExecutionCreateRequest, launchPlan *admin
...
+	}
+	logger.Warn(ctx, "Returning an empty security context")
```
EngHabu, line 686: I think we should fall back on building Security Context from AuthRole if they are not specified in the request/launchplan... for the transition to deprecate AuthRole, I think Admin should be the place to make this translation...

flyteadmin_config.yaml
pmahindrakar-oss: Ok.reverted this to older values of postgress

pkg/manager/impl/execution_manager.go
pmahindrakar-oss: Agreed. Done

pkg/manager/impl/execution_manager.go
pmahindrakar-oss: Actually there is additional logic due to which we cant return just the launplan.Spec.SecurityContext since we now want to get this value from resolvedAuthRole. Let me know if that works

```
// Use launchplans security context if its set
	if launchPlan.Spec.SecurityContext != nil {
		return launchPlan.Spec.SecurityContext
	}
	logger.Warn(ctx, "Setting security context from auth Role")
	return &core.SecurityContext{
		RunAs: &core.Identity{
			IamRole:           resolvedAuthRole.AssumableIamRole,
			K8SServiceAccount: resolvedAuthRole.KubernetesServiceAccount,
		},
	}
```

pkg/manager/impl/execution_manager.go
pmahindrakar-oss: Done

pkg/workflowengine/impl/prepare_execution.go
pmahindrakar-oss: Done .Added a wrapper addPermissions which accepts both authRole and securityCtx

pkg/manager/impl/execution_manager.go
katrogan: looks great!

pmahindrakar-oss:
PTAL @katrogan @EngHabu

pkg/manager/impl/execution_manager.go
```diff
@@ -568,13 +568,16 @@ func (m *ExecutionManager) launchSingleTaskExecution(
...
+	resolvedAuthRole := resolvePermissions(request, launchPlan)
+	resolvedSecurityCtx := resolveSecurityCtx(ctx, request, launchPlan, resolvedAuthRole)
 	executionParameters := workflowengineInterfaces.ExecutionParameters{
 		Inputs:              request.Inputs,
 		AcceptedAt:          requestedAt,
 		Labels:              labels,
 		Annotations:         annotations,
 		ExecutionConfig:     executionConfig,
-		Auth:                resolvePermissions(&request, launchPlan),
+		Auth:                resolvedAuthRole,
```
EngHabu, line 579: So is it possible to just remove Auth field completely from ExecutionParameters? is this used anywhere that's not FlyteAdmin? if it only lives in flyteAdmin, then we can update the code to always use the SecurityContext and we make sure it's populated here.... right?

katrogan: +1 we should move away from passing Auth objects

pmahindrakar-oss: Make sense. Its not used outside admin. Have removed it from ExecutionParameters. It will remain in the ExecutionCreateRequest though until all clients migrate after which we can remove it completely along with the protos.

pmahindrakar-oss: Done

katrogan: one more thing to remove from flyteidl in flyte 1.0.0 😄

katrogan (approved):

PR #307, opened 2021-12-15
Execution event cluster validation

katrogan:

# TL;DR
Allow modifying the cluster state for queued workflow execution events and validate all other events come from the expected cluster.

---
**NOTE**

We could optimize this call by adding a cluster column for node and task executions in the db, i'm not sure if that's entirely necessary though

---

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1929

## Follow-up issue
_NA_


pkg/manager/impl/validation/shared_execution.go
```diff
@@ -0,0 +1,34 @@
...
+		Name:    executionID.Name,
+	})
+	if err != nil {
+		logger.Debugf(ctx, "Failed to find existing execution with id [%+v] with err: %v", executionID, err)
+		return err
+	}
+	if workflowExecution.Cluster != cluster {
+		errorMsg := fmt.Sprintf("Cluster/producer from event [%s] does not match existing workflow execution cluster: [%s]",
+			workflowExecution.Cluster, cluster)
+		return errors.NewIncompatibleClusterError(ctx, errorMsg, workflowExecution.Cluster)
```
pmahindrakar-oss, line 31: Can we also add name of the cluster it was compared against. i.e cluster string in the IncompatibleClusterError

pkg/manager/impl/validation/shared_execution.go
```diff
@@ -0,0 +1,34 @@
...
+
+func ValidateCluster(ctx context.Context, db repositories.RepositoryInterface, executionID *core.WorkflowExecutionIdentifier, cluster string) error {
+	if len(cluster) == 0 || cluster == common.DefaultProducerID {
+		return nil
+	}
+	workflowExecution, err := db.ExecutionRepo().Get(ctx, repoInterfaces.Identifier{
+		Project: executionID.Project,
+		Domain:  executionID.Domain,
+		Name:    executionID.Name,
+	})
```
pmahindrakar-oss, line 23: We do already check if execution exists here https://github.com/flyteorg/flyteadmin/blob/296655a415d36e20bdedf98e41f890eb4901480c/pkg/manager/impl/node_execution_manager.go#L92 before calling ValidateCluster.

Can ValidateCluster be made independent of the DB call inorder to reduce the same checks on the DB or probably have a common refactored Validate function which does these checks

pkg/manager/impl/validation/shared_execution.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/flyteorg/flyteadmin/pkg/common"
+	"github.com/flyteorg/flyteadmin/pkg/errors"
+	"github.com/flyteorg/flyteadmin/pkg/repositories"
+	repoInterfaces "github.com/flyteorg/flyteadmin/pkg/repositories/interfaces"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
+	"github.com/flyteorg/flytestdlib/logger"
+)
+
+func ValidateCluster(ctx context.Context, db repositories.RepositoryInterface, executionID *core.WorkflowExecutionIdentifier, cluster string) error {
```
pmahindrakar-oss, line 15: May be a doc that this validate executionId belongs to passed in cluster string.

Also can we mention somewhere why its ok to have workflowExecution.Cluster different than common.DefaultProducerID and still raise no error

pmahindrakar-oss (dismissed):

EngHabu (dismissed):
Looks good to me... Praful had some comments though...

pkg/manager/impl/validation/shared_execution.go
katrogan: that's already included in the error message defined in the line above

pkg/manager/impl/validation/shared_execution.go
katrogan: Already done here: https://github.com/flyteorg/flyteadmin/pull/307/files#diff-89fdbc388b31a2407358d353ba0dc95fb7b9a4862b0528fc0792aabb67e8b178R15 but I can copy the comment

pkg/manager/impl/validation/shared_execution.go
katrogan: sure, done

pkg/manager/impl/validation/shared_execution.go
pmahindrakar-oss: Ok . that should be good then.

katrogan:
PTAL @pmahindrakar-oss @EngHabu

EngHabu (approved):

PR #309, opened 2021-12-22
Validate datetime

lupasarin:
## _Read then delete this section_

_- Make sure to use a concise title for the pull-request._

_- Use #patch, #minor or #major in the pull-request title to bump the corresponding version. Otherwise, the patch version
will be bumped. [More details](https://github.com/marketplace/actions/github-tag-bump)_

# TL;DR
_Please replace this text with a description of what this PR accomplishes._

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


README.md
```diff
@@ -24,6 +24,8 @@ Flyte Admin has a few useful make targets for linting and testing. Please use th
 minor bugs and linting errors.
 
 ```
+  # Please make sure you have goimports installed, i.e.
+  $ go install golang.org/x/tools/cmd/goimports@latest
```
katrogan, line 28: you can actually run `make install` which should download all the go tooling you need to run goimports

katrogan:
thanks @lupasarin for making this change! can you also update [validateLiteralMap](https://github.com/flyteorg/flyteadmin/blob/269b482c45b07352f9c1c78a2b85e14bd85d4266/pkg/manager/impl/validation/validation.go#L193) which is used for launch plan and create execution validation with a similar check?

pkg/manager/impl/validation/validation_test.go
```diff
@@ -334,3 +358,42 @@ func TestValidateOutputData(t *testing.T) {
...
+					Value: &core.Scalar_Primitive{
+						Primitive: &core.Primitive{
+							Value: &core.Primitive_Datetime{Datetime: timestamppb.New(timestamp)},
+						},
+					},
+				},
+			},
+		}))
+	})
+	// cannot generate invalid timestamp with wrong format or e.g. too big month or year.
```
katrogan, line 397: one option here is to generate the invalid timestamp proto from your java application, dump this to json and save that json blob as a test file in this repo. Then, to implement this test case you can parse the proto from json using [jsonpb Unmarshal](https://pkg.go.dev/github.com/golang/protobuf/jsonpb) and validate that

pkg/manager/impl/validation/validation.go
```diff
@@ -268,3 +277,38 @@ func ValidateOutputData(outputData *core.LiteralMap, maxSizeInBytes int64) error
...
+	err := timestamp.CheckValid()
+	if err != nil {
+		return errors.NewFlyteAdminErrorf(codes.InvalidArgument, err.Error())
+	}
+	return nil
+}
+
+func isDateTime(input *core.Literal) bool {
+	literalValue := input.Value
+	if reflect.ValueOf(literalValue).String() != "<*core.Literal_Scalar Value>" {
```
katrogan, line 297: instead of using reflection can you just check if the oneof value is nil, something like

`return input.GetScalar() != nil && input.GetScalar().GetPrimitive() != nil && inpuGetScalar().GetPrimitive().GetDatetime() != nil`

pkg/manager/impl/validation/validation_test.go
lupasarin: Thanks for the suggestion! We finally found out we could do this using timestamppb type directly, constructed using Seconds+Nanos.

pkg/manager/impl/validation/validation.go
lupasarin: Done, thanks for the hint!

pkg/manager/impl/validation/validation_test.go
katrogan: even better, awesome!

pkg/manager/impl/validation/validation.go
```diff
@@ -268,3 +276,21 @@ func ValidateOutputData(outputData *core.LiteralMap, maxSizeInBytes int64) error
 	}
 	return errors.NewFlyteAdminErrorf(codes.ResourceExhausted, "Output data size exceeds platform configured threshold (%+v > %v)", outputSizeInBytes, maxSizeInBytes)
 }
+
+func ValidateDatetime(literal *core.Literal) error {
+	if literal == nil {
+		return nil
+	}
+
+	timestamp := literal.Value.(*core.Literal_Scalar).Scalar.Value.(*core.Scalar_Primitive).Primitive.GetValue().(*core.Primitive_Datetime).Datetime
```
katrogan, line 285:
```suggestion
	timestamp := literal.GetScalar().GetPrimitive().GetDatetime()
```

katrogan: (also because of the DCO requirement please don't actually accept this suggestion from the Github UI because it won't be signed off 😅 )

katrogan (dismissed):
Thank you for updating all the function comments and refactoring tests into specific, named cases! Just one more suggestion and should be good to go 🙏

pkg/manager/impl/validation/validation.go
lupasarin: Thanks a lot for the suggestion! Now it's done :)

pkg/manager/impl/validation/validation.go
```diff
@@ -268,3 +276,21 @@ func ValidateOutputData(outputData *core.LiteralMap, maxSizeInBytes int64) error
 	}
 	return errors.NewFlyteAdminErrorf(codes.ResourceExhausted, "Output data size exceeds platform configured threshold (%+v > %v)", outputSizeInBytes, maxSizeInBytes)
 }
+
+func ValidateDatetime(literal *core.Literal) error {
+	if literal == nil {
+		return nil
+	}
```
ckiosidis, line 283: does this nil check make sense?

the fixedInput should not be nil as far as i understand.
This default behaviour will not protect us from future refactorings

lupasarin: Do you mean that we should fail if literal == nil? Or that it will never happen?

ckiosidis: I think so and if necessary do refactorings on the places of usage to accomodate for the nil usecase

lupasarin: As mentioned via private message, I changed the check to throw an error instead. Just to be on the safest side.

pkg/manager/impl/validation/validation_test.go
```diff
@@ -334,3 +359,88 @@ func TestValidateOutputData(t *testing.T) {
...
+						Primitive: &core.Primitive{
+							Value: &core.Primitive_Datetime{Datetime: &timestamp},
+						},
+					},
+				},
+			},
+		}))
+	})
+	t.Run("datetime with value above max", func(t *testing.T) {
+		timestamp := timestamppb.Timestamp{Seconds: 253402300800, Nanos: 0} // = 0000-12-31T23:59:59.999999999Z
```
ckiosidis, line 429: let's also add a test for the java minimum (which should not be supported)

```
MIN
The minimum supported Instant, '-1000000000-01-01T00:00Z'.

lupasarin: Done!

katrogan (approved):

katrogan (approved):

PR #318, opened 2022-01-12
Migrate cluster resource controller to call admin service endpoints

katrogan:
# TL;DR
To support scenarios where cluster resource controller is deployed on a separate cluster without flyteadmin database credentials, this change moves all db call to flyteadmin service calls.

A big part of this change was refactoring execution cluster target logic so that a separate component is used to list **all** execution cluster targets. This is now independent of the GetTarget() logic which depends on database access to fetch matchable attributes overrides.

Tested on flyte sandbox by deleting a project quota, having the updated cluster resource controller run, and then asserting the project quota got re-added.

Note, that we'll still need to keep the execution cluster config as part of deploying the cluster resource controller, but if it's deployed on the same cluster as the flytepropeller cluster, the in target cluster provider will suffice.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2059

## Follow-up issue
_NA_


go.mod
```diff
@@ -44,6 +44,7 @@ require (
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.0
+	github.com/zalando/go-keyring v0.1.1
```
pmahindrakar-oss, line 47: I remember having this discussion with @EngHabu  where we didn't want to add this dependency in stdlib and keep it only in flytectl where we used the keyring 
Now that we have flyteadmin also needing this dependency, is it worth moving this to stdlib and also Token implemetation

cmd/entrypoints/util/token_cache_keyring.go
```diff
@@ -0,0 +1,58 @@
+package pkce
+
```
pmahindrakar-oss, line 2: May be we can move this from flytectl to stdlib so that admin can consume aswell?

go.mod
katrogan: no longer relevant - marking resolved

cmd/entrypoints/util/token_cache_keyring.go
katrogan: thanks for the callout, removed!

cmd/entrypoints/clusterresource.go
```diff
@@ -3,71 +3,66 @@ package entrypoints
...
 	Short: "This command administers the ClusterResourceController. Please choose a subcommand.",
 }
 
-func GetLocalDbConfig() repositoryConfig.DbConfig {
-	return repositoryConfig.DbConfig{
-		Host:   "localhost",
-		Port:   5432,
-		DbName: "postgres",
-		User:   "postgres",
+func getClusterResourceController(ctx context.Context) clusterresource.Controller {
```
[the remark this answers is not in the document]
katrogan: sounds good, done

katrogan:
PTAL @EngHabu @pmahindrakar-oss

cmd/entrypoints/clusterresource.go
```diff
@@ -3,32 +3,50 @@ package entrypoints
...
+	if err != nil {
+		panic(err)
+	}
+
+	clientSet, err := admin.ClientSetBuilder().WithConfig(admin.GetConfig(ctx)).Build(ctx)
+	if err != nil {
+		panic(err)
 	}
+
+	return clusterresource.NewClusterResourceController(clientSet.AdminClient(), listTargetsProvider, scope)
```
EngHabu, line 49: Will it crash if it can't talk to admin? just making sure it'll... specially for enabling auth in admin, this needs to know to detect that and crash...

cmd/entrypoints/util/token_cache_keyring.go
EngHabu: Still see this...

EngHabu (approved):

cmd/entrypoints/clusterresource.go
katrogan: it will dump a bunch of error logs complaining admin is unreachable and fail

cmd/entrypoints/util/token_cache_keyring.go
katrogan: sorry, removed the references to it but forgot to delete. mind taking another look?

EngHabu (approved):

PR #322, opened 2022-01-18
Fixed k8secret entrypoint script and added Status info get execution response

pmahindrakar-oss:
# TL;DR

* Fixes the k8secret entrypoint script for listing target providers when there are empty clusterconfigs.
* Added stateOccuredAt time to model and UpdateExecution impl
* Returns ExecutionStatus in the execution closure. Also added backward compatible responses for the same.

## Type
 - [X] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue

https://github.com/flyteorg/flyte/issues/2045

## Follow-up issue
_NA_


cmd/entrypoints/k8s_secret.go
```diff
@@ -99,7 +102,15 @@ func persistSecrets(ctx context.Context, _ *pflag.FlagSet) error {
 	initializationErrorCounter := scope.NewSubScope("secrets").MustNewCounter(
 		"flyteclient_initialization_error",
 		"count of errors encountered initializing a flyte client from kube config")
-	listTargetsProvider, err := executioncluster.NewListTargets(initializationErrorCounter, executioncluster.NewExecutionTargetProvider(), configuration.ClusterConfiguration())
+
+	var listTargetsProvider interfaces.ListTargetsInterface
```
katrogan, line 106: thank you for fixing!

pkg/manager/impl/execution_manager_test.go
```diff
@@ -1820,6 +1835,81 @@ func TestGetExecution_TransformerError(t *testing.T) {
...
+				Project: "project",
+				Domain:  "domain",
+			},
+		})
+		assert.Error(t, err)
+	})
+
+	t.Run("empty status passed", func(t *testing.T) {
+		repository := repositoryMocks.NewMockRepository()
+		updateExecFunc := func(ctx context.Context, execModel models.Execution) error {
```
katrogan, line 1853: can we also add an assert that this gets called?

pkg/manager/impl/execution_manager_test.go
```diff
@@ -1820,6 +1835,81 @@ func TestGetExecution_TransformerError(t *testing.T) {
...
+		updateResponse, err := execManager.UpdateExecution(context.Background(), admin.ExecutionUpdateRequest{
+			Id: &executionIdentifier,
+		})
+		assert.NoError(t, err)
+		assert.NotNil(t, updateResponse)
+	})
+
+	t.Run("archive status passed", func(t *testing.T) {
+		repository := repositoryMocks.NewMockRepository()
+		updateExecFunc := func(ctx context.Context, execModel models.Execution) error {
```
katrogan, line 1869: ditto

pkg/repositories/config/migrations.go
```diff
@@ -359,6 +359,28 @@ var Migrations = []*gormigrate.Migration{
...
+		ID: "2022-01-11-execution-state",
+		Migrate: func(tx *gorm.DB) error {
+			return tx.AutoMigrate(&models.Execution{})
+		},
+		Rollback: func(tx *gorm.DB) error {
+			return tx.Table("execution").Migrator().DropColumn(&models.Execution{}, "state")
+		},
+	},
+
+	// Add stateUpdatedAt to execution model
```
katrogan, line 374: why do these need to be two separate migrations?

pkg/repositories/models/execution.go
```diff
@@ -56,4 +56,10 @@ type Execution struct {
 	// The user responsible for launching this execution.
 	// This is also stored in the spec but promoted as a column for filtering.
 	User string `gorm:"index" valid:"length(0|255)"`
+	// GORM doesn't save the zero value for ints, so we use a pointer for the State field
+	State *int32 `gorm:"default:0"`
+	// State updated at to indicate the archival/activation time
+	StateUpdatedAt *time.Time
```
katrogan, line 62: nit: does this need to be a model column? can we just store it in the execution closure?

pkg/repositories/transformers/execution.go
```diff
@@ -228,6 +236,41 @@ func UpdateExecutionModelState(
...
+
+	var closure admin.ExecutionClosure
+	err := proto.Unmarshal(executionModel.Closure, &closure)
+	if err != nil {
+		return errors.NewFlyteAdminErrorf(codes.Internal, "Failed to unmarshal execution closure: %v", err)
+	}
+	// Update the indexed columns
+	stateInt := int32(state)
+	executionModel.State = &stateInt
+	occurredAt := time.Now()
```
katrogan, line 251: can we pass this in to the function and determine the request time in the pkg/rpc/ or manager layer?

pkg/repositories/models/execution.go
```diff
@@ -56,4 +56,10 @@ type Execution struct {
...
+	// The user responsible for activating/archiving this execution.
+	StateUpdatedBy string `gorm:"index" valid:"length(0|255)"`
```
katrogan, line 64: why does this need an index?

pkg/repositories/models/execution.go
```diff
@@ -56,4 +56,10 @@ type Execution struct {
 	// The user responsible for launching this execution.
 	// This is also stored in the spec but promoted as a column for filtering.
 	User string `gorm:"index" valid:"length(0|255)"`
+	// GORM doesn't save the zero value for ints, so we use a pointer for the State field
+	State *int32 `gorm:"default:0"`
```
katrogan, line 60: should this be indexed?

pkg/repositories/transformers/execution.go
```diff
@@ -293,6 +343,22 @@ func FromExecutionModel(executionModel models.Execution) (*admin.Execution, erro
 	}, nil
 }
 
+func GetStateFromModelOldExecs(executionModel models.Execution) (*admin.ExecutionStateChangeDetails, error) {
```
katrogan, line 346: not a huge fan of this name, could we change it to something like 'PopulateDefaultStateChangeDetails` and add a comment on why it exists?

pkg/repositories/transformers/execution_test.go
```diff
@@ -481,15 +488,24 @@ func TestFromExecutionModel(t *testing.T) {
...
 		StartedAt:      startedAtProto,
+		StateChangeDetails: &admin.ExecutionStateChangeDetails{
+			State:      admin.ExecutionState_EXECUTION_ACTIVE,
+			OccurredAt: createdAtProto,
+		},
 	}
 	closureBytes, _ := proto.Marshal(&closure)
 
 	executionModel := models.Execution{
+		BaseModel: models.BaseModel{
```
katrogan, line 506: can you add the expected state column values to the execution model too?

pkg/rpc/adminservice/execution.go
```diff
@@ -142,6 +142,31 @@ func (m *AdminService) GetExecution(
 	return response, nil
 }
 
+func (m *AdminService) UpdateExecution(
+	ctx context.Context, request *admin.ExecutionUpdateRequest) (*admin.ExecutionUpdateResponse, error) {
+	defer m.interceptPanic(ctx, request)
+	requestedAt := time.Now()
```
katrogan, line 148: can we use this requestedAt as the state change occurred at value?

pmahindrakar-oss: Yes . works well. Also it helps in better unit testing the execution manager as this value is injected

pkg/repositories/transformers/execution.go
pmahindrakar-oss: Done

pkg/repositories/models/execution.go
pmahindrakar-oss: Added. This would be useful for filtering

pkg/repositories/models/execution.go
pmahindrakar-oss: I think if we have time range queries  to get set of executions then this might be useful, but we can always decide this based on the usecases and then add them later. I have removed it for now

pkg/repositories/transformers/execution.go
pmahindrakar-oss: Done

pkg/repositories/models/execution.go
pmahindrakar-oss: Thought was to support faster range based queries for state, but we can always add it later as a separate column when the need arises. I have currently kept this only in the closure

pkg/repositories/config/migrations.go
pmahindrakar-oss: NA as i have removed the stateUpdatedAt column now.

The thought was to have simple logic in rollback . 

If we have two columns migrated in the same migration key then we might have to unnecessarily rollback a successful column migration due to failed migration on other columns 
OR we might have to handle those cases in the rollback. 

I think it works well if the columns are related which is true in this case since both are related to state details change, but if they are unrelated then i am not sure if we should fail the migration if one of the column migrations failed in a multicolumn migration

Eg : the two columns could be for two different features , then it would block both the features.
Let me know what you think

pkg/repositories/transformers/execution_test.go
pmahindrakar-oss: Done

pkg/repositories/models/execution.go
katrogan: sure sounds good! i don't really anticipate we'll want to query on state changes immediately but this will be okay to not have backfilled if we change our minds

pkg/repositories/config/migrations.go
katrogan: agreed, since the column additions are part of the same change it makes sense that the migration is "transactional" in this case and both should be added in one shot

katrogan (commented):
Do we need StateUpdatedAtBy in the model: https://github.com/flyteorg/flyteadmin/pull/322/files#diff-11c27d9c35c6ddfead817ee15718acf15ac4b84f1933c20311c8de38018affb6R62? I don't believe that the console will necessarily filter on this, it's just useful to record the action itself somewhere in the closure

pkg/repositories/config/migrations.go
```diff
@@ -359,6 +359,17 @@ var Migrations = []*gormigrate.Migration{
...
 	},
+
+	// Add state to execution model.
+	{
+		ID: "2022-01-11-execution-state",
+		Migrate: func(tx *gorm.DB) error {
+			return tx.AutoMigrate(&models.Execution{})
+		},
+		Rollback: func(tx *gorm.DB) error {
+			return tx.Table("execution").Migrator().DropColumn(&models.Execution{}, "state")
```
katrogan, line 370: and "state_updated_by" too?

pkg/repositories/models/execution.go
katrogan: I still see it in the model definition

pkg/repositories/config/migrations.go
pmahindrakar-oss: removed stateUpdatedBy column as we are tracking it as part of the closure

katrogan (approved):

EngHabu (approved):

PR #326, opened 2022-01-20
Configurable cluster resource sync: in cluster or standalone

katrogan:

# TL;DR
Refactor cluster resource controller sync to either use an in-cluster implementation which directly calls the database to fetch cluster resource matchable attributes and all user projects OR to accomplish the same using the admin service endpoint.

The former implementation is useful for the flytesandbox scenario where we run cluster resource sync as an initContainer and keeps backwards compatibility for users running an authenticated admin service. This refactor allows us to still deploy the sync process as a standalone deployment that fetches data solely using the flyteadmin service.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2059

## Follow-up issue
_NA_


cmd/entrypoints/clusterresource.go
```diff
@@ -41,12 +48,37 @@ func getClusterResourceController(ctx context.Context, scope promutils.Scope, co
 		panic(err)
 	}
 
-	clientSet, err := admin.ClientSetBuilder().WithConfig(admin.GetConfig(ctx)).Build(ctx)
-	if err != nil {
-		panic(err)
+	var adminDataProvider interfaces.FlyteAdminDataProvider
+	if configuration.ClusterResourceConfiguration().IsStandaloneDeployment() {
+		clientSet, err := admin.ClientSetBuilder().WithConfig(admin.GetConfig(ctx)).Build(ctx)
```
pmahindrakar-oss, line 53: What happens when admin has auth enabled.
Does this need to be built with tokenCacheProvider?

cmd/entrypoints/clusterresource.go
```diff
@@ -41,12 +48,37 @@ func getClusterResourceController(ctx context.Context, scope promutils.Scope, co
...
+			BaseConfig: repositoryConfig.BaseConfig{
+				LogLevel: dbLogLevel,
+			},
+			Host:         dbConfigValues.Host,
+			Port:         dbConfigValues.Port,
+			DbName:       dbConfigValues.DbName,
+			User:         dbConfigValues.User,
+			Password:     dbConfigValues.Password,
+			ExtraOptions: dbConfigValues.ExtraOptions,
+		}
```
pmahindrakar-oss, line 74: I have seen this been constructed in different places. Is it possible to restrict this object construction at one place as the DBConfig is the common one being used by all consumers including admin and some init containers

pmahindrakar-oss (dismissed):

cmd/entrypoints/clusterresource.go
katrogan: it shouldn't, it's never called by an end user's machine but from a service

cmd/entrypoints/clusterresource.go
katrogan: oh good call, in fact this already exists as a method 😅 updated

pmahindrakar-oss (approved):

PR #351, opened 2022-02-18
Using isDynamic flag for distinguishing subworkflows and dynamic workflows

pmahindrakar-oss:

# TL;DR

Follow up to this PR https://github.com/flyteorg/flyteidl/pull/256
Testing Done is updated in the above PR

TODO : Update the flyteidl to released version once PR 256 is checkedin

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue

https://github.com/flyteorg/flyte/issues/2132

### DIFF
--- a/pkg/repositories/transformers/node_execution.go
+++ b/pkg/repositories/transformers/node_execution.go
@@ -240,6 +240,9 @@ func FromNodeExecutionModel(nodeExecutionModel models.NodeExecution) (*admin.Nod
 	}
 	if len(nodeExecutionModel.ChildNodeExecutions) > 0 {
 		nodeExecutionMetadata.IsParentNode = true
+		if len(nodeExecutionModel.DynamicWorkflowRemoteClosureReference) > 0 {
+			nodeExecutionMetadata.IsDynamic = true
+		}
 	}
 	return &admin.NodeExecution{
 		Id: &core.NodeExecutionIdentifier{
--- a/go.mod
+++ b/go.mod
@@ -168,3 +168,5 @@ require (
 )
 
 replace github.com/robfig/cron/v3 => github.com/unionai/cron/v3 v3.0.2-0.20210825070134-bfc34418fe84
+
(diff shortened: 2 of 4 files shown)


pkg/repositories/transformers/node_execution.go
```diff
@@ -240,6 +240,9 @@ func FromNodeExecutionModel(nodeExecutionModel models.NodeExecution) (*admin.Nod
 	}
 	if len(nodeExecutionModel.ChildNodeExecutions) > 0 {
 		nodeExecutionMetadata.IsParentNode = true
+		if len(nodeExecutionModel.DynamicWorkflowRemoteClosureReference) > 0 {
```
katrogan, line 243: is this right? it won't be backwards compatible for very old executions if it's a concern. should we checking against the workflow closure? cc @EngHabu

pmahindrakar-oss: @katrogan , This field was introduced in April 21 . I agree older executions than these we wont be able to fetch this info.

I see the same issue in using the NodeExecutionClosure.
The taskNodeMetadata is not being populated with DyanmicWorkflow
https://github.com/flyteorg/flyteidl/blob/master/gen/pb-go/flyteidl/admin/node_execution.pb.go#L678

Whereas the Event does get populated with it. Admin doesn't save it 
https://github.com/flyteorg/flyteidl/blob/0ff9b16ad4c37971dfaa1e51b5fe970bae40e356/gen/pb-go/flyteidl/event/event.pb.go#L464

If we add the change to now save this closure , then this will only be available for new dynamic workflows.

@katrogan can you elaborate on how we can check against workflow closure

EngHabu: So what we lose with this for those very old executions is that we won't visualize the dynamic graph... and I think that's ok because they fully rely on the DynamicWorkflowRemoteClosureReference anyway to do so... having this flag will just simplify their logic a tad bit...

EngHabu (approved):

PR #352, opened 2022-02-18
Use database filters to ensure execution event phase transitions are valid

katrogan:
# TL;DR
Use database filters to ensure execution event phase transitions are valid.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/2186

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/executions/phase_filters.go
+++ b/pkg/manager/impl/executions/phase_filters.go
@@ -103,7 +103,7 @@ func GetUpdateNodeExecutionFilters(eventPhase core.NodeExecution_Phase) (filters
 func GetUpdateTaskExecutionFilters(eventPhase core.TaskExecution_Phase) (filters []common.InlineFilter, err error) {
 	// With terminal executions we don't need to add the check for the same phase because we always check the existing
 	// execution isn't already terminal. RUNNING can have multiple phase versions
-	if !(common.IsTaskExecutionTerminal(eventPhase) || eventPhase != core.TaskExecution_RUNNING) {
+	if !(common.IsTaskExecutionTerminal(eventPhase) || eventPhase == core.TaskExecution_RUNNING) {
 		notAlreadySamePhaseFilter, err := common.NewSingleValueFilter(common.Execution, common.NotEqual, phaseField, eventPhase.String())
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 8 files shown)


pkg/manager/impl/executions/phase_filters.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument, "Unrecognized execution entity [%+v] for non-terminal filters", entity)
+	}
+}
+
+func newSchedulingFilter() (common.InlineFilter, error) {
+	return common.NewRepeatedValueFilter(common.Execution, common.ValueIn, "phase", SchedulingWorkflowPhases)
+}
+
+func GetUpdateExecutionFilters(eventPhase core.WorkflowExecution_Phase) (filters []common.InlineFilter, err error) {
+	// With the exeception of queued events, it's never acceptable to move into the same phase.
```
EngHabu, line 57: Hmmm I'm not sure this is accurate. If propeller picks up a WF, sends running event then fails to write the CRD back to ApiServer. It'll attempt to do the same operation again... why is this a bad thing not just an idempotent operation?

EngHabu (commented):
I've one question..... but as an overall design question, it feels like we are trying to replicate what RevisionVersion does for CRDs for WFs and Tasks using Phases... 
In the case of K8s, it allows any kind of update as long as the caller knows what they are doing (i.e.  they have the latest version of the CRD)...

pkg/manager/impl/executions/phase_filters.go
katrogan: oh interesting, the current implementation doesn't let us update to an existing phase. i thought propeller swallowed the already exists in that case e.g. https://github.com/flyteorg/flyteadmin/blob/master/pkg/manager/impl/execution_manager.go#L1183
andhttps://github.com/flyteorg/flytepropeller/blob/c016dabbfef6037bead59590b42326dabe89f957/pkg/controller/workflow/executor.go#L261

perhaps propeller should also be modified to swallow the NotFound when RowsAffected == 0 gets triggered as part of meeting the new phase conditions. but perhaps that is getting too finnicky

katrogan:
Re: your point above, I agree that this verification is a little clunky but I worry doing a get and then update will incur a performance penalty. what do you think? if we pursue that i don't think the issue this change addresses is fundamentally significant enough to introduce that design in the short-term :\

PR #356, opened 2022-02-23
Validate k8s pods resource

pingsutw:
Signed-off-by: Kevin Su <pingsutw@apache.org>

# TL;DR
Resources would be checked when using pod tasks like they are checked in python tasks.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Invoke `validateK8sPod` if running pod tasks

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2187

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/validation/task_validator.go
+++ b/pkg/manager/impl/validation/task_validator.go
@@ -45,7 +45,8 @@ func validateContainer(task core.TaskTemplate, taskConfig runtime.TaskResourceCo
 // This is called for a task with a non-nil k8s pod.
 func validateK8sPod(task core.TaskTemplate, taskConfig runtime.TaskResourceConfiguration) error {
 	if task.GetK8SPod().PodSpec == nil {
-		return nil
+		return errors.NewFlyteAdminErrorf(codes.InvalidArgument,
+			"invalid TaskSpecification, pod tasks should specify their target as a K8sPod with a defined pod spec")
 	}
 	var podSpec corev1.PodSpec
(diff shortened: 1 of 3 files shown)


pkg/manager/impl/validation/task_validator.go
```diff
@@ -44,6 +42,31 @@ func validateContainer(task core.TaskTemplate, taskConfig runtime.TaskResourceCo
 	return nil
 }
 
+// This is called for a task with a non-nil k8s pod.
+func validateK8sPod(task core.TaskTemplate, taskConfig runtime.TaskResourceConfiguration) error {
+	if task.GetK8SPod().PodSpec == nil {
+		return nil
```
katrogan, line 48: is this actually valid?

katrogan (dismissed):

pkg/manager/impl/validation/task_validator.go
pingsutw: Updated it, thanks

pkg/repositories/database_integration_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration
```
katrogan, line 1: why add this?

pingsutw: This is automatically generated by `goimports` (goimports -w pkg/*) 
Could we add this? So we don't need to manually delete it when running `goimports`

pingsutw: thank you

PR #374, opened 2022-03-19
Pr into pr: fix migrations for sqlite

katrogan:
no idea how rollback even works for sqlite since dropping columns appears to be [very convoluted](https://www.sqlitetutorial.net/sqlite-alter-table/)


kumare3:
I do not like this PR

katrogan:
@kumare3 do you dislike the uuid in place of autoincrement `id` or the sqlite vs postgres specific migrations?

katrogan:
@kumare3 do you dislike the uuid in place of autoincrement `id` or the sqlite vs postgres specific migrations?

PR #375, opened 2022-03-21
Record TaskExecutionMetadata from task events

hamersaw:
# TL;DR
Curently, the TaskExecutionMetadata is being provided by each task event but it isn't stored anywhere. This PR enables storing storing this information including merging.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2288

## Follow-up issue
_NA_


pkg/repositories/transformers/task_execution.go
```diff
@@ -240,6 +242,94 @@ func mergeCustom(existing, latest *_struct.Struct) (*_struct.Struct, error) {
...
+		return existing
+	}
+
+	for i, externalResource := range latest {
+		var index int
+		if externalResource.GetIndex() == 0 && externalResource.GetRetryAttempt() == 0 && externalResource.GetPhase() == 0 {
+			// updating an external resource to retryAttempt 0 and phase 0 is invalid, because that
+			// is the initial state, so we use this condition to signify an external resource
+			// without these parameters (ex. a simple ExternalResourceID).
+			index = i
```
katrogan, line 285: is there any chance another externalResource in latest might have already claimed index `i`? could you leave a comment for this method about the general assumption behind indices, like any ordering constraints for example?

pkg/repositories/transformers/task_execution.go
```diff
@@ -240,6 +242,94 @@ func mergeCustom(existing, latest *_struct.Struct) (*_struct.Struct, error) {
...
+func mergeMetadata(existing, latest *event.TaskExecutionMetadata) *event.TaskExecutionMetadata {
+	if existing == nil {
+		return latest
+	}
+
+	if latest == nil {
+		return existing
+	}
+
+	if latest.GeneratedName != "" && existing.GeneratedName != latest.GeneratedName {
```
katrogan, line 318: if I'm understanding correctly, latest can never explicitly unset any fields (i.e. set them to their zero value)?

hamersaw: I tried to implement logical merging with the ability to unset fields that make sense to unset. For example, `GeneratedName` is set to the task ID, `PluginIdentifier` is the plugin name, and `InstanceClass` is whether it is interruptible or not. To me, it doesn't make sense to allow null values for these, so in hopes to (potentially) reduce event construction complexity we allow updates to have nulls. I think it makes sense for some fields to be updatable to default values, and tried to allow it. For example the `Phase` in the mergeExternalResource function. I am not tied to this logic. If you, or anybody, feels strongly about it, I'm happy to update. Thoughts?

pkg/repositories/transformers/task_execution.go
hamersaw: I updated the comment. I think it makes sense, please let me know what you think.

katrogan: thank you!

pkg/repositories/transformers/task_execution.go
katrogan: this sounds reasonable, i think in other events we take the same approach (merging log links, etc) - thanks for explaining!

katrogan (approved):

hamersaw:
@katrogan per our conversation I updated this to maintain separate ExternalResourceInfo protos for each unique <Index, RetryAttempt> pair. This means that any plugin which reports multiple ExternalResourceInfo protos needs to set the Index field for them to be handled properly. I think this is a sane requirement.

katrogan (dismissed):

katrogan (approved):

PR #378, opened 2022-03-23
Added execution config changes

pmahindrakar-oss:

# TL;DR

#### Using only the  values.yaml changes 
```
      flyteadmin:
        roleNameKey: "iam.amazonaws.com/role"
        profilerPort: 10254
        metricsScope: "flyte:"
        labels:
          "FlyteLabelKey" : "FlyteLabelValue"
        annotations:
          "FlyteAnnotationKey" : "FlyteAnnotationValue"
        metadataStoragePrefix:
          - "metadata"
          - "admin"
        eventVersion: 2
        testing:
          host: http://flyteadmin
```

Launched pod  

```
 kubectl get pods  -n flytesnacks-development f5e21f3c6ec3048d8b80-n0-0 -o yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
    flyteannotationkey: FlyteAnnotationValue
  creationTimestamp: "2022-03-24T14:57:29Z"
  labels:
    domain: development
    execution-id: f5e21f3c6ec3048d8b80
    flytelabelkey: FlyteLabelValue
    interruptible: "false"
    node-id: n0
    project: flytesnacks
    shard-key: "15"
    task-name: core-flyte-basics-lp-greet
    workflow-name: core-flyte-basics-lp-go-greet
```

### With overriden project domain attributes for WORKFLOW_EXECUTION_CONFIG

```
lytectl get workflow-execution-config -p flytesnacks -d development
{
  "project": "flytesnacks",
  "domain": "development",
  "max_parallelism": 10,
  "security_context": {
    "run_as": {
      "k8s_service_account": "default"
    }
  },
  "raw_output_data_config": {
    "output_location_prefix": "cliOutputLocationPrefix"
  },
  "labels": {
    "values": {
      "cliLabelKey": "cliLabelValue"
    }
  },
  "annotations": {
    "values": {
      "cliAnnotationKeyNew": "cliAnnotationValue"
    }
  }
}%
```

```
kubectl get pods  -n flytesnacks-development fe1c87252016c4cf6942-n0-0 -o yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    cliAnnotationKeyNew: cliAnnotationValue
    cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
  creationTimestamp: "2022-03-24T14:58:37Z"
  labels:
    cliLabelKey: cliLabelValue
    domain: development
    execution-id: fe1c87252016c4cf6942
    interruptible: "false"
    node-id: n0
    project: flytesnacks
    shard-key: "8"
    task-name: core-flyte-basics-lp-greet
    workflow-name: core-flyte-basics-lp-go-greet
```

Worflow level attributes have a bug and hence couldn't be tested and also the current changes don't use workflow level execution config.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue

https://github.com/flyteorg/flyte/issues/2070

## Follow-up issue
_NA_


pkg/runtime/application_config_provider.go
```diff
@@ -38,6 +38,11 @@ var flyteAdminConfig = config.MustRegisterSection(flyteAdmin, &interfaces.Applic
 	EventVersion:          2,
 	AsyncEventsBufferSize: 100,
 	MaxParallelism:        25,
+	Labels:                map[string]string{"defaultFlyteLabelKey": "defaultFlyteLabelValue"},
+	Annotations:           map[string]string{"defaultFlyteAnnotationKey": "defaultFlyteAnnotationValue"},
+	OutputLocationPrefix:  "defaultFlyteOutputLocationPrefix",
+	K8SServiceAccount:     "default",
+	AssumableIamRole:      "default",
```
pmahindrakar-oss, line 45: Are we ok with these defaults @katrogan

pkg/manager/impl/execution_manager.go
```diff
@@ -428,41 +428,71 @@ func (m *ExecutionManager) getInheritedExecMetadata(ctx context.Context, request
 	return parentNodeExecutionID, sourceExecutionID, nil
 }
 
+// WorkflowExecutionConfigInterface is used as common interface for capturing the common behavior catering to the needs
+// of fetching the WorkflowExecutionConfig across LaunchPlanSpec, ExecutionCreateRequest
+// MatchableResource_WORKFLOW_EXECUTION_CONFIG and ApplicationConfig
+type WorkflowExecutionConfigInterface interface {
```
katrogan, line 434: why does this need to be an interface?

pkg/runtime/application_config_provider.go
katrogan: Hm, besides setting the `K8SServiceAccount` are the rest of these actually useful for a flyte execution? maybe we can omit the rest

pkg/manager/impl/execution_manager.go
pmahindrakar-oss: It becomes very convenient now to use the same merge function 

```mergeIntoExecConfig(workflowExecConfig *admin.WorkflowExecutionConfig, spec WorkflowExecutionConfigInterface)```

for launchplan spec, executionCreateRequest, matchable resorce and application config.
Without these we will have to define individual functions for each of these object types.

Abstracting this common functionality across all these make it easier to manage the code.
Let me know if you think otherwise

Comment (katrogan) [reply pkg/manager/impl/execution_manager.go:434]: I see, maybe one day https://github.com/flyteorg/flyte/issues/2242 will simplify this!

File: pkg/manager/impl/execution_manager.go [hunk 3]
```diff
@@ -428,41 +428,77 @@ func (m *ExecutionManager) getInheritedExecMetadata(ctx context.Context, request
...
+
+// Merge into workflowExecConfig from spec
+func mergeIntoExecConfig(workflowExecConfig *admin.WorkflowExecutionConfig, spec WorkflowExecutionConfigInterface) {
+	if workflowExecConfig.GetMaxParallelism() == 0 && spec.GetMaxParallelism() > 0 {
+		workflowExecConfig.MaxParallelism = spec.GetMaxParallelism()
+	}
+	if workflowExecConfig.GetSecurityContext() == nil && spec.GetSecurityContext() != nil {
+		workflowExecConfig.SecurityContext = spec.GetSecurityContext()
+	}
+	// Launchplan spec has label, annotation and rawOutputDataConfig initialized with empty values.
```
katrogan, line 455: ah, is this from flytekit serialization? thank you for documenting this

pkg/manager/impl/execution_manager.go
```diff
@@ -428,41 +428,77 @@ func (m *ExecutionManager) getInheritedExecMetadata(ctx context.Context, request
...
-		}, nil
-	}
-	if launchPlan != nil && launchPlan.Spec.MaxParallelism > 0 {
-		return &admin.WorkflowExecutionConfig{
-			MaxParallelism: launchPlan.Spec.MaxParallelism,
-		}, nil
+
+	workflowExecConfig := &admin.WorkflowExecutionConfig{}
+	// merge the request spec into workflowExecConfig
+	mergeIntoExecConfig(workflowExecConfig, request.Spec)
```
katrogan, line 481: why this call instead of initializing the workflowExecConfig struct inline with value from the request.Spec?

pkg/manager/impl/util/shared.go
```diff
@@ -233,3 +234,22 @@ func GetTaskExecutionModel(
 	}
 	return &taskExecutionModel, nil
 }
+
+// GetMatchableResource gets matchable resource for resourceType and project - domain combination.
+// Returns an error if such a resource is not found.
```
katrogan, line 239: Can you add that this returns nil if nothing is found but it won't error?

pkg/manager/impl/execution_manager.go
```diff
@@ -428,41 +428,77 @@ func (m *ExecutionManager) getInheritedExecMetadata(ctx context.Context, request
...
+		(spec.GetAnnotations() != nil && len(spec.GetAnnotations().Values) > 0) {
+		workflowExecConfig.Annotations = spec.GetAnnotations()
+	}
+}
+
 // Produces execution-time attributes for workflow execution.
 // Defaults to overridable execution values set in the execution create request, then looks at the launch plan values
 // (if any) before defaulting to values set in the matchable resource db and further if matchable resources don't
 // exist then defaults to one set in application configuration
 func (m *ExecutionManager) getExecutionConfig(ctx context.Context, request *admin.ExecutionCreateRequest,
```
katrogan, line 476: what if a user specifies some fields in the execution config in the request spec, but not all? 

I think we should then treat that execution config as finalized, and the omitted fields as intentionally excluded.

currently, we keep proceeding with the fallbacks regardless of whether the request spec is populated, or barring that, if the launch plan is then populated. i'm not sure this is right, what do you think? if one of these levels has any non-empty values then should we treat that execution config as the final execution config we apply?

pkg/manager/impl/util/shared_test.go
```diff
@@ -474,3 +476,64 @@ func TestListActiveLaunchPlanVersionsFilters(t *testing.T) {
...
+					Target: &admin.MatchingAttributes_WorkflowExecutionConfig{
+						WorkflowExecutionConfig: &admin.WorkflowExecutionConfig{
+							MaxParallelism: 12,
+						},
+					},
+				},
+			}, nil
+		}
+
+		_, err := GetMatchableResource(context.Background(), resourceManager, resourceType, project, domain)
```
katrogan, line 504: nit: also check that the expected resource is returned?

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -59,6 +61,19 @@ type ApplicationConfig struct {
...
 	// and parallelism/concurrency of MapTasks is independent from this.
 	MaxParallelism int32 `json:"maxParallelism"`
+	// Labels to apply to the execution resource.
+	Labels map[string]string `json:"labels,omitempty"`
+	// Annotations to apply to the execution resource.
+	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// Optional: security context override to apply this execution.
+	// iam_role references the fully qualified name of Identity & Access Management role to impersonate.
+	AssumableIamRole string `json:"assumableIamRole"`
```
katrogan, line 71: nit: add comments to pflag?

katrogan (commented):
mind updating the PR title before merging?

pkg/manager/impl/execution_manager.go
pmahindrakar-oss: I wanted this to look like how we do override for go variables and scopes actually define what that level defines the value for that variable
https://go.dev/play/p/jdkv14fHY8N

But we don't have a way to tell protobuf fields that they are set but those are explicitly being set to nil or empty.
i have seen where some libraries provide a behavior of tristate

i.e set (having value/ having no value) / unset(uninitialized)
Something like this tristateBoolean having isUndef https://docs.oracle.com/cd/E26098_01/apirefs.1112/e17493/oracle/ide/util/TriStateBoolean.html#isUndef__

That way it easier to distinguish what the user wanted to do. 

Also I think in the current state that issue comes even for maxParallelism = 0 in execution create request  and launch plan spec having the same value and  it will fall back to other overrides though the user intended to set maxParallelism = 0 in execution create Request 

Having a tristate behavior solves such cases and make it easier to reason across overrides and avoid using custom logic IMO.

To keep logic similar to previous one , i am adding a isChanged bool flag in merge function which is checked after the merge from execution create request and launchplan spec and if it gets set then we return that config object.

Let me know what you think.

pkg/manager/impl/execution_manager.go
pmahindrakar-oss: Two reasons
* It allows us to not make changes in merge aswell getExecutionConfig whenever a new field is introduced. 
* Since we now want old behavior where if any field is set in request spec then return that execution config immediately instead of checking for other overrides.

So i have kept the function with its return value checked. Let me know what you think

pkg/manager/impl/util/shared.go
pmahindrakar-oss: Done. misread the function

pkg/manager/impl/util/shared_test.go
pmahindrakar-oss: Done .

pkg/runtime/interfaces/application_configuration.go
pmahindrakar-oss: We are not using this as pflag config and none of the other fields in this config can be passed from command line.
If we want that behavior then i would need to add a pflag generator and add pflag annotation to all.
Let me know if we need that now or in any future PR's

pkg/manager/impl/execution_manager.go
pmahindrakar-oss: Yeah that would help simplify a lot. I couldn't find a way to do this with current protobuf and faced the same issue with another service we wrote where we had to duplicate the message fields .

pkg/manager/impl/execution_manager.go
pmahindrakar-oss: Yes that right
```
        labels=entity.labels or _common_models.Labels({}),
        annotations=entity.annotations or _common_models.Annotations({}),
        auth_role=entity._auth_role or _common_models.AuthRole(),
        raw_output_data_config=entity.raw_output_data_config or _common_models.RawOutputDataConfig(""),
```
https://github.com/flyteorg/flytekit/blob/b08e2b6d4cd42f6dc9be86bcec3044b3138e329c/flytekit/tools/translator.py#L224

go.mod
```diff
@@ -170,3 +170,5 @@ require (
 )
 
 replace github.com/robfig/cron/v3 => github.com/unionai/cron/v3 v3.0.2-0.20210825070134-bfc34418fe84
+
+replace github.com/flyteorg/flyteidl => github.com/flyteorg/flyteidl v0.24.4-0.20220324115508-8fb0189903bf
```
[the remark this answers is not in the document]
pmahindrakar-oss: Yes. updated to latest idl

pkg/runtime/interfaces/application_configuration.go
katrogan: ah okay, nevermind then

pkg/manager/impl/execution_manager.go
katrogan: re: the tri-state, we do something funky with one-ofs for notification overrides to permit a user to explicitly [disable notifications](https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/admin/execution.proto#L263) as opposed to interpreting the optional field as omitted. we could consider updating the execution config in the proto definitions to have some kind of bool flag for "is set" so that even an empty execution config is considered explicitly set

isChanged looks good for now too!

PR #382, opened 2022-03-29
Record IsParent and IsDynamic from propeller events

katrogan:
# TL;DR
In order to pre-empt the preloading of child node executions this PR adds a new "InternalData" field to the NodeExecution model. This contains a serialized version of an admin-internal proto which so far contains the original EventVersion tied to the node execution creation (via propeller event).

When a Get or List call returns a node execution with no InternalData or InternalData indicating event version `0`, then we issue the more expensive DB call with child node execution preloading. 

This approach slows down List calls for "older" type node executions preceding the accompanying flytepropeller change as a trade-off for improved performance going forward.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2282

## Follow-up issue
_NA_


pkg/repositories/transformers/node_execution.go
```diff
@@ -223,6 +234,31 @@ func UpdateNodeExecutionModel(
...
+				return errors.NewFlyteAdminErrorf(codes.Internal,
+					"failed to unmarshal node execution metadata with error: %+v", err)
+			}
+		}
+		if request.Event.IsParent {
+			nodeExecutionMetadata.IsParentNode = true
+		}
+		if request.Event.IsDynamic {
+			nodeExecutionMetadata.IsDynamic = true
+		}
```
pmahindrakar-oss, line 253:
```suggestion
		if request.Event.IsParent {
			nodeExecutionMetadata.IsParentNode = true
		}
		if request.Event.IsDynamic {
			nodeExecutionMetadata.IsDynamic = true
		}
```
```suggestion
		nodeExecutionMetadata.IsParentNode = request.Event.IsParent
                nodeExecutionMetadata.IsDynamic= request.Event.IsDynamic
```

pkg/repositories/config/migration_models.go
```diff
@@ -51,6 +51,8 @@ type NodeExecution struct {
 	LaunchedExecution models.Execution `gorm:"foreignKey:ParentNodeExecutionID;references:ID"`
 	// In the case of dynamic workflow nodes, the remote closure is uploaded to the path specified here.
 	DynamicWorkflowRemoteClosureReference string
+	// Metadata that is only relevant to the flyteadmin service that is used to parse the model and track additional attributes.
+	InternalData []byte
 }
```
pmahindrakar-oss, line 56: Just for my understanding , why do we have to maintains these models files only for migration purpose

pmahindrakar-oss, line 56: Just for my understanding , why do we have to maintains these models files only for migration purpose

katrogan: hey @pmahindrakar-oss does this comment at the top help? https://github.com/flyteorg/flyteadmin/blob/master/pkg/repositories/config/migration_models.go#L10

pkg/repositories/transformers/node_execution.go
katrogan: ah let me add a comment, not every event sends IsParent _and_ IsDynamic as an artifact of how propeller handles dynamic nodes. So i only want to explicitly set the fields, when they're set in the event itself.

pkg/repositories/config/migration_models.go
pmahindrakar-oss: Got it thanks for pointing those out.

pmahindrakar-oss (approved):

EngHabu (approved):
Thank you! looks good to me

EngHabu (dismissed):

EngHabu (approved):

PR #385, opened 2022-04-04
Fixed isScheduleEmpty function for flyte schedules

pmahindrakar-oss:

# TL;DR

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/2316

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/launch_plan_manager.go
+++ b/pkg/manager/impl/launch_plan_manager.go
@@ -142,6 +142,9 @@ func isScheduleEmpty(launchPlanSpec admin.LaunchPlanSpec) bool {
 	if schedule == nil {
 		return true
 	}
+	if schedule.GetCronSchedule() != nil && len(schedule.GetCronSchedule().Schedule) != 0 {
+		return false
+	}
 	if len(schedule.GetKickoffTimeInputArg()) != 0 {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


pkg/manager/impl/launch_plan_manager.go
```diff
@@ -142,6 +142,9 @@ func isScheduleEmpty(launchPlanSpec admin.LaunchPlanSpec) bool {
 	if schedule == nil {
 		return true
 	}
+	if schedule.GetCronSchedule() != nil && len(schedule.GetCronSchedule().Schedule) != 0 {
+		return false
+	}
 	if len(schedule.GetKickoffTimeInputArg()) != 0 {
 		return false
 	}
```
pmahindrakar-oss, line 150: @katrogan i didn't change this previous logic but trying to understand why do we have this check . We can't have a schedule with just this arg so wanted to understand the purpose of this check

katrogan: great question (and git blame is not very helpful here) I think we can safely assume the schedule expression oneof is not nil then that's all we need to check

katrogan (approved):

pkg/manager/impl/launch_plan_manager.go
pmahindrakar-oss: Ok cool.Merging this will remove that logic in followup

PR #392, opened 2022-04-07
Always register the metadata endpoint

EngHabu:
Signed-off-by: Haytham Abuelfutuh <haytham@afutuh.com>

# TL;DR
Expose a new property in Public Client metadata and always registers the metadata endpoint (even when auth is not configured)

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Tracking Issue


pkg/config/config.go
```diff
@@ -26,6 +26,8 @@ type ServerConfig struct {
 	DeprecatedThirdPartyConfig authConfig.ThirdPartyConfigOptions `json:"thirdPartyConfig" pflag:",Deprecated please use auth.appAuth.thirdPartyConfig instead."`
 
 	DataProxy DataProxyConfig `json:"dataProxy" pflag:",Defines data proxy configuration."`
+
+	ExportedDNS config.URL `json:"exportedDNS" pflag:",Defines the http endpoint the service is accessible at."`
```
EngHabu, line 30: What should I call this 😬, @wild-endeavor ?

kumare3: serviceEndpoint?

wild-endeavor: HTTP_URL? :)

wild-endeavor: Can we add a comment to explain under what circumstance this should be used?

pkg/server/service.go
```diff
@@ -107,8 +107,8 @@ func newGRPCServer(ctx context.Context, pluginRegistry *plugins.Registry, cfg *c
 
 	configuration := runtime2.NewConfigurationProvider()
 	service.RegisterAdminServiceServer(grpcServer, adminservice.NewAdminServer(ctx, pluginRegistry, configuration, cfg.KubeConfig, cfg.Master, dataStorageClient, scope.NewSubScope("admin")))
+	service.RegisterAuthMetadataServiceServer(grpcServer, authCtx.AuthMetadataService())
```
katrogan, line 110: how does this work if UseAuth isn't true?

katrogan (dismissed):

kumare3 (dismissed):

PR #395, opened 2022-04-07
Track EventVersion in TaskExecutionClosure

hamersaw:
# TL;DR
TaskExecutionEvents now include an event_version field to enable versioning of the data proto. This PR uses that field to track the EventVersion within the TaskExecutionClosure. The first use of this will be to track the update in subtask metadata reporting for map tasks.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
_NA_

## Follow-up issue
_NA_


pkg/repositories/transformers/task_execution.go
```diff
@@ -374,6 +375,9 @@ func UpdateTaskExecutionModel(ctx context.Context, request *admin.TaskExecutionE
 		return errors.NewFlyteAdminErrorf(codes.Internal, "failed to merge task event custom_info with error: %v", err)
 	}
 	taskExecutionClosure.Metadata = mergeMetadata(taskExecutionClosure.Metadata, request.Event.Metadata)
+	if request.Event.EventVersion > taskExecutionClosure.EventVersion {
```
katrogan, line 378: do we expect the task event version to change during task execution? is this more akin to [phase version](https://github.com/flyteorg/flyteidl/blob/8a5d301e1d855586b5d6e561e68cf0a9ab55b02b/protos/flyteidl/event/event.proto#L184)?

hamersaw: Great question, it is not like phase version which is continually incremented. I added this logic because propeller may be updated while a task is running. For example, in the current map task subtask metadata reporting case - if we update propeller during a run the first half will be reported as is (version 0), the second half will be reported with the updated approach (version 1). Both will be incomplete, but I think handling this completely will add lot of overly complex logic. I attempted to allow task event version to only be incremented to be a little more defensive. I noticed the node execution event version is only set during create - I'm OK going that route as well if we want?

katrogan (approved):

katrogan (approved):

PR #408, opened 2022-04-15
Deprecated the security context part of ExecutionParameters

pmahindrakar-oss:

# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/pkg/manager/impl/execution_manager_test.go
+++ b/pkg/manager/impl/execution_manager_test.go
@@ -4250,11 +4250,14 @@ func TestGetExecutionConfig(t *testing.T) {
 		resourceManager: &resourceManager,
 		config:          applicationConfig,
 	}
+	launchPlan := &admin.LaunchPlan{
+		Spec: &admin.LaunchPlanSpec{},
+	}
 	execConfig, err := executionManager.getExecutionConfig(context.TODO(), &admin.ExecutionCreateRequest{
 		Project: workflowIdentifier.Project,
 		Domain:  workflowIdentifier.Domain,
 		Spec:    &admin.ExecutionSpec{},
-	}, nil)
+	}, launchPlan)
 	assert.NoError(t, err)
 	assert.Equal(t, execConfig.MaxParallelism, int32(100))
 }
--- a/pkg/manager/impl/execution_manager_test.go
+++ b/pkg/manager/impl/execution_manager_test.go
@@ -3697,6 +3697,12 @@ func TestCreateSingleTaskExecution(t *testing.T) {
 
 func TestGetExecutionConfigOverrides(t *testing.T) {
 
+	deprecatedAuthRoleK8sServiceAccount := "deprecated auth role service account"
+	deprecatedAuthRoleAssumableIamRole := "deprecated auth role iam"
+
+	launchPlanDeprecatedAuthRoleK8sServiceAccount := "launch plan deprecated auth role service account"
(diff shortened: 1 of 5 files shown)


pkg/manager/impl/execution_manager_test.go
```diff
@@ -3847,6 +3922,43 @@ func TestGetExecutionConfigOverrides(t *testing.T) {
...
+						K8SServiceAccount: launchPlanK8sServiceAccount,
+					},
+				},
+				MaxParallelism: launchPlanMaxParallelism,
+			},
+		}
+		execConfig, err := executionManager.getExecutionConfig(context.TODO(), request, launchPlan)
+		assert.NoError(t, err)
+		assert.Equal(t, launchPlanMaxParallelism, execConfig.MaxParallelism)
+		assert.Equal(t, launchPlanK8sServiceAccount, execConfig.SecurityContext.RunAs.K8SServiceAccount)
```
katrogan, line 3958: is this correct? if we set the deprecated field in the CreateExecutionRequest shouldn't that take precedence over whatever is in the launch plan?

pkg/manager/impl/execution_manager_test.go
```diff
@@ -3922,7 +4063,29 @@ func TestGetExecutionConfigOverrides(t *testing.T) {
 		assert.Equal(t, rmLabels, execConfig.GetLabels().Values)
 		assert.Equal(t, rmAnnotations, execConfig.GetAnnotations().Values)
 	})
-	t.Run("matchable resource partial config", func(t *testing.T) {
+	t.Run("launchplan with no config except deprecated auth role but having matchable attribute", func(t *testing.T) {
```
katrogan, line 4066: also not sure about this case, if the launch plan has any overridable field set in the spec then shouldn't we get all the execution config attributes from the LP and not the matchable attribute?

pkg/manager/impl/execution_manager_test.go
pmahindrakar-oss: This gives security context the precedence if its sets on the launch plan.  

If its not set as in this test `request full config with deprecated auth role` then it uses the deprecated auth role.

Here the assumption is if user has migrated to launchplans with new way of passing security context then he should move to execution create requests also with the same or do we want to support this case and use the deprecated  fields from request

pkg/manager/impl/execution_manager_test.go
pmahindrakar-oss: Again the same arg here too i was thinking , 
If the user already is aware of using security context in one place then can we expect him to use in others too ?

pkg/manager/impl/execution_manager_test.go
katrogan: hm I see what you're saying wrt the user has migrated on their end from entirely using auth role to security ctx, i just don't know if we can safely make that assumption. perhaps i'm being too defensive. we could document this somewhere user-facing perhaps

katrogan (approved):

pmahindrakar-oss:
Holding off on this since the existing fix worked . Will revisit this
cc : @katrogan

PR #412, opened 2022-04-21
Proposal: Keep populating execution config attributes from fallbacks

katrogan:

# TL;DR
Keep populating execution config attributes from fallbacks

The [ExecutionConfig](https://github.com/flyteorg/flyteidl/blob/254280d131b89f7a7c5af07bfec163450a321add/protos/flyteidl/admin/matchable_resource.proto#L104,L119) object only exists as a collection at the MatchableAttribute layer.

But the specific fields it contains are configurable at the CreateExecutionRequest and LaunchPlan spec levels, _individually_.

If a user sets one of these fields in the CreateExecutionRequest and LaunchPlan spec we should proceed with fallback for other values. This way if say, SecurityCtx is defined at CreateExecutionRequest but RawOutputMetadataPrefix is in the LaunchPlan, the RawOutputMetadataPrefix should not be set to empty at the end of the resolution.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


EngHabu (dismissed):
How about fields within SecurityContext (for example)? should they get merged... merging is hard :D as you are well aware...

katrogan:
> How about fields within SecurityContext (for example)? should they get merged

Good question, I think it's okay to consider one of iam role or k8s service account being set as treating the entire message be the definitive value. What do you think, is this intuitive?

EngHabu (approved):

PR #417, opened 2022-04-29
Added grpc health probe binary for readiness and liveness check

pmahindrakar-oss:

# TL;DR

Checks for both GRPC service and HTTP service used for REST 

Tested this on sandbox .

Related PR in flyte https://github.com/flyteorg/flyte/pull/2431

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

https://github.com/flyteorg/flyte/issues/1698

## Follow-up issue
_NA_

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,14 +17,14 @@ RUN go mod download
 # COPY the rest of the source code
 COPY . /go/src/github.com/flyteorg/flyteadmin/
 
-# This 'linux_compile' target should compile binaries to the /artifacts directory
-# The main entrypoint should be compiled to /artifacts/flyteadmin
-RUN make linux_compile
-
 # Pull GRPC health probe binary for liveness and readiness checks
 RUN GRPC_HEALTH_PROBE_VERSION=v0.4.11 && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -13,13 +13,19 @@ WORKDIR /go/src/github.com/flyteorg/flyteadmin
...
+
 # COPY the rest of the source code
 COPY . /go/src/github.com/flyteorg/flyteadmin/
 
 # This 'linux_compile' target should compile binaries to the /artifacts directory
 # The main entrypoint should be compiled to /artifacts/flyteadmin
 RUN make linux_compile
 
+# Pull GRPC health probe binary for liveness and readiness checks
+RUN GRPC_HEALTH_PROBE_VERSION=v0.4.11 && \
```
katrogan, line 25: in terms of dockerfile layering, should this be before we run compile?

pmahindrakar-oss: Since we don't use this tool for flyteadmin compilation hence there is no issue with the ordering .

Any other reason you see where we would need this before

katrogan (approved):

Dockerfile
EngHabu: I think we should move this before line 8 to avoid rerunning it with every change in go.mod or the code... otherwise looks good to me

katrogan (approved):

PR #423, opened 2022-05-12
Implement signal service for human-in-the-loop workflow orchestration

hamersaw:
# TL;DR
Adds SignalService implementation to support manual signals to trigger GateNodes in workflows.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/208

## Follow-up issue
_NA_


pkg/repositories/models/signal.go
```diff
@@ -0,0 +1,15 @@
+package models
+
+// Signal primary key
+type SignalKey struct {
+	ExecutionKey
+	SignalID string `gorm:"primary_key;index" valid:"length(0|255)"`
```
katrogan, line 6: should the execution key be part of the primary key? if so, you might need to follow the [migration models](https://github.com/flyteorg/flyteadmin/blob/master/pkg/repositories/config/migration_models.go#L10,L12) approach b/c of unique index names

pkg/repositories/transformers/signal.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
+
+	"github.com/golang/protobuf/proto"
+
+	"google.golang.org/grpc/codes"
+)
+
+func CreateSignalModel(signalID *core.SignalIdentifier, signalType *core.LiteralType, signalValue *core.Literal) (models.Signal, error) {
+	signalModel := models.Signal{}
+	if signalID != nil {
```
katrogan, line 17: when would this ever be nil?

pkg/repositories/transformers/signal.go
```diff
@@ -0,0 +1,134 @@
...
+		signalKey := &signalModel.SignalKey
+		if signalID.ExecutionId != nil {
+			executionKey := &signalKey.ExecutionKey
+			if signalID.ExecutionId.Project != "" {
```
katrogan, line 21: nit, I think the preferred check is `len(signalId.ExecutionId.Project) > 0` from Russ Cox's [reply](https://groups.google.com/g/golang-nuts/c/7Ks1iq2s7FA)

pkg/repositories/models/signal.go
hamersaw: OK, so I think I understand this, but just want to clarify. With the currently definition the `signals` table looks like:
```
flyteadmin=# \d signals
                                          Table "public.signals"
      Column       |           Type           | Collation | Nullable |               Default
-------------------+--------------------------+-----------+----------+-------------------------------------
 id                | bigint                   |           | not null | nextval('signals_id_seq'::regclass)
 created_at        | timestamp with time zone |           |          |
 updated_at        | timestamp with time zone |           |          |
 deleted_at        | timestamp with time zone |           |          |
 execution_project | text                     |           | not null |
 execution_domain  | text                     |           | not null |
 execution_name    | text                     |           | not null |
 signal_id         | text                     |           | not null |
 type              | bytea                    |           | not null |
 value             | bytea                    |           |          |
Indexes:
    "signals_pkey" PRIMARY KEY, btree (execution_project, execution_domain, execution_name, signal_id)
    "idx_signals_deleted_at" btree (deleted_at)
    "idx_signals_id" btree (id)
    "idx_signals_signal_id" btree (signal_id)
```

So the ExecutionKey is part of the primary key (as it needs to be) and we have indexes on the signal primary key, deleted_at, id, and signal_id. This does not result in the index name overlap you describe because the [ExecutionKey](https://github.com/flyteorg/flyteadmin/blob/52b5e20c37bc17ef2fcbdc7d4de8b8a6c714f0dd/pkg/repositories/models/execution.go#L13-L17) currently does not specify any indexes over it's fields (ie. `Project`, `Domain`, `Name`). If it did, we would have to create a separate model to use just in the migrations that did not include those indexes. Is that correct?

pkg/manager/mocks/signal.go
```diff
@@ -0,0 +1,56 @@
+package mocks
+
+import (
+	"context"
+
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
+)
+
+type GetOrCreateSignalFunc func(ctx context.Context, request admin.SignalGetOrCreateRequest) (*admin.Signal, error)
```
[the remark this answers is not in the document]
hamersaw: Defining these manually did seem a little weird :smile:. I was trying to keep uniformity in the repo, were many of the other managers (and repositories / etc) implemented before mockery adoption and that's why we didn't use auto-generated mocks there?

pkg/repositories/transformers/signal.go
hamersaw: It should never be, a signal without an ID is invalid. Included it for completeness. Perhaps throwing an error would be better here?

pkg/repositories/transformers/signal.go
hamersaw: TIL thanks!

pkg/manager/mocks/signal.go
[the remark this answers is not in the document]
katrogan: yeah there's no good reason for hand-rolling these and I'm sorry you already went through the exercise of writing them! consistency with the other mocks is definitely low importance :)

pkg/repositories/transformers/signal.go
katrogan: sure that sounds good

pkg/repositories/models/signal.go
katrogan: ah okay awesome, i didn't realize that the pkey already had a deterministic & unique name. as long as you don't need an index on execution key then what you have now looks great!

>  If it did, we would have to create a separate model
yep, exactly!

katrogan (approved):

wild-endeavor (approved):

PR #429, opened 2022-05-20
Reduce default metric keys, make configurable

katrogan:

# TL;DR
 Reduces potential for admin OOM when storing labeled counters for metrics keys' cross product

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2521

## Follow-up issue
_NA_

### DIFF
--- a/cmd/entrypoints/serve.go
+++ b/cmd/entrypoints/serve.go
@@ -9,14 +9,11 @@ import (
 
 	_ "net/http/pprof" // Required to serve application.
 
-	"github.com/flyteorg/flyteadmin/pkg/common"
 	"github.com/flyteorg/flyteadmin/pkg/server"
 	"github.com/flyteorg/flytestdlib/logger"
 	"github.com/spf13/cobra"
 
 	runtimeConfig "github.com/flyteorg/flyteadmin/pkg/runtime"
-	"github.com/flyteorg/flytestdlib/contextutils"
-	"github.com/flyteorg/flytestdlib/promutils/labeled"
 )
 
 var pluginRegistryStore = plugins.NewAtomicRegistry(plugins.NewRegistry())
@@ -45,10 +42,4 @@ func init() {
 	// Command information
 	RootCmd.AddCommand(serveCmd)
 	RootCmd.AddCommand(secretsCmd)
-
-	// Set Keys
-	labeled.SetMetricKeys(contextutils.AppNameKey, contextutils.ProjectKey, contextutils.DomainKey,
-		contextutils.ExecIDKey, contextutils.WorkflowIDKey, contextutils.NodeIDKey, contextutils.TaskIDKey,
-		contextutils.TaskTypeKey, contextutils.PhaseKey, contextutils.LaunchPlanIDKey, common.ErrorKindKey, common.RuntimeTypeKey,
-		common.RuntimeVersionKey)
 }
--- a/pkg/runtime/application_config_provider.go
+++ b/pkg/runtime/application_config_provider.go
@@ -4,6 +4,7 @@ import (
 	"github.com/flyteorg/flyteadmin/pkg/common"
 	"github.com/flyteorg/flyteadmin/pkg/runtime/interfaces"
 	"github.com/flyteorg/flytestdlib/config"
+	"github.com/flyteorg/flytestdlib/contextutils"
 	"github.com/flyteorg/flytestdlib/database"
 )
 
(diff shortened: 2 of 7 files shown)


pkg/server/service.go
```diff
@@ -53,6 +58,16 @@ func Serve(ctx context.Context, pluginRegistry *plugins.Registry, additionalHand
 	return serveGatewayInsecure(ctx, pluginRegistry, serverConfig, authConfig.GetConfig(), storage.GetConfig(), additionalHandlers, adminScope)
 }
 
+func SetMetricKeys(appConfig *runtimeIfaces.ApplicationConfig) {
+	var keys = make([]contextutils.Key, 0, len(appConfig.MetricKeys))
+	for _, keyName := range appConfig.MetricKeys {
+		key := contextutils.Key(keyName)
+		keys = append(keys, key)
+	}
+	logger.Warningf(context.TODO(), "setting metrics keys to %+v", keys)
```
EngHabu, line 67: Why is this a warning?

pkg/server/service.go
```diff
@@ -53,6 +58,16 @@ func Serve(ctx context.Context, pluginRegistry *plugins.Registry, additionalHand
...
+	labeled.SetMetricKeys(keys...)
```
EngHabu, line 68: Can you check if keys is empty and not call the function?

pkg/server/service.go
```diff
@@ -44,6 +48,7 @@ var defaultCorsHeaders = []string{"Content-Type"}
 func Serve(ctx context.Context, pluginRegistry *plugins.Registry, additionalHandlers map[string]func(http.ResponseWriter, *http.Request)) error {
 	serverConfig := config.GetConfig()
 	configuration := runtime2.NewConfigurationProvider()
+	SetMetricKeys(configuration.ApplicationConfiguration().GetTopLevelConfig())
```
EngHabu, line 51: Will this cause an issue with single binary? or does it not call Serve? we will be calling SetMetric keys multiple times now..

pkg/runtime/application_config_provider.go
```diff
@@ -22,6 +23,7 @@ const MB = KB * KB
 var flyteAdminConfig = config.MustRegisterSection(flyteAdmin, &interfaces.ApplicationConfig{
 	ProfilerPort:          metricPort,
 	MetricsScope:          "flyte:",
+	MetricKeys:            []string{contextutils.ProjectKey.String(), contextutils.DomainKey.String()},
```
EngHabu, line 26: I think the default should have a few more, right? WorkflowID, taskID, phase... etc. should probably continue to be there... (to keep close to current behavior)... deployments that don't need this granularity can choose to override the config...

pkg/server/service.go
katrogan: it does call serve but currently sets the metric keys in an init() which i was going to update in a subsequent pr pulling in this admin release https://github.com/flyteorg/flyte/blob/3642ae82fea0d244a0bb3ece36838225ef2109ee/cmd/single/start.go#L189

pkg/runtime/application_config_provider.go
katrogan: ok, done

pkg/server/service.go
katrogan: sorry, downgraded

pkg/server/service.go
katrogan: done

pkg/server/service.go
EngHabu: What will you update the single binary to do?

katrogan: i'll update it to pick up the admin release and remove the above snippet setting labeled metric keys in init() and use the config instead

EngHabu (dismissed):

cmd/entrypoints/serve.go
```diff
@@ -36,6 +34,7 @@ var serveCmd = &cobra.Command{
 				logger.Panicf(ctx, "Failed to Start profiling and Metrics server. Error, %v", err)
 			}
 		}()
+		server.SetMetricKeys(cfg.ApplicationConfiguration().GetTopLevelConfig())
```
kumare3, line 37: I do not know if you can do this? I think its a onStartup Property

kumare3: it seems it can be run only once. With this now you will have to make sure no one ever invokes it. I like it though!

https://github.com/flyteorg/flytestdlib/blob/15cac7bec8470974050ed02343cda8191e5f14bc/promutils/labeled/keys.go#L26

katrogan: yeah exactly, it'll panic if we it's called >1x, but for now people shouldn't be calling it themselves except for weird cases like single binary, which i will update to set the labels via the config after this release

PR #439, opened 2022-06-06
Moving backward compat authrole code in one place and also removed redundant fields

pmahindrakar-oss:

# TL;DR
With the introduction of the default values in service account 
https://github.com/flyteorg/flyteadmin/blob/6e03bd4ac2ad3bb950e8fa1ed64f4bc5c6058bcd/pkg/runtime/application_config_provider.go#L34

And this logic 
https://github.com/flyteorg/flyteadmin/blob/6e03bd4ac2ad3bb950e8fa1ed64f4bc5c6058bcd/pkg/manager/impl/execution_manager.go#L788

Hence even if authRole  which is deprecated field was set in launchplan or execution request, it was being ignored
since this logic was being resolved after the application default values were added.

Also as part of this PR cleaned up redundant fields in ExecutionParameters for Security Context and fetching those from 

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/execution_manager.go
+++ b/pkg/manager/impl/execution_manager.go
@@ -511,6 +511,14 @@ func (m *ExecutionManager) getExecutionConfig(ctx context.Context, request *admi
 		if launchPlan.Spec.WorkflowId != nil {
 			workflowName = launchPlan.Spec.WorkflowId.Name
 		}
+		// Backward compatibility changes to get security context from auth role.
+		resolvedAuthRole := resolveAuthRole(request, launchPlan)
+		resolvedSecurityCtx := resolveSecurityCtx(ctx, workflowExecConfig.GetSecurityContext(), resolvedAuthRole)
+		if workflowExecConfig.GetSecurityContext() == nil &&
+			(len(resolvedSecurityCtx.GetRunAs().GetK8SServiceAccount()) > 0 ||
+				len(resolvedSecurityCtx.GetRunAs().GetIamRole()) > 0) {
+			workflowExecConfig.SecurityContext = resolvedSecurityCtx
+		}
 	}
 
 	matchableResource, err := util.GetMatchableResource(ctx, m.resourceManager,
@@ -524,6 +532,7 @@ func (m *ExecutionManager) getExecutionConfig(ctx context.Context, request *admi
 		workflowExecConfig = mergeIntoExecConfig(workflowExecConfig,
 			matchableResource.Attributes.GetWorkflowExecutionConfig())
 	}
+
 	//  merge the application config into workflowExecConfig
 	workflowExecConfig = mergeIntoExecConfig(workflowExecConfig, m.config.ApplicationConfiguration().GetTopLevelConfig())
 	logger.Infof(ctx, "getting the workflow execution config from application configuration")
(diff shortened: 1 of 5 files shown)


pkg/manager/impl/execution_manager.go
```diff
@@ -511,6 +511,14 @@ func (m *ExecutionManager) getExecutionConfig(ctx context.Context, request *admi
 		if launchPlan.Spec.WorkflowId != nil {
 			workflowName = launchPlan.Spec.WorkflowId.Name
 		}
+		// Backward compatibility changes to get security context from auth role.
```
katrogan, line 514: this is getting so hairy :( can you add a more detailed comment why this is necessary?

katrogan (dismissed):

convexquad:
@pmahindrakar-oss if it is possible for you to finish this PR and commit it soon, that would be great! I need this fix in order to complete the upgrade of my Flyte backend system (i.e. Flyte Admin, Propeller, etc).

However, I would rather have this fix be made correctly the first time - if you need more time to keep working on it or to keep testing, I would rather you do that than try to rush the fix.

convexquad:
@katrogan could you take this the rest of the way if for some reason @pmahindrakar-oss is not available to finish it?

pmahindrakar-oss:
@convexquad would update on this today.  Unit tests do cover a lot of the older code too and i will try simulate workflows in older deprecated fields.
Also would be helpful if your team starts migrating your older workflows to new security context as that is the model going forward. We don't have a post deprecation strategy yet but wanted to get your opinion to avoid such issues in the future.

Having deprecated code across releases increases the test matrix and higher maintainability cost. 
Would prefer to remove deprecated field after 3 monthly releases and communication in the release notes for the same.

So a field marked as deprecated would be mentioned in the release and post 3 releases this field would be removed.
This would help remove a lot of tech debt to be carried forward.

pkg/manager/impl/execution_manager.go
pmahindrakar-oss: Hi @katrogan added comments for these

pmahindrakar-oss:
I have tested this my changing flytectl code to send older authRole during registration of workflows and verified that the same authRole is being picked by the workflows during execution in both regular and scheduled workflow case.
@katrogan @convexquad please review when you get a chance.

katrogan (approved):

convexquad (commented):
👍 I think this change will work correctly for scheduled launch plans - they should pick up the correct k8s service account with this change.

However @pmahindrakar-oss just to be clear, this is technically not a backwards compatibility issue. Without this change, scheduled launch plans will only run with the "default" k8s service account even with the most recent flytectl, flytekit, Flyte Admin, etc. versions.

convexquad:
@katrogan could you merge this PR and release a new Flyte Admin version for it, I guess. I can put it through the checks in my test cluster so we should catch any secondary issues quickly at least.

katrogan:
hey @convexquad if you'd like I've built an image with these changes: ghcr.io/flyteorg/flyteadmin:v1.1.20-security-ctx that you can test with. that way we can catch any outstanding issues before merging this change in

convexquad:
> hey @convexquad if you'd like I've built an image with these changes: ghcr.io/flyteorg/flyteadmin:v1.1.20-security-ctx that you can test with. that way we can catch any outstanding issues before merging this change in

@katrogan that works, I will post a comment on this PR or in the Flyte Slack workspace once I am done testing

katrogan:
thank you @convexquad for taking the time to test out this change (and hopefully it works!🤞)

convexquad:
@katrogan I tested launching workflows with a scheduled launch plan; with `flytectl create execution` and from Flyte Console and I got the correct declared k8s service account in all three tests. Could you merge this and publish a release?

katrogan (approved):

pmahindrakar-oss:
@convexquad Most recent flytectl and flytekit and any newly registered launchplan scheduled or unscheduled will be registered with new security context field. This change will take care of any older registered launchplans which had authRole in them.

Thanks for checking on this testing out.

PR #442, opened 2022-06-08
Add notification validation

katrogan:
# TL;DR
Validate notifications include non-empty recipients and are triggered only on terminal phases

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2601

## Follow-up issue
_NA_


pmahindrakar-oss (approved):

pkg/manager/impl/validation/execution_validator.go
```diff
@@ -62,6 +62,11 @@ func ValidateExecutionRequest(ctx context.Context, request admin.ExecutionCreate
 	if err := validateLiteralMap(request.Inputs, shared.Inputs); err != nil {
 		return err
 	}
+	if request.Spec.GetNotifications() != nil {
+		if err := validateNotifications(request.Spec.GetNotifications().Notifications); err != nil {
+			return err
+		}
+	}
```
pmahindrakar-oss, line 69: We can have launchplan metadata also use NotificationList and that way we can move L65 if check and just pass the NotificationList from both the places.

pmahindrakar-oss: Sorry missed this yeah i meant we should probably have this consistent in protos too

PR #445, opened 2022-06-11
Show non-PgError errors when connecting to db

slai:
# TL;DR
When network (or other non-database) errors occur while connecting to the database, the error is masked by a panic due to a bug in unwrapping the error.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

When connecting to a database instance that doesn't exist, that error is currently masked due to an invalid assumption that all wrapped errors returned from gorm are of type `*pgconn.PgError`. When a network error occurs, the returned error is of type `*net.OpError`. This causes the following panic -

```
{"json":{"src":"base.go:73"},"level":"fatal","msg":"caught panic: interface conversion: error is *net.OpError, not *pgconn.PgError [goroutine 1 [running]:\nruntime/debug.Stack()\n\t/home/user/.local/go/src/runtime/debug/stack.go:24
...
```

... and doesn't show what the actual error is.

This PR fixes that by using the functions in the `errors` package to check if the error chain contains an error of type `*pgconn.PgError`, and if not, simply returns the error.

It also removes the ConnectError interface, as it is no longer needed - the `errors.As` function will automatically check the whole error chain and unwrap as necessary.

With this fix, the error becomes -

```
{"json":{"src":"base.go:82"},"level":"fatal","msg":"failed to connect to `host=localhost user=postgres database=flyte`: dial error (dial tcp 127.0.0.1:5432: connect: connection refused)","ts":"2022-06-11T20:48:28+01:00"}
```

## Tracking Issue
N/A

## Follow-up issue
N/A

### DIFF
--- a/pkg/repositories/database.go
+++ b/pkg/repositories/database.go
@@ -2,15 +2,14 @@ package repositories
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
-	"reflect"
 	"strings"
 
 	"github.com/flyteorg/flytestdlib/database"
 
-	repoErrors "github.com/flyteorg/flyteadmin/pkg/repositories/errors"
 	"gorm.io/driver/sqlite"
 
 	"github.com/flyteorg/flytestdlib/logger"
@@ -116,15 +115,7 @@ func createPostgresDbIfNotExists(ctx context.Context, gormConfig *gorm.Config, p
 		return gormDb, nil
 	}
 
-	// Check if its invalid db code error
-	cErr, ok := err.(repoErrors.ConnectError)
-	if !ok {
-		logger.Errorf(ctx, "Failed to cast error of type: %v, err: %v", reflect.TypeOf(err),
-			err)
-		return nil, err
-	}
-	pqError := cErr.Unwrap().(*pgconn.PgError)
-	if pqError.Code != pqInvalidDBCode {
+	if !isInvalidDBPgError(err) {
 		return nil, err
 	}
 
@@ -152,6 +143,17 @@ func createPostgresDbIfNotExists(ctx context.Context, gormConfig *gorm.Config, p
 	}
 	// Now try connecting to the db again
 	return gorm.Open(dialector, gormConfig)
+}
+
+func isInvalidDBPgError(err error) bool {
+	pgErr := &pgconn.PgError{}
+	if !errors.As(err, &pgErr) {
+		// err chain does not contain a pgconn.PgError
+		return false
+	}
+
+	// pgconn.PgError found in chain and set to pgErr
+	return pgErr.Code == pqInvalidDBCode
 }
 
 func setupDbConnectionPool(ctx context.Context, gormDb *gorm.DB, dbConfig *database.DbConfig) error {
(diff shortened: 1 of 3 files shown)


pkg/repositories/database.go
```diff
@@ -116,17 +115,17 @@ func createPostgresDbIfNotExists(ctx context.Context, gormConfig *gorm.Config, p
...
 	}
 
-	// Check if its invalid db code error
-	cErr, ok := err.(repoErrors.ConnectError)
-	if !ok {
-		logger.Errorf(ctx, "Failed to cast error of type: %v, err: %v", reflect.TypeOf(err),
-			err)
+	// if it is an invalid db code error, continue, otherwise return
+	pgErr := &pgconn.PgError{}
+	if !errors.As(err, &pgErr) {
```
pmahindrakar-oss, line 120: Nit: Can you add a few unit tests to cover the newly added lines

pmahindrakar-oss (dismissed):

pkg/repositories/database.go
slai: It's pretty hard to write unit tests without refactoring all of this, because this function potentially opens multiple connections to Postgres directly in its code.

To do this, I'm thinking of creating a DialectorBuilder function/struct that takes a DSN and returns a Dialector, and pass it to `createPostgresDbIfNotExists` as an arg. DialectorBuilder would be created in the `GetDB` function (or further up). I would also need to create a mock DialectorBuilder which will return a mock Dialector that returns the configured errors.

Alternatively, I could just extract these few error handling lines of code and write some tests for it, but that feels like breaking up the code just for the sake of writing tests for specific lines, when the whole function should be tested.

(Ideally, I don't think flyteadmin should even have the permission to `CREATE DATABASE` - that should be a db admin only permission. But this is a much bigger change across the board.) 

Looking at the existing tests, I don't think there's any coverage here at the moment.

@pmahindrakar-oss any thoughts or other ideas? Would you like me to refactor this to include unit tests?

pkg/repositories/database.go
```diff
@@ -112,20 +111,14 @@ func createPostgresDbIfNotExists(ctx context.Context, gormConfig *gorm.Config, p
 
 	dialector := postgres.Open(getPostgresDsn(ctx, pgConfig))
 	gormDb, err := gorm.Open(dialector, gormConfig)
-	if err == nil {
```
slai, line 115: I changed this so it checks if it is an error first. Even though the logic is the same as before, I feel checking `if err != nil` immediately after the function creating the error is safer/more natural because this is the general Go error checking pattern.

slai: Sorry, ignore this - I changed it back. On second thought, the quirk in flow this introduces means it isn't really much cleaner than the original code.

pkg/repositories/database.go
pmahindrakar-oss: Ok sounds good. Thanks @slai

pmahindrakar-oss (dismissed):

pmahindrakar-oss:
@slai  Can you take care of the DCO issue.

pmahindrakar-oss (approved):

PR #452, opened 2022-06-30
Doc hub proposal

pingsutw:
# TL;DR
[Design Doc](https://docs.google.com/document/d/1wvCvM6pLQ17Zr0DcpS9es2vOFQsNeOaNN99ufGw8O1U/edit#heading=h.w5wnz0kbiild)

### Database Schema

**Task Table**
| Name  | Type |
| ------------- | ------------- |
| ... | ...|
| descriptionID | uint32|

**Description Table**
| Name  | Type |
| ------------- | ------------- |
| resource | string |
| project | string |
| domain | string |
| name | string |
| version | string |
| descriptionID | uint32|
| Digest | []byte |
| ShortDescription | string |
| LongDescription | []byte |
| link | string |

### Create a task/workflow description
we have added description_entity to [taskSpec](https://github.com/flyteorg/flyteidl/pull/303/files#diff-36ddf7e3b2379abbd02338b347c95dd8c15f504c448a6cdab1e2b52310f5cfffR55-R57) / [workflowSpec](https://github.com/flyteorg/flyteidl/pull/303/files#diff-c5c7d33a1cfa2e5a60dc2f8bfb2ef4daa230503c69b5f17215892239f06b7c94R59-R61), so we'll register a description entity when we register a task / workflow.

1. Check if we have the same description entity.
```SQL
Select * from Description WHERE resource = task AND project = flytesnacks AND domain = development AND NAME = tf_wf AND version = xyz
```
2. If the same description entity doesn't exist, then insert a new row to the Description Table. 
3. update descriptionID in the specific row in the task table.

### Get task description
Use descriptionID in the task table to find the specific description entity
```SQL
Select description.* LEFT JOIN task ON description.descriptionID = task.descriptionID AND description.project = task.project AND description.domain = task.domain AND description.name = task.name WHERE task.project = flytesnacks AND task.domain = development AND NAME = task.tf_wf AND task.version = a123456
```

### List task description
```SQL
Select description.* LEFT JOIN task ON description.descriptionID = task.descriptionID AND description.project = task.project AND description.domain = task.domain AND description.name = task.name WHERE task.project = flytesnacks AND task.domain = development AND NAME = task.tf_wf LIMIT 20
```

### Get / List the Task
Added short description to [task](https://github.com/flyteorg/flyteidl/pull/303/files#diff-36ddf7e3b2379abbd02338b347c95dd8c15f504c448a6cdab1e2b52310f5cfffR38-R40) proto, so we'll get short description when we get / list the task.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/531

## Follow-up issue
_NA_


pkg/manager/impl/task_manager.go
```diff
@@ -120,6 +121,15 @@ func (t *TaskManager) CreateTask(
 			contextWithRuntimeMeta, common.RuntimeVersionKey, finalizedRequest.Spec.Template.Metadata.Runtime.Version)
 		t.metrics.Registered.Inc(contextWithRuntimeMeta)
 	}
+
+	if request.Spec.DescriptionEntity != nil {
+		err = createDescriptionEntity(ctx, t.db, request.Spec.DescriptionEntity, *request.Id)
```
wild-endeavor, line 126: not atomic though... so we create the task first, then we try to create the description entity, and then we update the task/workflow with the description id.  i'd feel better if we can do this in one database transaction instead of three.

Also I'm a little confused here about what the flow is... for large descriptions, how/when does that get uploaded?  Is that something done client side by calling the data proxy for a signed url to upload?  I thought that offloading was going to happen in Admin and I'm not seeing it but I could be missing something, or maybe i'm remembering incorrectly.  how did we say the offloading process would work?

wild-endeavor (commented):
confused at a high level at how things are supposed to happen.  left some questions.

pkg/manager/impl/description_entity_manager.go
```diff
@@ -0,0 +1,181 @@
...
+			},
+			DescriptionID: descriptionID,
+		})
+		if err != nil {
+			logger.Errorf(ctx, "Failed to update descriptionID in workflows table: %v", err)
+			return err
+		}
+	}
+
+	return nil
```
katrogan, line 99: do we need to offload the long description somewhere in this method body?

pkg/repositories/transformers/workflow.go
```diff
@@ -53,6 +53,7 @@ func FromWorkflowModel(workflowModel models.Workflow) (admin.Workflow, error) {
 		Closure: &admin.WorkflowClosure{
 			CreatedAt: createdAt,
 		},
+		ShortDescription: workflowModel.ShortDescription,
```
katrogan, line 56: I think we're missing writing the short description to the workflow model in the first place.

pkg/manager/impl/description_entity_manager.go
```diff
@@ -0,0 +1,181 @@
...
+	return util.GetDescriptionEntity(ctx, d.db, *request.Id)
+}
+
+func (d *DescriptionEntityManager) ListDescriptionEntity(ctx context.Context, request admin.DescriptionEntityListRequest) (*admin.DescriptionEntityList, error) {
+	// Check required fields
+	if err := validation.ValidateDescriptionEntityListRequest(request); err != nil {
+		return nil, err
+	}
+	ctx = contextutils.WithProjectDomain(ctx, request.DescriptionEntityId.Project, request.DescriptionEntityId.Domain)
+	ctx = contextutils.WithWorkflowID(ctx, request.DescriptionEntityId.Name)
```
katrogan, line 118: this isn't necessarily always the workflow name though if it's for a different entity type, right?

pkg/repositories/transformers/description_entity.go
```diff
@@ -0,0 +1,86 @@
...
+		ShortDescription: descriptionEntity.ShortDescription,
+		LongDescription:  longDescriptionBytes,
+		SourceCode:       sourceCode,
+	}, nil
+}
+
+func FromDescriptionEntityModel(descriptionEntityModel models.DescriptionEntity) (*admin.DescriptionEntity, error) {
+
+	longDescription := admin.LongDescription{}
+	err := proto.Unmarshal(descriptionEntityModel.LongDescription, &longDescription)
```
katrogan, line 64: I see we're not offloading this? Also, do we always want to include this in list requests - that could get very bloated and might not be necessary

pkg/manager/impl/task_manager.go
```diff
@@ -120,6 +121,15 @@ func (t *TaskManager) CreateTask(
...
+		if err != nil {
```
katrogan, line 127: if the error is already exists, that's okay right?

pkg/repositories/models/description_entity.go
```diff
@@ -0,0 +1,32 @@
+package models
+
+import "github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
+
+// DescriptionEntityKey DescriptionEntity primary key
+type DescriptionEntityKey struct {
+	ResourceType core.ResourceType `gorm:"primary_key;index:description_entity_project_domain_name_version_idx" valid:"length(0|255)"`
```
katrogan, line 7: is this key for the reference entity (workflow, task) etc?

pkg/manager/impl/task_manager.go
katrogan: if we're not making this atomic then can we consider making a separate endpoint for registering description entities?

pkg/repositories/models/workflow.go
```diff
@@ -15,5 +15,8 @@ type Workflow struct {
 	TypedInterface          []byte
 	RemoteClosureIdentifier string `gorm:"not null" valid:"length(0|255)"`
 	// Hash of the compiled workflow closure
-	Digest []byte
+	Digest        []byte
+	DescriptionID uint `gorm:"index"`
+	// Read only and will not create this column in the task table
```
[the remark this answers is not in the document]
pingsutw: updated it. tl;dr, we join the description entity and task/workflow table when getting the task/workflow. To unmarshal the `short_description` in the description entity table to the workflow object, we have to add `short_description` here.

https://github.com/flyteorg/flyteadmin/blob/b05131fc301475a6c89680e7f1501ab2fa273d9b/pkg/repositories/gormimpl/workflow_repo.go#L102-L105

pkg/manager/impl/task_manager.go
pingsutw: made this atomic

pkg/manager/impl/description_entity_manager.go
pingsutw: we offload the long description in the flytekit side now. Let me know if it's better to do it in flyteadmin.

pkg/manager/impl/description_entity_manager.go
pingsutw: fixed it

pkg/manager/impl/task_manager.go
pingsutw: removed it. Now, creating tasks, creating description entity, and updating description ID are done in one DB transaction.

pingsutw:
cc @katrogan mind taking a look

pkg/repositories/transformers/description_entity.go
```diff
@@ -0,0 +1,94 @@
...
+	return &models.DescriptionEntity{
+		DescriptionEntityKey: models.DescriptionEntityKey{
+			ResourceType: id.ResourceType,
+			Project:      id.Project,
+			Domain:       id.Domain,
+			Name:         id.Name,
+			Version:      id.Version,
+		},
+		ShortDescription: descriptionEntity.ShortDescription,
+		LongDescription:  longDescriptionBytes,
```
wild-endeavor, line 57: in an earlier version of this code, the long description was in S3 or a blob store of some kind right?  it now will store everything in the db?

wild-endeavor:
and should that left join be a right join katrina?

pkg/manager/impl/description_entity_manager.go
katrogan: So any flyteadmin service client will have to make sure they similarly offload descriptions? This seems brittle, especially if we have to worry about read access being delegated to client code but have admin responsible for serving up the descriptions to any caller

pkg/repositories/models/workflow.go
```diff
@@ -15,5 +15,10 @@ type Workflow struct {
 	TypedInterface          []byte
 	RemoteClosureIdentifier string `gorm:"not null" valid:"length(0|255)"`
 	// Hash of the compiled workflow closure
-	Digest []byte
+	Digest        []byte
+	DescriptionID uint `gorm:"index"`
```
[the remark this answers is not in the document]
katrogan: Can we use gorm conventions to make this relational? https://gorm.io/docs/has_one.html

pkg/repositories/models/workflow.go
```diff
@@ -15,5 +15,10 @@ type Workflow struct {
...
+	// ShortDescription is saved in the description entity table. set this to read only so we won't create this column.
+	// Adding ShortDescription because we want to unmarshal the short description in the
+	// descriptionEntity table to workflow object.
+	ShortDescription string `gorm:"<-:false"`
```
katrogan, line 23: Because this is stored in the descriptions table, we'll have to do a left join now on every GetTask and GetWorkflow to populate the short description? Any reason not to store it inline?

pkg/repositories/gormimpl/task_repo.go
```diff
@@ -21,14 +22,32 @@ type TaskRepo struct {
...
+			if tx.Error != nil {
+				return r.errorTransformer.ToFlyteAdminError(tx.Error)
+			}
+			return nil
+		}
+		tx := r.db.Omit("id").Create(descriptionEntity)
+		if tx.Error != nil {
+			return r.errorTransformer.ToFlyteAdminError(tx.Error)
+		}
+		r.db.Last(descriptionEntity)
```
katrogan, line 39: you can use [FirstOrCreate](https://gorm.io/docs/advanced_query.html#FirstOrCreate) when you create the description entity to get is db id and avoid this lookup

pkg/repositories/gormimpl/description_entity_repo.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// DescriptionEntityRepo Implementation of DescriptionEntityRepoInterface.
+type DescriptionEntityRepo struct {
+	db               *gorm.DB
+	errorTransformer flyteAdminDbErrors.ErrorTransformer
+	metrics          gormMetrics
+}
+
+func (r *DescriptionEntityRepo) Create(ctx context.Context, input models.DescriptionEntity) (uint, error) {
```
katrogan, line 23: who calls this since you always Create as a side effect of workflow and task registration?

pkg/repositories/transformers/description_entity.go
katrogan: What are the length restrictions on the long description?

pkg/repositories/gormimpl/description_entity_repo.go
pingsutw: sorry, I forgot to remove this. removed it.

pkg/repositories/models/workflow.go
pingsutw: updated, store it inline now

pkg/repositories/transformers/description_entity.go
pingsutw: 4KB, flytekit will offload the long description if it exceeds 4KB when registering

pingsutw:
updating the tests

pingsutw:
some updates:
- Add more tests
- Store short description in task and workflow table, so we don't to join description entity table to get the short description.
- Remove descriptionID from task and workflow table because we don't need to join description entity table.

wild-endeavor (approved):

PR #458, opened 2022-07-19
offload the WorkflowClosure to storage for flytepropeller 

ckiosidis:
Depends on:
https://github.com/flyteorg/flytepropeller/pull/459

Reduce the size of the FlyteWorkflow CRD by offloading the workflowClosure to storage. The wfClosure will be picked up by propeller and the crd will be put together during the reconcilation loop.

- Adds an offloadWorkflowClosureToStorage config. 

When enabled flyteadmin would push parts of the CRD to storage and set the location of the file on the WorkflowClosureReference field of the CRD https://github.com/flyteorg/flytepropeller/blob/a969c7fde3804d713f89d42deab50114dbaa1bf3/pkg/apis/flyteworkflow/v1alpha1/workflow.go#L71

- Flytepropeller should be able to handle both CRD schemas based on if the WorkflowClosureReference is populated or not.
- Added the offloading right before pushing the CRD to kubernetes and cleared the unnecessary fields.

Depends on https://github.com/flyteorg/flytepropeller/pull/459 were the FlyteWorkflow CRD is modified

Signed-off-by: Babis Kiosidis <babisk@spotify.com>

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Tracking Issue

https://github.com/flyteorg/flyte/issues/2705

## Follow-up issue
_NA_
OR


ckiosidis:
This PR will fail because it contains some CRD changes that haven't been merged yet

ckiosidis:
cc @hamersaw

kumare3:
Are we using the workflow closure?

ckiosidis:
@kumare3 the PR skips the WorkflowClosure and grabs the fields straight from the *v1alpha1.FlyteWorkflow
It depends on changes https://github.com/flyteorg/flytepropeller/pull/459/files#diff-9e997ba8ae0dff4e900a2f8e6b40e3b15a5c1e45394ab1b111528ef3118a0df9R354-R359

hamersaw (dismissed):

pkg/workflowengine/impl/k8s_executor_test.go
```diff
@@ -224,9 +242,17 @@ func TestExecute_MiscError(t *testing.T) {
 		assert.Equal(t, namespace, ns)
 		return &fakeFlyteWorkflow
 	}
+
+	mockApplicationConfig := runtimeMocks.MockApplicationProvider{}
+	mockApplicationConfig.SetTopLevelConfig(runtimeInterfaces.ApplicationConfig{
+		OffloadWorkflowClosureToStorage: false,
```
pmahindrakar-oss, line 248: Can we have a unit test covering the true case and asserting on workflowspec, subworkflows and task to be nil.

flyteadmin_config.yaml
```diff
@@ -60,6 +60,7 @@ flyteadmin:
   metadataStoragePrefix:
     - "metadata"
     - "admin"
+  offloadWorkflowClosureToStorage: false
```
pmahindrakar-oss, line 63: Can we also add this flag in the values file with a comment on it usage

pmahindrakar-oss (dismissed):

flyteadmin_config.yaml
pmahindrakar-oss: Also a general question , since this is an application wide setting, 
Should we also be able to do this at the creation of execution level through a request spec

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -87,6 +87,9 @@ type ApplicationConfig struct {
 
 	// Prefix for where offloaded data from user workflows will be written
 	OutputLocationPrefix string `json:"outputLocationPrefix"`
+
+	// Enabling will use Storage (s3/gcs/etc) to offload static parts of CRDs.
+	OffloadWorkflowClosureToStorage bool `json:"offloadWorkflowClosureToStorage"`
```
kumare3, line 92:
```suggestion
	UseOffloadedWorkflowClosure bool `json:"offloadWorkflowClosureToStorage"`
```

pkg/runtime/application_config_provider.go
```diff
@@ -27,11 +27,12 @@ var flyteAdminConfig = config.MustRegisterSection(flyteAdmin, &interfaces.Applic
...
-	EventVersion:          2,
-	AsyncEventsBufferSize: 100,
-	MaxParallelism:        25,
-	K8SServiceAccount:     "",
+	MetadataStoragePrefix:           []string{"metadata", "admin"},
+	EventVersion:                    2,
+	AsyncEventsBufferSize:           100,
+	MaxParallelism:                  25,
+	K8SServiceAccount:               "",
+	UseOffloadedWorkflowClosure: false,
```
hamersaw, line 35: can we fix lint issue here?

pkg/manager/impl/execution_manager.go
```diff
@@ -845,12 +846,24 @@ func (m *ExecutionManager) launchExecutionAndPrepareModel(
 		return nil, nil, err
 	}
 
-	workflow, err := util.GetWorkflow(ctx, m.db, m.storageClient, *launchPlan.Spec.WorkflowId)
-
+	workflowModel, err := util.GetWorkflowModel(ctx, m.db, *launchPlan.Spec.WorkflowId)
```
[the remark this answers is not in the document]
hamersaw: @katrogan yeah, we need the `workflowModel` to get the `RemoteClosureIdentifier` so basically just inlined the function. Unless we want to change the util.GetWorkflow to return both the workflow and the model? Happy to change this up to better fit with the existing code.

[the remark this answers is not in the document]
katrogan: oh I see, maybe it's worth pulling out the 

```
	workflow, err := transformers.FromWorkflowModel(workflowModel)
	if err != nil {
		return nil, err
	}
	closure, err := FetchAndGetWorkflowClosure(ctx, store, workflowModel.RemoteClosureIdentifier)
	if err != nil {
		return nil, err
	}
	closure.CreatedAt = workflow.Closure.CreatedAt
	workflow.Closure = closure
```

block into a separate method that util.GetWorkflow and this thing can both call but I don't feel too strongly about it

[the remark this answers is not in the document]
ckiosidis: maybe it can come separately? Struggling a bit with merge conflicts and rebases in this pr

hamersaw (approved):

pkg/manager/impl/execution_manager.go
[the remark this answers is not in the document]
katrogan: sure we can save it for a follow-up

katrogan (approved):

PR #466, opened 2022-08-30
Increase `allowedExecutionNameLength` to 63

rahul-theorem:
Signed-off-by: Rahul Mehta <98349643+rahul-theorem@users.noreply.github.com>

# TL;DR
Increase `allowedExecutionNameLength` from 20 to 63 to allow more legible human-readable execution IDs

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
See title/summary

## Tracking Issue
Closes https://github.com/flyteorg/flyte/issues/2824

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/validation/execution_validator.go
+++ b/pkg/manager/impl/validation/execution_validator.go
@@ -16,7 +16,8 @@ import (
 	"google.golang.org/grpc/codes"
 )
 
-const allowedExecutionNameLength = 20
+// Maximum value length of a Kubernetes label
+const allowedExecutionNameLength = 63
 
 var executionIDRegex = regexp.MustCompile(`^[a-z][a-z\-0-9]*$`)
 
--- a/pkg/manager/impl/validation/execution_validator_test.go
+++ b/pkg/manager/impl/validation/execution_validator_test.go
@@ -162,9 +162,9 @@ func TestValidExecutionId(t *testing.T) {
 }
 
 func TestValidExecutionIdInvalidLength(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/manager/impl/validation/execution_validator.go
```diff
@@ -16,7 +16,8 @@ import (
 	"google.golang.org/grpc/codes"
 )
 
-const allowedExecutionNameLength = 20
+// Maximum value length of a Kubernetes label
+const allowedExecutionNameLength = 63
```
katrogan, line 20: should we also update the name length for [generated names](https://github.com/flyteorg/flyteadmin/blob/master/pkg/common/executions.go#L11)?

rahul-theorem: Defer to other maintainers, seems OK to leave the randomly generated string short (since a longer name might make the frontend less legible). If it's preferable to have both be consistent, I'm happy to make the change (cc @kumare3 @eapolinario)

katrogan: sure, just wanted to check if the omission was intentional :)
I think the chance of collisions for the 20 char generated name isn't a concern, so keeping as is sounds good 👍

rahul-theorem:
Just fixed the failing test, @katrogan would it be possible to re-run the CI checks (looks like it's required since this is my first PR to flyteadmin)?

katrogan (approved):

PR #477, opened 2022-09-16
#minor Add default execution cluster option for a multi-cluster setup

Antaxify:
# TL;DR
Add option to configure a defaultExecutionLabel as fallback when using multiple execution clusters.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

In a Flyte multi-cluster setup, we can control which cluster is used by setting "execution cluster labels" on projects/workflows. However, if an execution does not match a any label, a cluster is randomly chosen from all enabled clusters. This is not ideal if one of the clusters is in any way specialized and may only execute certain workloads.

The idea is to extend the clusters config by adding a defaultExecutionLabel as shown in this example below. If no execution label matches for an execution, the configured defaultExecutionLabel is used. This way, we can easily configure one or more default execution clusters and even apply weights on them.

This feature would be purely opt-in: If defaultExecutionLabel is not set, the original behavior of selecting a cluster at random is preserved.

```
clusters:
  defaultExecutionLabel: defaultclusters
  labelClusterMap:
    defaultclusters:
      - id: defaultcluster1
        weight: 0.5
      - id: defaultcluster2
        weight: 0.5
    specializedcluster:
      - id: specializedcluster
        weight: 1
  ...
```

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2882

## Follow-up issue
_NA_

### DIFF
--- a/pkg/executioncluster/impl/random_cluster_selector.go
+++ b/pkg/executioncluster/impl/random_cluster_selector.go
@@ -32,6 +32,7 @@ type RandomClusterSelector struct {
 	equalWeightedAllClusters random.WeightedRandomList
 	labelWeightedRandomMap   map[string]random.WeightedRandomList
 	resourceManager          managerInterfaces.ResourceInterface
+	defaultExecutionLabel    string
 }
 
 func getRandSource(seed string) (rand.Source, error) {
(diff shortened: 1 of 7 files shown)


hamersaw:
I think this looks reasonable cc @katrogan @pmahindrakar-oss thoughts?

Antaxify:
@hamersaw I added the DCO signoff right after opening the PR, without it, CI is not starting afaik.

Antaxify:
@hamersaw added the DCO

katrogan (approved):

PR #478, opened 2022-09-23
Remove default config

wild-endeavor:
# TL;DR
Removing unnecessary configuration.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Just clean up.

### DIFF
--- a/auth/config/config.go
+++ b/auth/config/config.go
@@ -111,6 +111,14 @@ var (
 						Scopes:        []string{"all", "offline", "access_token"},
 						Public:        true,
 					},
+					"flytepropeller": {
+						ID:            "flytepropeller",
+						Secret:        []byte(`$2a$06$xAAvtAwLXh4TkB7KrkSyLuakqJjH6lOXyVd6iPQcuvuyYdeT7C.sC`), // Change this.
+						RedirectURIs:  []string{"http://localhost:3846/callback"},
+						ResponseTypes: []string{"token"},
+						GrantTypes:    []string{"refresh_token", "client_credentials"},
+						Scopes:        []string{"all", "offline", "access_token"},
+					},
 				},
 			},
 		},
--- a/auth/config/config.go
+++ b/auth/config/config.go
@@ -113,7 +113,7 @@ var (
 					},
 					"flytepropeller": {
 						ID:            "flytepropeller",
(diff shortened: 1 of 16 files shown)


auth/config/config.go
```diff
@@ -111,14 +111,6 @@ var (
...
 						Public:        true,
 					},
-					"flytepropeller": {
-						ID:            "flytepropeller",
-						Secret:        []byte(`$2a$06$pxs1AkG81Kvrhpml1QiLSOQaTk9eePrU/7Yab9y07h3x0TglbaoT6`), // = "foobar"
-						RedirectURIs:  []string{"http://localhost:3846/callback"},
-						ResponseTypes: []string{"token"},
-						GrantTypes:    []string{"refresh_token", "client_credentials"},
-						Scopes:        []string{"all", "offline", "access_token"},
-					},
```
pmahindrakar-oss, line 121: This would be  default config used by selfAuthServer for creating the in memory clients, though not a recommended for production usage but has been there for quick try out. 
Other option would be to override this in the configMap which would require update to the helm chart. 
Any reason we are removing this.

pmahindrakar-oss (approved):

auth/config/config.go
pmahindrakar-oss: synced up offline

auth/config/config.go
```diff
@@ -111,14 +111,6 @@ var (
 						Scopes:        []string{"all", "offline", "access_token"},
 						Public:        true,
 					},
-					"flytepropeller": {
-						ID:            "flytepropeller",
-						Secret:        []byte(`$2a$06$pxs1AkG81Kvrhpml1QiLSOQaTk9eePrU/7Yab9y07h3x0TglbaoT6`), // = "foobar"
```
EngHabu, line 116: Should we leave the settings for flytepropeller and only remove the secret (if that's a valid config)

wild-endeavor: I don't think so?  I feel like I can see a future version where then an empty password is somehow valid.

EngHabu: Maybe generate a random long string and hash it and put the hash there and never store the string anywhere? I just don't want the default to be "wrong"... maybe just inaccessible...

EngHabu (approved):

EngHabu (approved):

PR #482, opened 2022-10-05
Skipping of cached task outputs via execution config

MorpheusXAUT:
# TL;DR
flyteadmin sets the `SkipCache` flag in the execution config to instruct flytepropeller to ignore cached data for an execution.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added
 - [X] Any pending items have an associated Issue

## Complete description
Similar to the `Interruptible` override, a new `SkipCache` execution config flag was introduced. flyteadmin ensures the flag is passed onto flytepropeller accordingly for `CreateExecution` and `RelaunchExecution` requests.

As this uses currently unmerged versions of [`flyteidl`](https://github.com/flyteorg/flyteidl/pull/318), [`flyteplugins`](https://github.com/flyteorg/flyteplugins/pull/285), [`flytestdlib`](https://github.com/flyteorg/flytestdlib/pull/141) and [`flytepropeller`](https://github.com/flyteorg/flytepropeller/pull/489), the PR is created as a draft (for review) until the respective new versions are available.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2867

## Follow-up issue
_NA_


pkg/runtime/interfaces/application_configuration.go
```diff
@@ -78,6 +78,11 @@ type ApplicationConfig struct {
 	Annotations map[string]string `json:"annotations,omitempty"`
 	// Interruptible indicates whether all tasks should be run as interruptible by default (unless specified otherwise via the execution/workflow/task definition)
 	Interruptible bool `json:"interruptible"`
+	// SkipCache indicates all workflows and tasks should skip all their cached results and re-compute their outputs,
+	// overwriting any already stored data.
+	// Note that setting this setting to `true` effectively disabled all caching in Flyte as all executions launched
+	// will have their SkipCache setting enabled.
+	SkipCache bool `json:"skipCache"`
```
hamersaw, line 85: Is there a known use-case for this? Or just adding for completeness?

MorpheusXAUT: There is no use-case I could think of that would make sense, however since [`MergeExecutionConfig`](https://github.com/blackshark-ai/flyteadmin/blob/cache-eviction-execution-override/pkg/manager/impl/util/shared.go#L261) is using a [`WorkflowExecutionConfigInterface`](https://github.com/blackshark-ai/flyteadmin/blob/cache-eviction-execution-override/pkg/manager/impl/shared/iface.go#L12) and the application configuration is used as a fallback for other values as well, this was required to satisfy the interface.

It currently returns the actual setting (in case someone wants to set that), however we could also just always have it return `false` for the `ApplicationConfig` and not make it configurable at all on application level.

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -158,11 +163,16 @@ func (a *ApplicationConfig) GetInterruptible() *wrappers.BoolValue {
...
 }
 
+func (a *ApplicationConfig) GetOverwriteCache() bool {
+	return a.OverwriteCache
+}
+
 // GetAsWorkflowExecutionConfig returns the WorkflowExecutionConfig as extracted from this object
 func (a *ApplicationConfig) GetAsWorkflowExecutionConfig() admin.WorkflowExecutionConfig {
-	// These two should always be set, one is a number, and the other returns nil when empty.
+	// These three should always be set, the first two are a number/bool, and the other returns nil when empty.
```
hamersaw, line 172: nit - can we clean up this comment? it is a little difficult to understand.

hamersaw: perfect, thanks!

hamersaw (approved):

pkg/manager/impl/util/shared.go
```diff
@@ -291,5 +291,10 @@ func MergeIntoExecConfig(workflowExecConfig admin.WorkflowExecutionConfig, spec
 	if workflowExecConfig.GetInterruptible() == nil && spec.GetInterruptible() != nil {
 		workflowExecConfig.Interruptible = spec.GetInterruptible()
 	}
+
+	if !workflowExecConfig.GetOverwriteCache() && spec.GetOverwriteCache() {
```
katrogan, line 295: Interesting - so we can essentially disable caching for an entire project+domain?

katrogan (approved):

pkg/manager/impl/util/shared.go
MorpheusXAUT: I thought about that, similar to how I previously added the `interruptible` flag, however I didn't see the exact use case of overwriting it to `false` (being able to distinguish between not set/`false`).
Since the main intended use case was for a single execution, I didn't anticipate we'd need the "tri-state" bool, especially since it's also less confusing in the console UI for end users.

@katrogan If you'd prefer to have the option, I can certainly adapt it since it'll be a minor change anyways.

katrogan: Yeah I suppose the project-domain and even application setting use cases are themselves probably very nice to haves that I can't imagine are used frequently. 👍 We can always revisit this if it ends up being a feature request

PR #483, opened 2022-10-10
Populate TaskNodeMetadata on CreateNodeExecutionModel to correctly display fast cache

hamersaw:
# TL;DR
Current the [CacheStatus](https://github.com/flyteorg/flyteadmin/blob/d1ebaaea3b2025d78ce7c34e6881fa140143bb06/pkg/repositories/models/node_execution.go#L48) field on node executions is only set when [updating an existing model](https://github.com/flyteorg/flyteadmin/blob/d1ebaaea3b2025d78ce7c34e6881fa140143bb06/pkg/repositories/transformers/node_execution.go#L220-L225). In the case of fast cache work, where only a single node execution event will be sent for a node (ie. cache hit / succeeded), this information is never set and as a result the "cache hit" icon is never displayed in the UI. This PR sets CacheStatus if it exists on the first node execution event.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2886

## Follow-up issue
_NA_


pkg/manager/impl/node_execution_manager_test.go
```diff
@@ -132,6 +132,9 @@ func TestCreateNodeEvent(t *testing.T) {
 		StartedAt: occurredAtProto,
 		CreatedAt: occurredAtProto,
 		UpdatedAt: occurredAtProto,
+		TargetMetadata: &admin.NodeExecutionClosure_TaskNodeMetadata{
```
katrogan, line 135: why is this change necessary? we shouldn't be initializing the metadata in the closure unless it's sourced from the event itself, right?

katrogan: Ah sorry, I didn't realize the event under test actually had a non-empty [TargetMetadata](https://github.com/flyteorg/flyteadmin/blob/6a275357525ae9b6f31a5a62a55095849721cfd4/pkg/manager/impl/node_execution_manager_test.go#L68) - and we should therefore expect to see the closure here have an initialized TargetMetadata

katrogan (approved):

PR #485, opened 2022-10-12
Adding configurable audience property for flyte clients

pmahindrakar-oss:
Signed-off-by: pmahindrakar-oss <prafulla.mahindrakar@gmail.com>

# TL;DR
Most Authorization server provide a way to use default audience which can be used to request auth tokens which is what flyte relied on before this PR.

Following IDL PR https://github.com/flyteorg/flyteidl/pull/329 allows configuring this in client config

The current PR allows to configure this in the  PublicClientConfig endpoint of flyteadmin

If client config contains the audience field then that would be used otherwise whatever is available in PublicClientConfigResponse would be used

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue

https://github.com/flyteorg/flyte/issues/2959

## Follow-up issue
_NA_
OR


katrogan (approved):

auth/config/third_party_config.go
```diff
@@ -10,6 +10,7 @@ type FlyteClientConfig struct {
 	ClientID    string   `json:"clientId" pflag:",public identifier for the app which handles authorization for a Flyte deployment"`
 	RedirectURI string   `json:"redirectUri" pflag:",This is the callback uri registered with the app which handles authorization for a Flyte deployment"`
 	Scopes      []string `json:"scopes" pflag:",Recommended scopes for the client to request."`
+	Audience    string   `json:"audience" pflag:",Audience to use when initiating OAuth2 authorization requests."`
```
katrogan, line 13: should this have a default value?

katrogan (approved):

auth/config/third_party_config.go
pmahindrakar-oss: This is the dns name of the hosted flyte endpoint. Most of the auth providers like okta provide a way of specifying this during auth server creation and have a default value already available. This property allows to override this default in the auth provider. Since this is dependent on the user environment i haven't added a default value for it and expect to either leave it blank in which case it uses auth server default which is the case for our internal deployments too or have user configured audience value.

PR #486, opened 2022-10-12
Persist flyte deck during workflow recovery

pingsutw:
# TL;DR
Add deckURI to closure when calling [CreateNodeExecutionModel](https://github.com/flyteorg/flyteadmin/blob/7a56117d27b4a6b100d2e5442991d534ed08f6a0/pkg/manager/impl/node_execution_manager.go#L88).
In recovery mode, we call [CreateNodeEvent](https://github.com/flyteorg/flyteadmin/blob/7a56117d27b4a6b100d2e5442991d534ed08f6a0/pkg/manager/impl/node_execution_manager.go#L212) only once, which means we'll only call [createNodeExecutionWithEvent](https://github.com/flyteorg/flyteadmin/blob/7a56117d27b4a6b100d2e5442991d534ed08f6a0/pkg/manager/impl/node_execution_manager.go#L263) and won't call [updateNodeExecutionWithEvent](https://github.com/flyteorg/flyteadmin/blob/7a56117d27b4a6b100d2e5442991d534ed08f6a0/pkg/manager/impl/node_execution_manager.go#L270). Therefore, we should also add deck path to closure when creating node execution (currently, we only add deck path to closure when updating the node execution).
## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
To reproduce, register and run below workflow
```python
import flytekit
from flytekit import task, workflow


@task()
def t1() -> int:
    flytekit.Deck("demo", "#Hello Flyte\n##Hello Flyte\n###Hello Flyte")
    return 1


@task()
def t2(a: int) -> int:
    print(d)
    return a


@workflow
def wf():
    t2(a=t1())


if __name__ == "__main__":
    wf()
```
image

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2930

## Follow-up issue
_NA_


pkg/repositories/transformers/node_execution.go
```diff
@@ -118,6 +118,7 @@ func CreateNodeExecutionModel(ctx context.Context, input ToNodeExecutionModelInp
 		Phase:     input.Request.Event.Phase,
 		CreatedAt: input.Request.Event.OccurredAt,
 		UpdatedAt: input.Request.Event.OccurredAt,
+		DeckUri:   input.Request.Event.DeckUri,
```
katrogan, line 121: is the deck URI always written on every event? do we need to update `UpdateNodeExecutionModel` too?

pingsutw: we did it already. [addTerminalState](https://github.com/flyteorg/flyteadmin/blob/master/pkg/repositories/transformers/node_execution.go#L98) in UpdateNodeExecutionModel

pingsutw: > is the deck URI always written on every event

Only when the task succeeds.

hamersaw:
@pingsutw per your comment adding the `deckURI` already happens in the [addTerminalState](https://github.com/flyteorg/flyteadmin/blob/master/pkg/repositories/transformers/node_execution.go#L98) function. It looks like this is already called in the [CreateNodeExecutionModel function](https://github.com/flyteorg/flyteadmin/blob/1817f38b977e9e821f732fb439cdd2f0acaead21/pkg/repositories/transformers/node_execution.go#L137) and the `RECOVERED` phase is already [registered as a terminal node phase](https://github.com/flyteorg/flyteadmin/blob/1817f38b977e9e821f732fb439cdd2f0acaead21/pkg/common/executions.go#L33). So shouldn't the deckURI already be getting set when recovering a node? Can you help me understand why this is not the case?

pingsutw:
@hamersaw Thanks, good catch. we do add deckURI in [addTerminalState](https://github.com/flyteorg/flyteadmin/blob/master/pkg/repositories/transformers/node_execution.go#L98), so we don't need this change anymore. closing it.

PR #487, opened 2022-10-13
More data in workflow exists err msg #2522

jerempy:
Return more data if workflow already exists with different structure. Will have another PR in the python flytekit to compare the values. Flytekit PR: https://github.com/flyteorg/flytekit/pull/1235

# TL;DR
More data in err message for [#2522](https://github.com/flyteorg/flyte/issues/2522)

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
Fixes https://github.com/flyteorg/flyte/issues/2522

## Follow-up issue
_NA_
OR

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -16,10 +16,16 @@ type FlyteAdminError interface {
 	Error() string
 	Code() codes.Code
 	GRPCStatus() *status.Status
-	WithDetails(details *admin.EventFailureReason) (FlyteAdminError, error)
+	WithDetails(details ProtoMessager) (FlyteAdminError, error)
 	String() string
 }
 
+// This allows us to pass different protos to the FlyteAdminError.WithDetails() method
+type ProtoMessager interface {
+	ProtoMessage()
+	Reset()
+	String() string
+}
 type flyteAdminErrorImpl struct {
 	status *status.Status
 }
@@ -43,7 +49,7 @@ func (e *flyteAdminErrorImpl) String() string {
 // enclose the error in the format that grpc server expect from golang:
 //
 //	https://github.com/grpc/grpc-go/blob/master/status/status.go#L133
-func (e *flyteAdminErrorImpl) WithDetails(details *admin.EventFailureReason) (FlyteAdminError, error) {
+func (e *flyteAdminErrorImpl) WithDetails(details ProtoMessager) (FlyteAdminError, error) {
 	s, err := e.status.WithDetails(details)
 	if err != nil {
 		return nil, err
@@ -110,7 +116,7 @@ func NewIncompatibleClusterError(ctx context.Context, errorMsg, curCluster strin
 
 func NewWorkflowExistsDifferentStructureError(ctx context.Context, workflow *admin.Workflow) FlyteAdminError {
 	errorMsg := "workflow with different structure already exists"
-	statusErr, transformationErr := NewFlyteAdminError(codes.FailedPrecondition, errorMsg).WithDetails(&admin.CreateWorkflowFailureReason{
+	statusErr, transformationErr := NewFlyteAdminError(codes.InvalidArgument, errorMsg).WithDetails(&admin.CreateWorkflowFailureReason{
 		Reason: &admin.CreateWorkflowFailureReason_ExistsDifferentStructure{
 			ExistsDifferentStructure: &admin.WorkflowErrorExistsDifferentStructure{
 				Id: workflow.GetId(),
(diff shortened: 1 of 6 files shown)


EngHabu (dismissed):

EngHabu:
Should we do the same for tasks?

pkg/manager/impl/workflow_manager.go
```diff
@@ -169,8 +169,11 @@ func (w *WorkflowManager) CreateWorkflow(
 			return nil, errors.NewFlyteAdminErrorf(
 				codes.AlreadyExists, "identical workflow already exists with id %v", request.Id)
 		}
-		return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument,
-			"workflow with different structure already exists with id %v", request.Id)
+		// A workflow exists with different structure - sending much data to help debug on sdk end
+		return nil, errors.NewFlyteAdminErrorf(
+			codes.InvalidArgument,
+			"workflow with different structure already exists with that id: %v. workflow model: %+v.",
```
katrogan, line 175: if I'm understanding the original issue correctly, we want flytekit to be able to fetch the existing workflow with an identical digest and then do the comparison on the client-side in order to surface what in the workflow structure has specifically changed.

Right now the `existingMatchingWorkflow` type is an internal db model, with a serialized closure that's not particularly parseable by the flytekit client

I think what we should do is add a proper error type with structure, similar to [these](https://github.com/flyteorg/flyteadmin/blob/master/pkg/errors/errors.go#L82,L109) to store the existing workflow identifier and return this new error in this block. That way the flytekit client can then do a follow-up api call to Get the matching workflow, which will return the [full workflow closure](https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/admin/workflow.proto#L35) in a structured format that flytekit error handling can sensibly use to compare to the workflow definition it is trying to register

jerempy: @katrogan awesome feedback. thanks much! I took another shot at it. Let me know what you think and looks good, then I'll re-do the python end to in this PR: https://github.com/flyteorg/flytekit/pull/1235

katrogan:
awesome! I think for completeness you might first need to make a change in flyteidl, much like we have for the current [EventFailureReason](https://github.com/flyteorg/flyteidl/blob/967c75fc4761778836f542e6193712680af3afd3/protos/flyteidl/admin/event.proto#L25)s which allow us to specify the `Reason` structure in [event write errors](https://github.com/flyteorg/flyteadmin/blob/master/pkg/errors/errors.go#L85,L88). Your new message can embed the workflow identifier in the reason.

Since your change isn't related to event handling I think the best thing would be to define a new message in the flyteadmin idl, perhaps in [workflow.proto](https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/admin/workflow.proto) to capture WorkflowCreateFailureReason or something similar.

katrogan:
hey @jerempy thanks for making the idl changes! I think you should be able to pick them up here by pinning the idl version in go.mod with your [new workflow error](https://github.com/flyteorg/flyteidl/releases/tag/v1.2.0)

jerempy:
@katrogan :high-five-emoji: 😎 and we're back. Ready for review: I added in the new protos

jerempy:
@EngHabu I created the new issue for tracking doing this with tasks: https://github.com/flyteorg/flyte/issues/3027 🙏 if you could review this one again plz.

pkg/errors/errors.go
```diff
@@ -16,10 +16,16 @@ type FlyteAdminError interface {
 	Error() string
 	Code() codes.Code
 	GRPCStatus() *status.Status
-	WithDetails(details *admin.EventFailureReason) (FlyteAdminError, error)
+	WithDetails(details ProtoMessager) (FlyteAdminError, error)
```
EngHabu, line 19: I believe there is a `proto.Message` interface that looks awfully similar. maybe take a look at that?

EngHabu (commented):
Thank you for making the change! I've a small comment

pkg/errors/errors.go
jerempy: @EngHabu you're right! Found it and plugged it int. Thanks

jerempy:
Any insight what I should do here about the missing project code coverage? 😅

EngHabu (dismissed):
Cool thank you.
I'll merge it for you, don't worry

pkg/manager/impl/workflow_manager.go
```diff
@@ -161,16 +161,22 @@ func (w *WorkflowManager) CreateWorkflow(
...
-			return nil, errors.NewFlyteAdminErrorf(
-				codes.AlreadyExists, "identical workflow already exists with id %v", request.Id)
+		if bytes.Equal(workflowDigest, existingWorkflowModel.Digest) {
+			return nil, errors.NewWorkflowExistsIdenticalStructureError(ctx, &request)
 		}
-		return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument,
-			"workflow with different structure already exists with id %v", request.Id)
+		// A workflow exists with different structure
+		existingCtx := getWorkflowContext(ctx, request.Id)
+		existingWorkflow, err := util.GetWorkflow(existingCtx, w.db, w.storageClient, *request.Id)
```
katrogan, line 173: instead of calling get again on the existing workflow, can we just have `NewWorkflowExistsDifferentStructureError` take in a core.Identifier, since it doesn't need the full admin protobuf def?

katrogan (changes requested):
mind updating the errors to accept just the workflow identifier so we don't have to do a full get/transform on the existing workflow?

jerempy:
@katrogan done. looks better this way. nice call.

katrogan (approved):

PR #489, opened 2022-10-20
Remove shard prefix from upload location

Dread1982:
# TL;DR
~~Reorder prefixes for uploads in bucket. StoragePrefix should come before shard prefix.~~
Remove shard prefix from upload path.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
~~Paths for uploads now follow the same convention like e.g. for data, instead of writing shard-prefixes directly in the root folder of the bucket.~~
Paths for uploads now don't contain a shard prefix. See discussion below for details.

Before:
`<bucket-name>/<shard-prefix>/<storage-prefix>`
e.g. `my-bucket/xy/upload/some-domain/...`

Now:
~~`<bucket-name>/<storage-prefix>/<shard-prefix>`~~
~~e.g. `my-bucket/upload/xy/some-domain/...`~~
`<bucket-name>/<storage-prefix>`
e.g. `my-bucket/upload/some-domain/...`

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/2545

## Follow-up issue
_NA_

### DIFF
--- a/dataproxy/service.go
+++ b/dataproxy/service.go
@@ -6,7 +6,6 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net/url"
-	"strings"
 	"time"
 
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/ioutils"
@@ -64,7 +63,7 @@ func (s Service) CreateUploadLocation(ctx context.Context, req *service.CreateUp
 	md5 := base64.StdEncoding.EncodeToString(req.ContentMd5)
 	urlSafeMd5 := base32.StdEncoding.EncodeToString(req.ContentMd5)
 
-	storagePath, err := createShardedStorageLocation(ctx, s.shardSelector, s.dataStore, s.cfg.Upload,
+	storagePath, err := createStorageLocation(ctx, s.dataStore, s.cfg.Upload,
 		req.Project, req.Domain, urlSafeMd5, req.Filename)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


dataproxy/service.go
```diff
@@ -140,9 +140,6 @@ func (s Service) validateCreateDownloadLocationRequest(req *service.CreateDownlo
 func createShardedStorageLocation(ctx context.Context, shardSelector ioutils.ShardSelector, store *storage.DataStore,
 	cfg config.DataProxyUploadConfig, keyParts ...string) (storage.DataReference, error) {
 	keySuffixArr := make([]string, 0, 4)
-	if len(cfg.StoragePrefix) > 0 {
```
EngHabu, line 143: Since the only thing we now put in keySuffix is keyParts, I think we can get rid of the former completely and avoid the extra initialization

Dread1982: Good point. I adjusted it.

sbrunk:
Hey @EngHabu could you trigger the CI again? Looks like I don't have permissions for the flyteadmin repo yet to do that.

wild-endeavor (approved):

EngHabu (changes requested):
Can we plz hold off I'm checking on s3 recommendation for maximizing read/write speed

wild-endeavor:
@Dread1982 sorry for the back and forth, but could you update the PR to just remove the shard prefix entirely?  The addition of the shard prefix at this point seems a bit like premature optimization, esp given that the prefix already has project and domain, the suffix can be specified in the request, and the S3 doc seems to imply that it's possible to shard, if needed, across the entire length of the prefix and not just the first couple characters.

See https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html for more.

Dread1982:
@wild-endeavor @EngHabu I updated the PR as suggested. Please have a look.

wild-endeavor (approved):

EngHabu (approved):
Thank you for updating the PR!

PR #490, opened 2022-10-24
Skip database already exists errors

wild-endeavor:
# TL;DR
When in single-binary mode, data catalog and admin both attempt to initialize the database at the same time.  Error check when getting an error back from the db create call to skip the error if the other goroutine created it in the meantime.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Spin up a local db, pause it in the function so that `createPostgresDbIfNotExists` will not return early, create the database manually, resume and inspect the error returned.  Testing was done locally with a postgres container.

### DIFF
--- a/pkg/repositories/database.go
+++ b/pkg/repositories/database.go
@@ -141,9 +141,9 @@ func createPostgresDbIfNotExists(ctx context.Context, gormConfig *gorm.Config, p
 
 	if result.Error != nil {
 		if !isPgErrorWithCode(result.Error, pqDbAlreadyExistsCode) {
-			logger.Warningf(ctx, "Got DB already exists for [%v], skipping...", pgConfig.DbName)
 			return nil, result.Error
 		}
+		logger.Warningf(ctx, "Got DB already exists error for [%s], skipping...", pgConfig.DbName)
 	}
 	// Now try connecting to the db again
 	return gorm.Open(dialector, gormConfig)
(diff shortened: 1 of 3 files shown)


katrogan (dismissed):
If we're going to be replicating some of this in datacatalog it might be worthwhile consolidating common code in https://github.com/flyteorg/flytestdlib/tree/master/database - but can be left as a todo

pkg/repositories/database.go
```diff
@@ -139,21 +140,24 @@ func createPostgresDbIfNotExists(ctx context.Context, gormConfig *gorm.Config, p
 	result := gormDb.Exec(createDBStatement)
 
 	if result.Error != nil {
-		return nil, result.Error
+		if !isPgErrorWithCode(result.Error, pqDbAlreadyExistsCode) {
+			logger.Warningf(ctx, "Got DB already exists for [%v], skipping...", pgConfig.DbName)
```
katrogan, line 144: nit: this log statement is wrong, right?

wild-endeavor: is it? why?

katrogan: because you're checking that it's not pqDbAlreadyExistsCode

katrogan (approved):

pkg/repositories/database.go
wild-endeavor: 🤦

PR #491, opened 2022-10-24
Expose kubeclient configuration

hamersaw:
# TL;DR
Adding `KubeClientConfig` to the FlyteAdmin configuration. This can be set at the `server` level, which then serves as a default of there are multiple clusters configured. Additionally, each cluster can set this individually.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
The only thing I don't like is setting `QPS` in the `KubeClientConfig` as an `int32` and then casting to `float32` when setting on the `kubeConfiguration`. This is necessary because we have to set pflags to "-" in the case of float32 because the library does not support floats, so the QPS value is not carried over.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -27,7 +27,7 @@ type ServerConfig struct {
 
 	DataProxy                DataProxyConfig  `json:"dataProxy" pflag:",Defines data proxy configuration."`
 	ReadHeaderTimeoutSeconds int              `json:"readHeaderTimeoutSeconds" pflag:",The amount of time allowed to read request headers."`
-	KubeClientConfig         KubeClientConfig `json:"kube-client-config" pflag:",Configuration to control the Kubernetes client"`
+	KubeClientConfig         KubeClientConfig `json:"kubeClientConfig" pflag:",Configuration to control the Kubernetes client"`
 }
 
 type DataProxyConfig struct {
@@ -52,7 +52,7 @@ type GrpcConfig struct {
 	MaxMessageSizeBytes int  `json:"maxMessageSizeBytes" pflag:",The max size in bytes for incoming gRPC messages"`
 }
 
-// KubeClientConfig contains the configuration used by flytepropeller to configure its internal Kubernetes Client.
+// KubeClientConfig contains the configuration used by flyteadmin to configure its internal Kubernetes Client.
 type KubeClientConfig struct {
 	// QPS indicates the maximum QPS to the master from this client.
 	// If it's zero, the created RESTClient will use DefaultQPS: 5
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -25,8 +25,9 @@ type ServerConfig struct {
 	// Deprecated: please use auth.AppAuth.ThirdPartyConfig instead.
 	DeprecatedThirdPartyConfig authConfig.ThirdPartyConfigOptions `json:"thirdPartyConfig" pflag:",Deprecated please use auth.appAuth.thirdPartyConfig instead."`
 
-	DataProxy                DataProxyConfig `json:"dataProxy" pflag:",Defines data proxy configuration."`
-	ReadHeaderTimeoutSeconds int             `json:"readHeaderTimeoutSeconds" pflag:",The amount of time allowed to read request headers."`
+	DataProxy                DataProxyConfig  `json:"dataProxy" pflag:",Defines data proxy configuration."`
+	ReadHeaderTimeoutSeconds int              `json:"readHeaderTimeoutSeconds" pflag:",The amount of time allowed to read request headers."`
+	KubeClientConfig         KubeClientConfig `json:"kube-client-config" pflag:",Configuration to control the Kubernetes client"`
 }
 
 type DataProxyConfig struct {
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -97,6 +110,11 @@ var defaultServerConfig = &ServerConfig{
 		},
 	},
 	ReadHeaderTimeoutSeconds: 32, // just shy of requestTimeoutUpperBound
+	KubeClientConfig: KubeClientConfig{
+		QPS:     100,
```
katrogan, line 114: nit: do you want to keep these defaults in sync with the RESTClient defaults or are these more sensible/preferred?

katrogan (approved):

pkg/runtime/interfaces/cluster_configuration.go
```diff
@@ -3,15 +3,18 @@ package interfaces
...
 type ClusterConfig struct {
-	Name     string `json:"name"`
-	Endpoint string `json:"endpoint"`
-	Auth     Auth   `json:"auth"`
-	Enabled  bool   `json:"enabled"`
+	Name             string                   `json:"name"`
+	Endpoint         string                   `json:"endpoint"`
+	Auth             Auth                     `json:"auth"`
+	Enabled          bool                     `json:"enabled"`
+	KubeClientConfig *config.KubeClientConfig `json:"kube-client-config,omitempty"`
```
pmahindrakar-oss, line 17: I dont think we use this fomat (kebab-case?) but camelCase for all the configs in flyteadmin.
We can keep it consistent with other configs in this repo. Propeller and plugins repo i see use this format but may be we can keep using camelCase here

pkg/config/config.go
```diff
@@ -25,8 +25,9 @@ type ServerConfig struct {
 	// Deprecated: please use auth.AppAuth.ThirdPartyConfig instead.
 	DeprecatedThirdPartyConfig authConfig.ThirdPartyConfigOptions `json:"thirdPartyConfig" pflag:",Deprecated please use auth.appAuth.thirdPartyConfig instead."`
 
-	DataProxy                DataProxyConfig `json:"dataProxy" pflag:",Defines data proxy configuration."`
-	ReadHeaderTimeoutSeconds int             `json:"readHeaderTimeoutSeconds" pflag:",The amount of time allowed to read request headers."`
+	DataProxy                DataProxyConfig  `json:"dataProxy" pflag:",Defines data proxy configuration."`
+	ReadHeaderTimeoutSeconds int              `json:"readHeaderTimeoutSeconds" pflag:",The amount of time allowed to read request headers."`
+	KubeClientConfig         KubeClientConfig `json:"kube-client-config" pflag:",Configuration to control the Kubernetes client"`
```
pmahindrakar-oss, line 30: similarly here , probably use camelCase to be consistent

pkg/flytek8s/client.go
```diff
@@ -37,23 +38,14 @@ func RemoteClusterConfig(host string, auth runtimeInterfaces.Auth) (*restclient.
...
-
 // Initializes a config using a variety of configurable or default fallback options that can be passed to a Kubernetes client on
 // initialization.
 func GetRestClientConfig(kubeConfig, master string,
 	k8sCluster *runtimeInterfaces.ClusterConfig) (*restclient.Config, error) {
 	var kubeConfiguration *restclient.Config
 	var err error
 
+	kubeClientConfig := &config.GetConfig().KubeClientConfig
 	if kubeConfig != "" {
```
pmahindrakar-oss, line 49: Can we rename kubeConfig to kubeConfigPathString or something  specific because we have too many variations of kubeConfig , kubeConfiguration , kubeClientConfig etc

pkg/config/config.go
```diff
@@ -51,6 +52,18 @@ type GrpcConfig struct {
 	MaxMessageSizeBytes int  `json:"maxMessageSizeBytes" pflag:",The max size in bytes for incoming gRPC messages"`
 }
 
+// KubeClientConfig contains the configuration used by flytepropeller to configure its internal Kubernetes Client.
```
pmahindrakar-oss, line 55: You mean flyteadmin here instead of flytepropeller

pkg/config/config.go
```diff
@@ -51,6 +52,18 @@ type GrpcConfig struct {
...
+type KubeClientConfig struct {
+	// QPS indicates the maximum QPS to the master from this client.
+	// If it's zero, the created RESTClient will use DefaultQPS: 5
+	QPS int32 `json:"qps" pflag:",Max QPS to the master for requests to KubeAPI. 0 defaults to 5."`
+	// Maximum burst for throttle.
+	// If it's zero, the created RESTClient will use DefaultBurst: 10.
+	Burst int `json:"burst" pflag:",Max burst rate for throttle. 0 defaults to 10"`
+	// The maximum length of time to wait before giving up on a server request. A value of zero means no timeout.
+	Timeout config.Duration `json:"timeout" pflag:",Max duration allowed for every request to KubeAPI before giving up. 0 implies no timeout."`
+}
```
pmahindrakar-oss, line 65: I feel the way this is used is not encompassing the entire kubeClientConfig but just the rest API configuration.
Basically we have the connection parameters coming from one place
* kubeConfigPath 
* runtimeInterfaces.ClusterConfig
But just the rest Api kube config from this . 
Can we make this specific to mean that so that its not confusing

pkg/flytek8s/client.go
```diff
@@ -63,14 +55,29 @@ func GetRestClientConfig(kubeConfig, master string,
...
-		return GetRestClientConfigForCluster(*k8sCluster)
+		kubeConfiguration, err = RemoteClusterConfig(k8sCluster.Endpoint, k8sCluster.Auth)
+		if err != nil {
+			return nil, err
+		}
+		logger.Debugf(context.Background(), "successfully loaded kube configuration from %v", k8sCluster)
+
+		if k8sCluster.KubeClientConfig != nil {
+			kubeClientConfig = k8sCluster.KubeClientConfig
+		}
```
pmahindrakar-oss, line 66: Can we log here that we are using the rest config from remote cluster override

pmahindrakar-oss (dismissed):

pkg/config/config.go
hamersaw: By RESTClient defaults do you mean the regular k8s defaults? I'm not sure I understand, but if so it could be useful. The only place these will come in to play is the speed that users can create workflow CRDs. A user was using a workflow which internally started 100 launchplans and found that they are throttled pretty aggressively.

pkg/runtime/interfaces/cluster_configuration.go
hamersaw: done

pkg/config/config.go
hamersaw: done

pkg/flytek8s/client.go
hamersaw: Thanks for pointing this out, I thought the same thing!

pkg/flytek8s/client.go
hamersaw: added a message

pkg/config/config.go
hamersaw: Do you suggest just renaming this configuration?

So I copied the config definition straight from propeller. IMO keeping these things uniform between the repos makes sense, but we already changed that by using camel instead of kebob case. Happy to change this however you suggest!

pmahindrakar-oss: I am ok with the current one with kebob case.

pmahindrakar-oss (approved):

pkg/config/config.go
katrogan: I mean the ones in the comment where you define the `KubeClientConfig`- not a big deal either way

PR #495, opened 2022-11-03
Adding CheckAndFetchInputsForExecution on single task executions

hamersaw:
# TL;DR
Currently executing a single task does not validate inputs on the auto-generated launchplan. However, an ancillary effect is the function used for this validation converts `nil` input values to an empty LiteralMap. This is important because executing `pyflyte run` on a task will call this stack and it fails because the input values are nil. This PR adds the validation to single task execution, allowing defaulted values and automatic handling of `nil` inputs.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/3008

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/execution_manager_test.go
+++ b/pkg/manager/impl/execution_manager_test.go
@@ -3903,6 +3903,18 @@ func TestCreateSingleTaskExecution(t *testing.T) {
 			}, nil
 		})
 
+	var launchplan *models.LaunchPlan
+	repository.LaunchPlanRepo().(*repositoryMocks.MockLaunchPlanRepo).SetCreateCallback(func(input models.LaunchPlan) error {
+		launchplan = &input
+		return nil
+	})
+	repository.LaunchPlanRepo().(*repositoryMocks.MockLaunchPlanRepo).SetGetCallback(func(input interfaces.Identifier) (models.LaunchPlan, error) {
+		if launchplan == nil {
+			return models.LaunchPlan{}, flyteAdminErrors.NewFlyteAdminError(codes.NotFound, "launchplan not found")
+		}
+		return *launchplan, nil
+	})
+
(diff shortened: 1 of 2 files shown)


hamersaw:
@pingsutw can you look at this and I'm interested on your thoughts as an alternative solution to https://github.com/flyteorg/flytepropeller/pull/498?

pkg/manager/impl/execution_manager_test.go
```diff
@@ -3947,45 +3959,17 @@ func TestCreateSingleTaskExecution(t *testing.T) {
...
-				},
-			},
-		},
-	}, time.Now())
 
+	// test once to create an initial launchplan
+	_, err := execManager.CreateExecution(context.TODO(), request, time.Now())
+	assert.NoError(t, err)
+
+	// test again to ensure existing launchplan retreival works
```
eapolinario, line 3971: typo: retreival

eapolinario (commented):
Can we add a test specific for a parameterless task?

hamersaw:
@eapolinario we could add a unit test for nil input interface. We are already testing that scenario in the [underlying function](https://github.com/flyteorg/flyteadmin/blob/4c58e781f52f429110f46b6fc86f99c171dcccce/pkg/manager/impl/validation/execution_validator_test.go#L139-L157). The issue here is that we weren't calling that function. Now that we are, I'm not sure it makes sense to add this test - if we do so for completeness it would make sense to add tests for every input combination (ie. castable, missing, etc).

eapolinario (approved):

wild-endeavor (approved):

PR #499, opened 2022-12-06
Update kafka config

pingsutw:
Signed-off-by: Kevin Su <pingsutw@apache.org>

# TL;DR
Json decoder failed to decode the version in yaml file because the `version` in [sarama.KafkaVersion](https://github.com/Shopify/sarama/blob/978aadd67322a6ee5b790d700698a5ec98aa9170/utils.go#L114) isn't exported. 

Json decoder can only assign the value to the exported field. https://pkg.go.dev/encoding/json#Unmarshal

> Struct values encode as JSON objects. Each exported struct field becomes a member of the object, using the field name as the object key, unless the field is omitted for one of the reasons given below.

- Add `KafkaVersion` in the Kafka config, and use `sarama.ParseKafkaVersion` to parse it.
- Remove `sender.Close` when creating Kafka Client, which is unnecessary. It will be used when sending the event to Kafka.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
- https://github.com/flyteorg/flyte/issues/3117
- https://flyte-org.slack.com/archives/CP2HDHKE1/p1669385925148139

## Follow-up issue
_NA_

### DIFF
--- a/pkg/runtime/interfaces/application_configuration.go
+++ b/pkg/runtime/interfaces/application_configuration.go
@@ -1,7 +1,6 @@
 package interfaces
 
 import (
-	"github.com/Shopify/sarama"
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
 	"github.com/flyteorg/flytestdlib/config"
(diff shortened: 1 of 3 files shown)


pingsutw:
Make sense, Updated it, thanks.

pkg/runtime/interfaces/application_configuration.go
```diff
@@ -194,7 +193,8 @@ type GCPConfig struct {
 }
 
 type KafkaConfig struct {
-	Version sarama.KafkaVersion
+	// The version of Kafka, e.g. 2.1.0, 0.8.2.0
+	Version string `json:"kafkaVersion"`
```
hamersaw, line 197:
```suggestion
	Version string `json:"version"`
```

hamersaw (approved):

katrogan (approved):

PR #500, opened 2022-12-06
Add httpProxyURL support in OAuth2

goyalankit:
# TL;DR
Allows operators to specify an external HTTP Proxy to do OAuth. 

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Often the On-Prem deployments don't have a direct external (internet) access to do OAuth2 and is usually enabled through an external Proxy. This commit allows operators of Flyte to provide an HTTP Proxy URL to do the OAuth.

## Tracking Issue
_NA_

fixes https://github.com/flyteorg/flyte/issues/

## Follow-up issue
_NA_


goyalankit:
> Hey @goyalankit can this be done/maintained through a mesh/service (e.g. linkerd / envoy)? that seems to be the right layer to encode these rules...

Thank you @EngHabu for the reply. We discussed this with @wild-endeavor and he believed that it might not be straightforward to do in ingress. And that it makes sense to put this in the code

auth/auth_context.go
```diff
@@ -133,6 +133,11 @@ func NewAuthenticationContext(ctx context.Context, sm core.SecretManager, oauth2
 		Timeout: IdpConnectionTimeout,
 	}
 
+	if options.UserAuth.HTTPProxyURL.String() != "" {
+		logger.Infof(ctx, "HTTPProxy URL for OAuth2 is: %s", options.UserAuth.HTTPProxyURL.String())
```
wild-endeavor, line 137: oh sorry - was looking at the wrong tab... this is good thanks.

wild-endeavor (approved):

wild-endeavor:
oh unit test failure - something with the mock.  could you take a look @goyalankit please?

auth/auth_context.go
```diff
@@ -133,6 +133,11 @@ func NewAuthenticationContext(ctx context.Context, sm core.SecretManager, oauth2
 		Timeout: IdpConnectionTimeout,
 	}
 
+	if options.UserAuth.HTTPProxyURL.String() != "" {
```
EngHabu, line 136: Nit:
```suggestion
	if len(options.UserAuth.HTTPProxyURL.String()) > 0 {
```

wild-endeavor (approved):

PR #501, opened 2022-12-09
Add launch entity type to execution model

katrogan:

# TL;DR
Add launch entity type to execution model to use in filtering List queries

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/3144

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/execution_manager.go
+++ b/pkg/manager/impl/execution_manager.go
@@ -975,7 +975,7 @@ func (m *ExecutionManager) launchExecutionAndPrepareModel(
 		InputsURI:             inputsURI,
 		UserInputsURI:         userInputsURI,
 		SecurityContext:       executionConfig.SecurityContext,
-		LaunchEntity:          request.Spec.LaunchPlan.ResourceType,
+		LaunchEntity:          launchPlan.Id.ResourceType,
 	})
 	if err != nil {
 		logger.Infof(ctx, "Failed to create execution model in transformer for id: [%+v] with err: %v",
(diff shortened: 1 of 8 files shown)


pkg/manager/impl/execution_manager.go
```diff
@@ -974,6 +975,7 @@ func (m *ExecutionManager) launchExecutionAndPrepareModel(
 		InputsURI:             inputsURI,
 		UserInputsURI:         userInputsURI,
 		SecurityContext:       executionConfig.SecurityContext,
+		LaunchEntity:          request.Spec.LaunchPlan.ResourceType,
```
pmahindrakar-oss, line 978: Instead of getting from request spec . Can we not get this from launchplan object.

Also i think we need the same change in launchSingleTaskExecution

katrogan: The launch plan object is a model which doesn't have a core.identifier. Just curious, what would the difference be if it's the same identifier? 

The same change is already done for single task executions in line 724 above.

pmahindrakar-oss: Wanted to avoid usage of request object beyond a certain scope of the function . Ideally starting of service function is where it should be scoped and since its a large function would be good to get this info from derived object if possible.

I see we have this.
```
launchPlan, err := transformers.FromLaunchPlanModel(launchPlanModel)``` 

We should be able to get this from `launchPlan.Id.ResourceType` right ?

I am fine though we this too.

pmahindrakar-oss (dismissed):

pkg/manager/impl/execution_manager.go
katrogan: thanks for explaining, updated the usage here

pmahindrakar-oss (approved):

PR #511, opened 2023-01-10
Forward all claims in userinfo response

katrogan:

# TL;DR
Return all claims from OIDC user info server response as custom metadata headers.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
_NA_

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/3225

## Follow-up issue
_NA_

### DIFF
--- a/auth/identity_context.go
+++ b/auth/identity_context.go
@@ -2,9 +2,10 @@ package auth
 
 import (
 	"context"
+	"time"
+
 	"github.com/flyteorg/flytestdlib/logger"
 	"google.golang.org/protobuf/types/known/structpb"
-	"time"
 
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/service"
 
--- a/auth/identity_context.go
+++ b/auth/identity_context.go
@@ -53,12 +53,10 @@ func (c IdentityContext) UserInfo() *service.UserInfoResponse {
 
 	userInfo := c.userInfo
 	if c.claims != nil {
-		logger.Infof(context.TODO(), "**Marshalling claims [%+v]", *c.claims)
 		m, err := structpb.NewStruct(*c.claims)
 		if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/server/service.go
```diff
@@ -200,6 +197,9 @@ func newHTTPServer(ctx context.Context, cfg *config.ServerConfig, _ *authConfig.
 		gwmuxOptions = append(gwmuxOptions, runtime.WithMetadata(auth.GetHTTPMetadataTaggingHandler()))
 	}
 
+	// This option sets subject in the user info response
+	gwmuxOptions = append(gwmuxOptions, runtime.WithForwardResponseOption(auth.GetUserInfoForwardResponseHandler()))
+
```
EngHabu, line 202: Doesn't need to move anymore, right?

auth/identity_context.go
```diff
@@ -48,6 +51,14 @@ func (c IdentityContext) UserInfo() *service.UserInfoResponse {
...
 	}
 
+	userInfo := c.userInfo
+	if c.claims != nil {
+		m, err := structpb.NewStruct(*c.claims)
+		if err != nil {
+			logger.Errorf(context.TODO(), "Failed to marshal claims [%+v] to struct: %v", c.claims, err)
+		}
+		userInfo.AdditionalClaims = m
+	}
```
EngHabu, line 61: Can we move this to the constructor NewIdentityContext? no need to do this every time...

EngHabu: Also, let's refactor the constructor to return an Error and don't silence the error from here?
Also, I think we can remove the error log as long as you augment the error you will return with the information

pkg/server/service.go
katrogan: done, thanks

EngHabu (approved):

PR #522, opened 2023-02-15
Only merge system defaults when querying workflow execution config for project attributes

wild-endeavor:
# TL;DR
The project-level-only matchable attribute endpoint currently merges in defaults from the server config (as this was the quickest way to get things going in light of the need to expose this to the UI and the complexity of the broader settings project).  There was a bug in the [initial implementation](https://github.com/flyteorg/flyteadmin/pull/480) though - these system level defaults should only be merged when the user is querying for workflow execution configs.  This just adds a check for that.

See the original PR for more info as well.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Also update some the default config.yaml to work with the new single-binary based demo sandbox.

## Tracking Issue
Just noticed this in testing.

### DIFF
--- a/flyteadmin_config.yaml
+++ b/flyteadmin_config.yaml
@@ -6,7 +6,7 @@ server:
   httpPort: 8088
   grpcPort: 8089
   grpcServerReflection: true
-  kube-config: /Users/ytong/.flyte/state/kubeconfig
+  kube-config: /Users/ytong/.flyte/sandbox/kubeconfig
   security:
     secure: false
     useAuth: false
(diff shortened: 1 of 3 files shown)


katrogan (approved):
Is it worth adding a unit test?

katrogan (dismissed):

katrogan (approved):

PR #537, opened 2023-03-09
Validate user-provided entity versions are url safe

katrogan:
# TL;DR
Ensures that user-provided entity version are url safe since they are used http paths to manipulate the entities and can lead to 404s when they include slashes.

Execution names, which are also user-provided and used in http paths are already [constrained](https://github.com/flyteorg/flyteadmin/blob/master/pkg/manager/impl/validation/execution_validator.go#L22) to a smaller subset of allowable chars

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Encountered with a user

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/3427

## Follow-up issue
_NA_

### DIFF
--- a/pkg/manager/impl/validation/validation.go
+++ b/pkg/manager/impl/validation/validation.go
@@ -24,7 +24,7 @@ var entityToResourceType = map[common.Entity]core.ResourceType{
 }
 
 // See https://www.rfc-editor.org/rfc/rfc3986#section-2.2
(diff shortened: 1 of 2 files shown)


katrogan:
cc @cosmicBboy

PR #546, opened 2023-03-29
Address resolution

wild-endeavor:
# TL;DR
Add logic to return and accept `flyte://` when getting data

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
Related to: https://github.com/flyteorg/flyteidl/pull/391


pkg/common/flyte_url.go
```diff
@@ -0,0 +1,92 @@
...
+	OUTPUT             = "o"
+	DECK               = "d"
+)
+
+func ParseFlyteURL(flyteURL string) (core.NodeExecutionIdentifier, *int, IOType, error) {
+	// flyteURL is of the form flyte://v1/project/domain/execution_id/node_id/attempt/[iod]
+	// where i stands for inputs.pb o for outputs.pb and d for the flyte deck
+	// If the retry attempt is missing, the io requested is assumed to be for the node instead of the task execution
+	zero := 0
+	re, err := regexp.Compile("flyte://v1/([a-zA-Z0-9_-]+)/([a-zA-Z0-9_-]+)/([a-zA-Z0-9_-]+)/([a-zA-Z0-9_-]+)(?:/([0-9]+))?/([iod])")
```
kumare3, line 25: can you move this to the top of the file? Why compile it again? Lets compile it once on app load?

pkg/common/flyte_url.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"fmt"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
+	"regexp"
+	"strconv"
+)
+
+type IOType string
```
kumare3, line 11: you can use `//go:generate enumer` and it will autogenerate the lookup as well. you wont need to 
example: https://github.com/flyteorg/flytepropeller/blob/dd49c9df317b9830c98641106d3b2cdd8fa18eb2/manager/config/config.go#L10

pkg/common/flyte_url.go
```diff
@@ -0,0 +1,92 @@
...
+	var attempt *int // nil means node execution, not a task execution
+	if len(matches) == 7 && matches[5] != "" {
+		a, err := strconv.Atoi(matches[5])
+		if err != nil {
+			return core.NodeExecutionIdentifier{}, &zero, UndefinedIo, fmt.Errorf("failed to parse attempt, %s", err)
+		}
+		attempt = &a
+	}
+	var ioType IOType
+	switch matches[len(matches)-1] {
```
kumare3, line 47: once you `//go:generate enumer` you wont need this switch, a simple lookup would work

pkg/common/flyte_url.go
wild-endeavor: oh right

dataproxy/service.go
```diff
@@ -231,9 +247,114 @@ func createStorageLocation(ctx context.Context, store *storage.DataStore,
 	return storagePath, nil
 }
 
+func (s Service) validateResolveArtifactRequest(req *service.GetDataRequest) error {
+	if req.GetFlyteUrl() == "" {
```
EngHabu, line 251: nit:
```suggestion
	if len(req.GetFlyteUrl()) == 0 {
```

dataproxy/service.go
```diff
@@ -231,9 +247,114 @@ func createStorageLocation(ctx context.Context, store *storage.DataStore,
...
+	}
+
+	return nil
+}
+
+func (s Service) GetTaskExecutionID(ctx context.Context, attempt int, nodeExecID core.NodeExecutionIdentifier) (*core.TaskExecutionIdentifier, error) {
+	taskExecs, err := s.taskExecutionManager.ListTaskExecutions(ctx, admin.TaskExecutionListRequest{
+		NodeExecutionId: &nodeExecID,
+		Limit:           1,
+		Filters:         fmt.Sprintf("eq(retry_attempt,%s)", strconv.Itoa(attempt)),
```
EngHabu, line 265: Is there a strongly typed way of doing this?

dataproxy/service.go
```diff
@@ -231,9 +247,114 @@ func createStorageLocation(ctx context.Context, store *storage.DataStore,
...
+	})
+	if err != nil || len(taskExecs.TaskExecutions) == 0 {
+		return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument, "failed to list task executions [%v]. Error: %v", nodeExecID, err)
```
EngHabu, line 268: I think we should have two separate errors... err == nil and len == 0 should not say "failed to list...."

pkg/common/flyte_url.go
```diff
@@ -0,0 +1,91 @@
...
+const (
+	ArtifactTypeUndefined ArtifactType = iota
+	ArtifactTypeI                      // inputs
+	ArtifactTypeO                      // outputs
+	ArtifactTypeD                      // deck
+)
+
+var re = regexp.MustCompile("flyte://v1/([a-zA-Z0-9_-]+)/([a-zA-Z0-9_-]+)/([a-zA-Z0-9_-]+)/([a-zA-Z0-9_-]+)(?:/([0-9]+))?/([iod])$")
+
+func ParseFlyteURL(flyteURL string) (core.NodeExecutionIdentifier, *int, ArtifactType, error) {
```
EngHabu, line 28: I kinda like to rewrite this to embed the names of the groups in the regex itself and use this utility function to return a map of group name to value... just to encapsulate the matching/ordering logic in one place.

```suggestion

func MatchRegex(reg *regexp.Regexp, input string) map[string]string {
	names := reg.SubexpNames()
	res := reg.FindAllStringSubmatch(input, -1)

	if len(res) == 0 {
		return nil
	}

	dict := make(map[string]string, len(names))
	for i := 1; i < len(res[0]); i++ {
		dict[names[i]] = res[0][i]
	}

	return dict
}

func ParseFlyteURL(flyteURL string) (core.NodeExecutionIdentifier, *int, ArtifactType, error) {
```

dataproxy/service.go
wild-endeavor: not really... i think we should punt on this.

pkg/common/flyte_url.go
```diff
@@ -0,0 +1,98 @@
...
+	proj := matches["project"]
+	domain := matches["domain"]
+	executionID := matches["exec"]
+	nodeID := matches["node"]
+	var attempt *int // nil means node execution, not a task execution
+	if matches["attempt"] != "" {
+		a, err := strconv.Atoi(matches["attempt"])
+		if err != nil {
+			return core.NodeExecutionIdentifier{}, nil, ArtifactTypeUndefined, fmt.Errorf("failed to parse attempt, %s", err)
+		}
```
EngHabu, line 54: nit, also print the value of attempt
```suggestion
	if attempt := matches["attempt"]; len(attempt) > 0 {
		a, err := strconv.Atoi(attempt)
		if err != nil {
			return core.NodeExecutionIdentifier{}, nil, ArtifactTypeUndefined, fmt.Errorf("failed to parse attempt [%v], %v", attempt, err)
		}
```

EngHabu (dismissed):

EngHabu (approved):

PR #523, opened 2023-02-15
Truncate execution error message in list execution calls

katrogan:
# TL;DR
For repeated execution failures with long error message stack traces, the call to list executions can exceed the default gRPC message size.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added
 - [ ] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description
 _How did you fix the bug, make the feature etc. Link to any design docs etc_

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3337

## Follow-up issue
_NA_

PR #555, opened 2023-05-04
Increase trimmedErrMessageLen

pradithya:
# TL;DR
Increase `trimmedErrMessageLen` so that error message in list of workflow execution is sufficiently informative.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed

## Complete description

The current value of `trimmedErrMessageLen` limit the error message shown in the workflow execution list page too much that it become uninformative. Users have to go to the corresponding execution details to get the full error message even for a short error message.
This PR increases the limit of error message length to 10240 so that most error message can be shown properly. 

https://user-images.githubusercontent.com/4023015/236148171-664ef81e-2525-4793-b0b1-8c18d0f7dab3.mov

## Follow-up issue
_NA_


pmahindrakar-oss (approved):
The message length limitation was added as part of this PR https://github.com/flyteorg/flyteadmin/pull/523 due to some ListExecution API's reaching the max grpc message limit. 
Max grpc limit size is 4MB, with that size even considering just this lenght of the error message , the List Api can return around 400 executions in the listApi but given that per page limit is 100 this should still be ok. More preferable IMO but would be separate trimmedErrMessageLen for list API calls

wild-endeavor (commented):
on second thought @pradithya - can you make this configurable?  your new value as the default is fine, but it would be nice to be configurable.

pradithya:
> on second thought @pradithya - can you make this configurable? your new value as the default is fine, but it would be nice to be configurable.

Make sense, let me have a look.

pradithya:
@wild-endeavor I updated the PR to make it configurable. But it seems the CI is failing due to Github action rate limit. Can you help retry it?

pingsutw (approved):
@pradithya Thanks, LGTM. Could you resolve branch conflict?

PR #557, opened 2023-05-08
Add RateLimit middleware

LaPetiteSouris:
# TL;DR

Solve https://github.com/flyteorg/flyte/issues/327

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

### Main changes
- Add a simple rate limiter middleware.
- The RateLimiter components rely on an in-memory map storage to store `userID` and `rate` per user ID. `rate` is a [native Golang package](https://pkg.go.dev/golang.org/x/time/rate) to track total count of requests.
- The in-memory storage is cleaned periodically to reduce memory footprint.
- The component, which is considered as part of the security scope, relies on the identity of the user provided by `auth` package. In fact, the best way to uniquely identify requests to rate limit is to track usage per authenticated user.
### Tradeoff
- To keep it simple, an in-memory map storage is used to track rate per user. This becomes inaccurate if multiple instances of `flyteadmin` is deployed. If this is the case and we are serious about rate limit, another improvement is needed, like introduction of Redis for example, which is out of scopes for this PR.

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/327

## Follow-up issue
_NA_


pkg/config/config.go
```diff
@@ -80,14 +80,23 @@ type ServerSecurityOptions struct {
...
 }
 
+// declare RateLimitConfig
+type RateLimitOptions struct {
+	Enabled           bool            `json:"enabled"`
+	RequestsPerSecond int             `json:"requestsPerSecond"`
+	BurstSize         int             `json:"burstSize"`
+	CleanupInterval   config.Duration `json:"cleanupInterval"`
+}
+
```
LaPetiteSouris, line 99: This part needs docs for clarification. Would some one please kindly show me where to update the docs for configs ?

Thanks

pkg/server/service.go
```diff
@@ -90,6 +90,11 @@ func newGRPCServer(ctx context.Context, pluginRegistry *plugins.Registry, cfg *c
 			auth.AuthenticationLoggingInterceptor,
 			middlewareInterceptors,
 		)
+		if cfg.Security.RateLimit.Enabled {
+			rateLimiter := plugins.NewRateLimiter(cfg.Security.RateLimit.RequestsPerSecond, cfg.Security.RateLimit.BurstSize, cfg.Security.RateLimit.CleanupInterval.Duration)
+			rateLimitInterceptors := plugins.RateLimiteInterceptor(*rateLimiter)
+			chainedUnaryInterceptors = grpcmiddleware.ChainUnaryServer(chainedUnaryInterceptors, rateLimitInterceptors)
+		}
 	} else {
```
LaPetiteSouris, line 98: This part is uncovered by integration tests. Looks like all parts related to auth middle ware is not yet covered by integration tests, which  is understandable as it is not straight forward to implement OAuth flow in integration tests.

I am not sure how to improve it though...

EngHabu (commented):
Thank you for your contribution!

pkg/config/config.go
LaPetiteSouris: Resolved as inline docs is updated

pkg/server/service.go
```diff
@@ -90,6 +90,11 @@ func newGRPCServer(ctx context.Context, pluginRegistry *plugins.Registry, cfg *c
 			auth.AuthenticationLoggingInterceptor,
 			middlewareInterceptors,
 		)
+		if cfg.Security.RateLimit.Enabled {
```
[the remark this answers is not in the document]
LaPetiteSouris: Fixed

pkg/server/service.go
LaPetiteSouris: manually tested, and looks like the unit test for RateLimiter is sufficient

plugins/rate_limit.go
```diff
@@ -0,0 +1,112 @@
...
+type RateLimitExceeded error
+
+// define a struct that contains a map of rate limiters, and a time stamp of last access and a mutex to protect the map
+type accessRecords struct {
+	limiter    *rate.Limiter
+	lastAccess time.Time
+}
+
+type LimiterStore struct {
+	accessPerUser   map[string]*accessRecords
```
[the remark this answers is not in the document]
LaPetiteSouris: Unfortunately it is rater complicate, as `accessRecords` holds `UserID` as key and a `rate` struct as object.

The `rate` struct is mutable, i.e: it gets updated every time we call `Allow` function, thus bypass the `Store` method, if a `sync.Map` is used.

Do you have a strong opinion on `sync.Map` or a normal map with `mutex` would be enough for now ?

pkg/config/config.go
```diff
@@ -80,14 +80,23 @@ type ServerSecurityOptions struct {
 	// These are the Access-Control-Request-Headers that the server will respond to.
 	// By default, the server will allow Accept, Accept-Language, Content-Language, and Content-Type.
 	// DeprecatedUser this setting to add any additional headers which are needed
-	AllowedHeaders []string `json:"allowedHeaders"`
+	AllowedHeaders []string         `json:"allowedHeaders"`
+	RateLimit      RateLimitOptions `json:"rateLimitOptions"`
```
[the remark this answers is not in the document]
LaPetiteSouris: fixed

pkg/config/config.go
```diff
@@ -80,14 +80,23 @@ type ServerSecurityOptions struct {
...
 	KeyFile         string `json:"keyFile"`
 }
 
+// declare RateLimitConfig
+type RateLimitOptions struct {
+	Enabled           bool            `json:"enabled"`
+	RequestsPerSecond int             `json:"requestsPerSecond"`
+	BurstSize         int             `json:"burstSize"`
+	CleanupInterval   config.Duration `json:"cleanupInterval"`
+}
```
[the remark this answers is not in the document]
LaPetiteSouris: fixed

plugins/rate_limit.go
```diff
@@ -0,0 +1,112 @@
...
+	mutex           *sync.Mutex
+	requestPerSec   int
+	burstSize       int
+	cleanupInterval time.Duration
+}
+
+// define a function named Allow that takes userID and returns RateLimitError
```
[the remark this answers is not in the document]
LaPetiteSouris: fixed

pkg/config/config.go
```diff
@@ -80,14 +80,23 @@ type ServerSecurityOptions struct {
...
+	AllowedHeaders []string         `json:"allowedHeaders"`
+	RateLimit      RateLimitOptions `json:"rateLimitOptions"`
 }
 
 type SslOptions struct {
 	CertificateFile string `json:"certificateFile"`
 	KeyFile         string `json:"keyFile"`
 }
 
+// declare RateLimitConfig
```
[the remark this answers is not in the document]
LaPetiteSouris: fixed

plugins/rate_limit.go
```diff
@@ -0,0 +1,113 @@
...
+	accessPerUser   map[string]*accessRecords
+	mutex           *sync.Mutex
+	requestPerSec   int
+	burstSize       int
+	cleanupInterval time.Duration
+}
+
+// Allow takes a userID and returns an error if the user has exceeded the rate limit
+func (l *LimiterStore) Allow(userID string) error {
+	l.mutex.Lock()
```
kumare3, line 37: This will cause massive contention - in effect every request is serialized for this block. I think the rate object should be accessed in parallel and then the lock should be on the lower level object.
Sync map would work right.
Otherwise this is probably a very dangerous PR

kumare3 (changes requested):

plugins/rate_limit.go
LaPetiteSouris: Thanks @kumare3 and @EngHabu for the suggestion on `sync.Map`

In https://github.com/flyteorg/flyteadmin/pull/557/commits/b35114b3e55b76402fd49ba738d5599f8175e57f , a `sync.Map` is used and the lock is applied only per `UserID`. This mitigates the concern on contention, which is also the main interest of `sync.Map`

However, a "global" lock is kept at `LimitStore` level to serve the `clean` func, which periodically purge unused entries to reduce memory footprint.

plugins/rate_limit.go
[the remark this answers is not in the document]
LaPetiteSouris: I resolved this conversation, as the issue is implemented and further discussion is tracked in the discussion below

plugins/rate_limit.go
LaPetiteSouris: Second thoughts, actually only a single lock is needed, down to the level of `rate` object. `sync.Map` already takes care of the hard part, so all `rate` object can be accessed in parallel, even in case of `clean`.

The latest commit https://github.com/flyteorg/flyteadmin/pull/557/commits/b2383af3ac0233e2962860eb8f0c67c97d0dd880 reflected the changes. Thanks so much for spotting this out. I definitely learn a new thing today. 👍

LaPetiteSouris:
Would anyone be available to re-review my PR or may be at least approve the workflows so that I have visibility on what are the remaining to fix/improve ?

Thanks a lot.

LaPetiteSouris:
Thanks for launching the CI/CD for me. The `Integration tests`[ phase failed unexpectedly](https://github.com/flyteorg/flyteadmin/actions/runs/4945206998/jobs/8884029871), I suppose due to some error really unrelated to the PR itself. Is this a flaky tests, related to CI/CD set up, or am I wrong ?

eapolinario:
> Thanks for launching the CI/CD for me. The `Integration tests`[ phase failed unexpectedly](https://github.com/flyteorg/flyteadmin/actions/runs/4945206998/jobs/8884029871), I suppose due to some error really unrelated to the PR itself. Is this a flaky tests, related to CI/CD set up, or am I wrong ?

Correct. Fixing the flaky test in https://github.com/flyteorg/flyteadmin/pull/563.

LaPetiteSouris:
> > Thanks for launching the CI/CD for me. The `Integration tests`[ phase failed unexpectedly](https://github.com/flyteorg/flyteadmin/actions/runs/4945206998/jobs/8884029871), I suppose due to some error really unrelated to the PR itself. Is this a flaky tests, related to CI/CD set up, or am I wrong ?
> 
> Correct. Fixing the flaky test in #563.

Thanks. I synced the PR with `master`. Need reviews.

eapolinario:
@LaPetiteSouris , can you merge master instead? This last update brought commits that don't have to do with your changes.

LaPetiteSouris:
@eapolinario Sorry for that. I did not pay attention.

I rebase to `master` to have proper commit set up.

eapolinario (commented):
If I understand correctly, we're essentially having a version of sync.Map with TTL. So just to make the code more readable, how about we move that logic behind a new type that does only that?

plugins/rate_limit.go
```diff
@@ -0,0 +1,116 @@
...
+	})
+	accessRecord.(*accessRecords).mutex.Lock()
+	defer accessRecord.(*accessRecords).mutex.Unlock()
+
+	accessRecord.(*accessRecords).lastAccess = time.Now()
+	l.accessPerUser.Store(userID, accessRecord)
+
+	if !accessRecord.(*accessRecords).limiter.Allow() {
+		return RateLimitExceeded(fmt.Errorf("rate limit exceeded"))
+	}
```
eapolinario, line 50: shouldn't this happen before we modify the map?

PR #565, opened 2023-05-17
Tiny url improvements

wild-endeavor:
# TL;DR
This adds the ability to return an individual input or output in the flyte url, e.g.
```
flyte://v1/flytesnacks/development/ad5cb55f675g2s78swgp/end-node/i/o0
```

There's currently no mechanism by which these URLs can be returned to the frontend however.  This can be worked on separately.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
* Refactored out the parsing logic to return a union instead.  go doesn't have unions so returning a struct.
* Split out the GetData function into one for node level and one for task execution level.
* Added a second regex object to try to match against.

## Tracking Issue
NA

### DIFF
--- a/dataproxy/service.go
+++ b/dataproxy/service.go
@@ -396,13 +396,13 @@ func (s Service) GetData(ctx context.Context, req *service.GetDataRequest) (
 	}
 
 	if executions.NodeExecID != nil {
-		return s.GetDataFromNodeExecution(ctx, *executions.NodeExecID, executions.IOType, executions.OutputName)
+		return s.GetDataFromNodeExecution(ctx, *executions.NodeExecID, executions.IOType, executions.LiteralName)
 	} else if executions.PartialTaskExecID != nil {
 		taskExecID, err := s.GetCompleteTaskExecutionID(ctx, *executions.PartialTaskExecID)
 		if err != nil {
 			return nil, err
 		}
-		return s.GetDataFromTaskExecution(ctx, *taskExecID, executions.IOType, executions.OutputName)
+		return s.GetDataFromTaskExecution(ctx, *taskExecID, executions.IOType, executions.LiteralName)
 	}
 
 	return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument, "failed to parse get data request %v", req)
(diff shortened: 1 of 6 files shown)


pkg/common/flyte_url_test.go
```diff
@@ -172,3 +173,110 @@ func TestMatchRegexDirectly(t *testing.T) {
...
+	t.Run("regex with specific output no attempt", func(t *testing.T) {
+		matches := MatchRegex(reSpecificOutput, "flyte://v1/fs/dev/abc/n0/o/o0")
+		fmt.Println(matches)
+
+	})
+
+	t.Run("regex with specific output no attempt", func(t *testing.T) {
+		matches := MatchRegex(reSpecificOutput, "flyte://v1/fs/dev/abc/n0-dn0-9-n0-n0/o/o0")
+		fmt.Println(matches)
+	})
```
eapolinario, line 187: worth asserting in these tests?

eapolinario (dismissed):
one minor comment, otherwise, LGTM.

pkg/common/flyte_url_test.go
wild-endeavor: done

pingsutw (approved):

dataproxy/service.go
```diff
@@ -347,11 +361,51 @@ func (s Service) GetData(ctx context.Context, req *service.GetDataRequest) (
...
+func (s Service) GetData(ctx context.Context, req *service.GetDataRequest) (
+	*service.GetDataResponse, error) {
+
+	logger.Debugf(ctx, "resolving flyte url query: %s", req.GetFlyteUrl())
+	err := s.validateResolveArtifactRequest(req)
+	if err != nil {
+		return nil, errors.NewFlyteAdminErrorf(codes.InvalidArgument, "failed to validate resolve artifact request. Error: %v", err)
+	}
+
+	executions, err := common.ParseFlyteURLToExecution(req.GetFlyteUrl())
```
eapolinario, line 393: nit: why is this in the plural?

eapolinario (dismissed):
LGTM

eapolinario (approved):

PR #567, opened 2023-05-19
Update startedAt timestamp only if not set

pmahindrakar-oss:
# TL;DR
Updation of the startedAt timestamp for same taskExecution happens in case of map tasks.
So when propeller send the TaskExecutionEvents for each mappedTask, admin records the event and if it finds it to be Running, then calls the addTaskStartedState function which update the startedAt timestamp.

This is fine to do in case of regular task but incase of mapped task, each mappedTask status causes an update to startedAt which is incorrect.
This PR fixes it by guarding it with if the values is not set

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/3702

## Follow-up issue
_NA_
OR

### DIFF
--- a/pkg/repositories/transformers/task_execution.go
+++ b/pkg/repositories/transformers/task_execution.go
@@ -5,25 +5,22 @@ import (
 	"sort"
 	"strconv"
 
-	"github.com/flyteorg/flyteadmin/pkg/runtime/interfaces"
-	"github.com/flyteorg/flytestdlib/storage"
-
-	"google.golang.org/protobuf/encoding/protojson"
-
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
 	_struct "github.com/golang/protobuf/ptypes/struct"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/flyteorg/flyteadmin/pkg/common"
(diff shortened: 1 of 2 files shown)


pmahindrakar-oss:
Integration tests failing on unrelated changes to this PR. Previous PR's aswell failing on the same issue

katrogan (approved):

hamersaw (approved):
An alternative solution that would fix the existing conditional rather than just add another is to update [this line](https://github.com/flyteorg/flyteadmin/blob/313ac294ac18fd70f90e9605e6a3f5544b69427c/pkg/repositories/transformers/task_execution.go#L401) to something like:
```
if existingTaskPhase != core.TaskExecution_RUNNING.String() && taskExecutionModel.StartedAt == nil {
```
I am OK with either fix.

pmahindrakar-oss:
@hamersaw the one i implemented i found it more unit testable aswell as since we are using this function at multiple places so wanted to make sure we define the behavior in the function itself to avoid any future calls causing the same regression

https://github.com/flyteorg/flyteadmin/blob/313ac294ac18fd70f90e9605e6a3f5544b69427c/pkg/repositories/transformers/task_execution.go#L174
I will use the current approach for that reason. Let me know if you disagree

PR #568, opened 2023-05-22
Save execution namespace in system metadata

katrogan:
# TL;DR
Save execution namespace in system metadata

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
N/A see summary

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/3705

## Follow-up issue
_NA_

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/cloudevents/sdk-go/v2 v2.8.0
 	github.com/coreos/go-oidc v2.2.1+incompatible
 	github.com/evanphx/json-patch v4.12.0+incompatible
-	github.com/flyteorg/flyteidl v1.5.5
+	github.com/flyteorg/flyteidl v1.5.7-0.20230522164738-0484cf8e75cc
 	github.com/flyteorg/flyteplugins v1.0.56
 	github.com/flyteorg/flytepropeller v1.1.87
 	github.com/flyteorg/flytestdlib v1.0.15
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/cloudevents/sdk-go/v2 v2.8.0
 	github.com/coreos/go-oidc v2.2.1+incompatible
 	github.com/evanphx/json-patch v4.12.0+incompatible
-	github.com/flyteorg/flyteidl v1.5.7-0.20230522164738-0484cf8e75cc
+	github.com/flyteorg/flyteidl v1.5.7
 	github.com/flyteorg/flyteplugins v1.0.56
 	github.com/flyteorg/flytepropeller v1.1.87
 	github.com/flyteorg/flytestdlib v1.0.15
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/cloudevents/sdk-go/v2 v2.8.0
 	github.com/coreos/go-oidc v2.2.1+incompatible
 	github.com/evanphx/json-patch v4.12.0+incompatible
-	github.com/flyteorg/flyteidl v1.5.5
+	github.com/flyteorg/flyteidl v1.5.7-0.20230522164738-0484cf8e75cc
```
katrogan, line 16: will rebase before merging

pkg/manager/impl/execution_manager_test.go
```diff
@@ -287,6 +287,7 @@ func TestCreateExecution(t *testing.T) {
 			assert.Equal(t, rawOutput, spec.RawOutputDataConfig.OutputLocationPrefix)
 			assert.True(t, proto.Equal(spec.ClusterAssignment, &clusterAssignment))
 			assert.Equal(t, "launch_plan", input.LaunchEntity)
+			assert.Equal(t, spec.GetMetadata().GetSystemMetadata().Namespace, "project-domain")
```
pmahindrakar-oss, line 290: can we add one test with namespace mapping which equals just  `project`

pkg/manager/impl/execution_manager.go
```diff
@@ -650,6 +650,7 @@ func (m *ExecutionManager) launchSingleTaskExecution(
 		UserInputsURI:         userInputsURI,
 		SecurityContext:       executionConfig.SecurityContext,
 		LaunchEntity:          taskIdentifier.ResourceType,
+		Namespace:             namespace,
```
pmahindrakar-oss, line 653: Can we also make a backward compat change for older executions which dont have this namespace metadata set that it returns the project-domain which is widely used namespace

pmahindrakar-oss (approved):

pkg/manager/impl/execution_manager.go
katrogan: I think we should respect the config if we're going to assume values rather than use the default mapping

pmahindrakar-oss: Yeah that works. wanted to make sure we have some data that we have valid data when we call the GET api even though the model wont have saved this when creating the execution

pkg/manager/impl/execution_manager_test.go
katrogan: The code is already testing that we're exercising namespace mapping config but added another test

katrogan:
PTAL @pmahindrakar-oss

PR #571, opened 2023-06-01
Flyte Execution tags

pingsutw:
# TL;DR
- List: http://localhost:30082/api/v1/executions/flytesnacks/development/f6e2ba11a96c24e79a88
- List with filter: http://localhost:30082/api/v1/executions/flytesnacks/development?filters=value_in(admin_tag.name,flyte%3Btest)&limit=25&sort_by.direction=DESCENDING&sort_by.key=created_at
- Get: http://localhost:30082/api/v1/executions/flytesnacks/development?limit=5

Add two tables (admin_tags, execution_admin_tags). 
`execution_admin_tags` is a join table containing admin_tags's ID and execution's ID. 
`admin_tags` save all the tag names.

To list executions that match specific tags, we will join the `execution`, `execution_admin_tags`, and `admin_tags` tables.

- execution_admin_tags

| Name    | Type |
| --------- | ------- |
| project (foreign key)  |   string  |
| domain (foreign key    |   string  |
| execution name (foreign key)  |   string  |
|tag_id (foreign key) | unit |

- admin_tags

| Name    | Type |
| --------- | ------- |
| id (primary key)   |  unit  |
| tag name (unique index)    |   string  (unique) |
| created_at   |   time.Time  |
| updated_at   |   time.Time  |
| deleted_at | time.Time |

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
image

## Tracking Issue
https://github.com/flyteorg/flyte/pull/3320
Blocked by https://github.com/flyteorg/flyteidl/pull/414

## Follow-up issue
_NA_


pkg/repositories/config/migrations.go
```diff
@@ -445,6 +445,26 @@ var LegacyMigrations = []*gormigrate.Migration{
 			return tx.Model(&models.Execution{}).Migrator().DropColumn(&models.Execution{}, "launch_entity")
 		},
 	},
+	// Create admin tags table.
+	{
+		ID: "2023-06-01-admin_tags",
+		Migrate: func(tx *gorm.DB) error {
+			return tx.AutoMigrate(&models.AdminTag{})
```
wild-endeavor, line 452: nope not this.  never again hahaha

pkg/repositories/config/migrations.go
```diff
@@ -445,6 +445,26 @@ var LegacyMigrations = []*gormigrate.Migration{
...
+		},
+		Rollback: func(tx *gorm.DB) error {
+			return tx.Migrator().DropTable("admin_tags")
+		},
+	},
+	// Add execution <-> admin_tags join table.
+	{
+		ID: "2023-06-01-execution_admin_tags",
+		Migrate: func(tx *gorm.DB) error {
+			return tx.AutoMigrate(&models.Execution{})
```
wild-endeavor, line 462: ask eduardo

pkg/repositories/config/migrations.go
eapolinario: yeah, let's not do this. The problem with using model files directly is that we can't track changes done to the underlying model (think of the case users are updating from different versions of flyteadmin). 

The template we've been using in this file is to declare the model to be migrated inline, e.g. https://github.com/flyteorg/flyteadmin/blob/master/pkg/repositories/config/migrations.go#L897-L920. Notice how the `ExecutionEvent` model is defined in the same scope as the migration.

eapolinario (commented):
Looking at the [RFC](https://github.com/flyteorg/flyte/pull/3320) we said we were going to have separate keys and values for labels, whereas the implementation in this PR we use flat values as the tags, right? It doesn't invalidate the overall approach (of having a many-to-many relationship) though.

pkg/repositories/models/execution.go
```diff
@@ -60,4 +61,11 @@ type Execution struct {
 	State *int32 `gorm:"index;default:0"`
 	// The resource type of the entity used to launch the execution, one of 'launch_plan' or 'task'
 	LaunchEntity string
+
+	Tags []AdminTag `gorm:"many2many:execution_admin_tags;"`
+}
+
+type AdminTag struct {
+	gorm.Model
+	Name string `valid:"length(0|255)"`
```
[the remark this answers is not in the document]
pingsutw: Updated it.

After some discussions in the community, we decide to use key-only tags first. If we want to use key-value pair tags in the feature, we can add a new column in the tags table and update the query in the `execution_repo.go`

pkg/repositories/config/migrations.go
```diff
@@ -1084,6 +1084,88 @@ var NoopMigrations = []*gormigrate.Migration{
 		},
 	},
 
+	{
+		ID: "pg-noop-2023-06-07-admin-tags",
```
eapolinario, line 1088: Drop the `pg-noop-` prefix, this was referring to the migrations added in https://github.com/flyteorg/flyteadmin/pull/542.

Also, can you move these migrations to the last position in the list of migrations? (Remember to update the date in the name).

pkg/repositories/models/execution.go
```diff
@@ -60,4 +62,11 @@ type Execution struct {
 	State *int32 `gorm:"index;default:0"`
 	// The resource type of the entity used to launch the execution, one of 'launch_plan' or 'task'
 	LaunchEntity string
+	// Tags associated with the execution
+	Tags []AdminTag `gorm:"many2many:execution_admin_tags;"`
+}
+
+type AdminTag struct {
+	gorm.Model
+	Name string `valid:"length(0|255)"`
```
eapolinario, line 71: you can drop the `valid` tag as that is not really used anywhere.

pkg/repositories/transformers/execution.go
```diff
@@ -119,6 +124,7 @@ func CreateExecutionModel(input CreateExecutionModelInput) (*models.Execution, e
 		User:                  requestSpec.Metadata.Principal,
 		State:                 &activeExecution,
 		LaunchEntity:          strings.ToLower(input.LaunchEntity.String()),
+		Tags:                  tags,
```
eapolinario, line 127: is it worth putting the writes behind a feature flag?

eapolinario (commented):
Looking good, just a couple of comments.

pkg/repositories/config/migrations.go
pingsutw: updated

pkg/repositories/models/execution.go
pingsutw: updated

pkg/repositories/transformers/execution.go
pingsutw: I feel like it's unnecessary because we won't add a new row to the `tags` table if users don't pass any tags to the execution Spec.

pkg/repositories/transformers/execution.go
```diff
@@ -99,6 +99,11 @@ func CreateExecutionModel(input CreateExecutionModelInput) (*models.Execution, e
 	}
 
 	activeExecution := int32(admin.ExecutionState_EXECUTION_ACTIVE)
+	tags := make([]models.AdminTag, len(input.RequestSpec.Tags))
+	for i, tag := range input.RequestSpec.Tags {
+		tags[i] = models.AdminTag{Name: tag}
+	}
```
eapolinario, line 105: This will end up creating separate rows in the tags table that contain the same value, right? As in, if I create two executions in which both set the tag "tagA", then we'll end up creating two rows in the Tags table with the name "tagA", correct?

pkg/repositories/gormimpl/execution_repo_test.go
```diff
@@ -270,6 +270,36 @@ func TestListExecutions_Order(t *testing.T) {
 	assert.True(t, mockQuery.Triggered)
 }
 
+func TestListExecutions_WithTags(t *testing.T) {
```
eapolinario, line 273: can you add a test where more than one execution is returned? For example, let's say we have:
- execution1 and tags T1 and T2
- execution2 and only tag T1
- execution3 and only tag T2

Can we have a test where the query for executions labeled with tag T1 returns execution1 and execution2, etc?

pkg/repositories/models/execution.go
```diff
@@ -60,4 +62,11 @@ type Execution struct {
 	State *int32 `gorm:"index;default:0"`
 	// The resource type of the entity used to launch the execution, one of 'launch_plan' or 'task'
 	LaunchEntity string
+	// Tags associated with the execution
+	Tags []AdminTag `gorm:"many2many:execution_admin_tags;"`
```
eapolinario, line 66: can you add the full schema of the new tables to the PR description?

eapolinario (commented):
I'm wondering if we have to worry about storage requirements at this point, so left a comment related to this.

Also, not sure if the queries to list executions are correct, so a couple comments in that area as well.

pkg/repositories/transformers/execution.go
pingsutw: I just set the `tag_name` column as unique so we won't create duplicate tags.

pkg/repositories/gormimpl/execution_repo_test.go
pingsutw: We use mock query, so actually we won't run the query in the database.

pkg/repositories/models/execution.go
pingsutw: updated it

pkg/repositories/gormimpl/execution_repo_test.go
eapolinario: Can you add an integration test to https://github.com/flyteorg/flyteadmin/blob/master/tests/execution_test.go?

pkg/repositories/models/execution.go
eapolinario: Can you add the example schemas containing the foreign keys not the go models?

pkg/repositories/transformers/execution.go
eapolinario: do we have to worry about fetching those tags? What happens if we try to save an execution that contains a new tag that's a duplicate? Does this cause an error?

pingsutw: No, it doesn't cause an error, and won't create a new row having the same tag.

pkg/repositories/gormimpl/execution_repo_test.go
pingsutw: added it

pkg/repositories/models/execution.go
pingsutw: Sorry, could you say more? What else should I add to the table schema in the PR description

eapolinario: oh, I meant the actual schema of the database tables created. I see that the description mentions foreign keys and primary keys, so adding the schemas is not really necessary. Thanks!

tests/execution_test.go
```diff
@@ -209,6 +217,26 @@ func TestListWorkflowExecutions(t *testing.T) {
 	assert.Equal(t, len(resp.Executions), 4)
 }
 
+func TestListWorkflowExecutionsWithTags(t *testing.T) {
+	truncateAllTablesForTestingOnly()
```
eapolinario, line 221: should we update [this](https://github.com/flyteorg/flyteadmin/blob/master/tests/bootstrap.go#L105) function to also truncate the newly added tables?

pkg/repositories/transformers/execution.go
eapolinario: The integration test shows a few errors: https://github.com/flyteorg/flyteadmin/actions/runs/5717905981/job/15492760870?pr=571#step:7:346.

tests/execution_test.go
```diff
@@ -209,6 +217,26 @@ func TestListWorkflowExecutions(t *testing.T) {
...
+		Id: &admin.NamedEntityIdentifier{
+			Project: "project1",
+			Domain:  "domain1",
+		},
+		Limit:   5,
+		Filters: "value_in(admin_tag.name, hello)",
+	})
+	assert.Nil(t, err)
+	assert.Equal(t, len(resp.Executions), 2)
+}
```
eapolinario, line 238: given that's not too difficult to write integration tests, can we validate what happens when duplicate labels are used as part of creating an execution?

eapolinario (commented):
Just a few comments on how to handle the creation of tags (and potential duplicates).

tests/execution_test.go
pingsutw: It's hard to do that here since we must register a workflow first. In these integration tests, we didn't really create any real workflow in the database.

pkg/repositories/config/migrations.go
```diff
@@ -1097,6 +1097,89 @@ var NoopMigrations = []*gormigrate.Migration{
...
+
+			return tx.AutoMigrate(&AdminTag{})
+		},
+		Rollback: func(tx *gorm.DB) error {
+			return nil
+		},
+	},
+
+	{
+		ID: "2023-07-20-noop-execution-admin-tags",
```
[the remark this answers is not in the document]
pingsutw: done

pkg/repositories/config/migrations.go
```diff
@@ -1097,6 +1097,89 @@ var NoopMigrations = []*gormigrate.Migration{
...
+
+			return tx.AutoMigrate(&AdminTag{})
+		},
+		Rollback: func(tx *gorm.DB) error {
+			return nil
+		},
+	},
+
+	{
+		ID: "2023-08-04-noop-migration-execution-admin-tags",
```
eapolinario, line 1117: can we drop the name `noop` from this migration? This migration creates the join table, so maybe add that to the name or at the very least leave a comment.

pingsutw: done

pkg/repositories/models/execution.go
```diff
@@ -60,4 +62,11 @@ type Execution struct {
...
+}
+
+type AdminTag struct {
+	gorm.Model
+	Name string `gorm:"index:,unique;size:255"`
```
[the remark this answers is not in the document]
pingsutw: We don't need it since we'll only list executions.

eapolinario (approved):

PR #573, opened 2023-06-02
Add logs and stats to sync loop

EngHabu:
# TL;DR
Add info logging for when a sync loop starts and ends and add stats at the end of the sync loop to the log.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/3743

### DIFF
--- a/pkg/clusterresource/controller.go
+++ b/pkg/clusterresource/controller.go
@@ -607,10 +607,12 @@ func (c *controller) Sync(ctx context.Context) error {
 				logger.Debugf(ctx, "Successfully created kubernetes resources for [%s]", namespace)
 				stats.Add(newStats)
 			}
+
+			logger.Infof(ctx, "Completed cluster resource creation loop for namespace [%s] with stats: [%+v]", namespace, newStats)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/clusterresource/controller.go
```diff
@@ -573,40 +585,50 @@ func (c *controller) Sync(ctx context.Context) error {
...
 				errs = append(errs, err)
 			} else {
 				c.metrics.ResourcesAdded.Inc()
 				logger.Debugf(ctx, "Successfully created kubernetes resources for [%s]", namespace)
+				stats.Add(newStats)
 			}
 		}
 	}
+
+	logger.Infof(ctx, "Completed cluster resource creation loop with stats: %+v", stats)
```
katrogan, line 613: Do you think it might be useful to log these stats per namespace as well too, since it's easier to reason about the deltas in created/updated at the template level than remembering they get applied n times across all namespaces?

katrogan (approved):

katrogan (approved):

PR #577, opened 2023-06-15
Add a user-specifiable root to upload link request

wild-endeavor:
# TL;DR
This is useful for uploading folders.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/dataproxy/service.go
+++ b/dataproxy/service.go
@@ -77,13 +77,13 @@ func (s Service) CreateUploadLocation(ctx context.Context, req *service.CreateUp
 	}
 
 	md5 := base64.StdEncoding.EncodeToString(req.ContentMd5)
-	urlSafeMd5 := base32.StdEncoding.EncodeToString(req.ContentMd5)
 
 	var prefix string
 	if req.FilenameRoot != "" {
 		prefix = req.FilenameRoot
 	} else {
-		prefix = urlSafeMd5
+		// url safe base32 encoding
+		prefix = base32.StdEncoding.EncodeToString(req.ContentMd5)
 	}
 	storagePath, err := createStorageLocation(ctx, s.dataStore, s.cfg.Upload,
 		req.Project, req.Domain, prefix, req.Filename)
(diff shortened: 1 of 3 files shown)


dataproxy/service.go
```diff
@@ -77,10 +77,16 @@ func (s Service) CreateUploadLocation(ctx context.Context, req *service.CreateUp
 	}
 
 	md5 := base64.StdEncoding.EncodeToString(req.ContentMd5)
-	urlSafeMd5 := base32.StdEncoding.EncodeToString(req.ContentMd5)
 
+	var prefix string
+	if req.FilenameRoot != "" {
+		prefix = req.FilenameRoot
+	} else {
```
EngHabu, line 84: One reason the previous path was "safe" without RBAC is that even if two people try to upload to the same path, unless they are uploading the same file, they won't end up with conflicting content in the same location...

I understand the motivation behind the change and maintaining a consistent structure...
Would it make sense to maintain the safety here? 
Perhaps it's a client-side driven logic to include the SHA of the ordered SHAs of all files of a directory as part of the FilenameRoot? 
so you end up with:
s3://<something>/<SHAofSHAs>/dir/name/path/filename.foo

Perhaps the SHAofSHAs logic is something like:
build a json of [MD5] = filename
```json
"dir": {
  "filename": <md5>,
...
}
```

Then computing the SHA of the json string?

wild-endeavor: so do this computation on the client side? or are you saying change the idl to take a json string, and admin hashes the json string?

EngHabu: client side yes... I think it'll need a whole lot more changes to incorporate that behavior on the server side...
On top of my head, a new API that's built for uploading directories:
```
rpc UploadDirectory(stream UploadDirectoryRequest) UploadDirectoryResponse

message UploadDirectoryRequest {
    string directory_md5 = 1;
    string filename = 2;
    bytes file_contents = 3;
}
```

or something like the Multipart upload semantics
```
rpc StartUploadDirectory(StartUploadDirectoryRequest) StartUploadDirectoryResponse
rpc UploadFile(UploadFileRequest) UploadFileResponse
rpc TerminateUploadDirectory(TerminateUploadDirectoryRequest) TerminateUploadDirectoryResponse
```

And then the server can continue to generate/enforce SHAs of directories... and fail the call if SHA doesn't match or if the client fails to complete the upload within a specified time...

EngHabu (dismissed):

EngHabu (approved):

eapolinario (approved):

PR #578, opened 2023-06-19
Alter ID sequence to bigint

honnix:
# TL;DR
Alter ID sequence to bigint.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

I tested the sql on one of our sequences and it worked as expected.

## Complete description
As described by https://github.com/flyteorg/flyte/issues/3777, some of the ID sequences are not bigint due to historical reason. It makes sense to use bigint because integer is easy to run out of.

Note that this PR does not try to change `cycle` in order to keep backward compatibility, and also `bigint` will practically make that unnecessary.

For reference https://www.postgresql.org/docs/current/sql-altersequence.html

## Tracking Issue
Fixes https://github.com/flyteorg/flyte/issues/3777

## Follow-up issue
_NA_

### DIFF
--- a/pkg/repositories/config/migrations.go
+++ b/pkg/repositories/config/migrations.go
@@ -1083,16 +1083,39 @@ var NoopMigrations = []*gormigrate.Migration{
 			return nil
 		},
 	},
+
+	{
+		ID: "2023-06-19-id-sequence-to-bigint",
+		Migrate: func(tx *gorm.DB) error {
+			db, err := tx.DB()
+			if err != nil {
+				return err
+			}
+			return alterIDSequenceType(db)
+		},
+		Rollback: func(tx *gorm.DB) error {
+			return nil
+		},
+	},
 }
 
 var Migrations = append(LegacyMigrations, NoopMigrations...)
 
 func alterTableColumnType(db *sql.DB, columnName, columnType string) error {
-
 	var err error
 	for _, table := range tables {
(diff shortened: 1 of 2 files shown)


kumare3:
Cc @eapolinario

eapolinario:
@honnix , one concern with this change is the possibility of blocking the tables where sequences are not bigints already. We want to minimize downtime, esp. in the case of large tables like `task_executions` and `node_executions`. 

In the recent past we [coordinated](https://flyte-org.slack.com/archives/C03CYSL783B/p1681382011320999) tests with other members of the Spotify team. Would it be possible to run a similar test on your side? Sorry for the inconvenience, but database migrations require this extra care.

honnix:
@eapolinario Good question. We have tested this on our production system and the change took fraction of second. It doesn't do anything to existing sequences, but only changes a single row of the sequence table.

honnix:
It is worth noting that all the `ID` columns in different tables are all `bigint` already, at least in our production database that has been evolving through the past couple of years. It's only the sequence type being inconsistent across different sequences.

eapolinario (approved):
Thanks for explaining. In that case, this PR looks good.

honnix:
Thank you! Please help merge it when suitable.

PR #587, opened 2023-07-10
Remove content md5 requirement

wild-endeavor:
# TL;DR
Remove the requirement that content md5 is provided.

Depends on https://github.com/flyteorg/flytestdlib/pull/160/

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
If the full native path is fully specified, not random, then we check to see if it exists.
If it exists then the user needs to have given a hash, which admin will check against the etags.
etags are computed differently for different stores, so we try a few different ways.
if the tag doesn't match, then we error.  if no hash was given then also error.

## Tracking Issue
This is related to https://github.com/flyteorg/flytekit/pull/1674/


dataproxy/service.go
```diff
@@ -51,12 +52,45 @@ type Service struct {
...
+		knownLocation, err := createStorageLocation(ctx, s.dataStore, s.cfg.Upload,
+			req.Project, req.Domain, req.FilenameRoot, req.Filename)
+		if err != nil {
+			return nil, errors.NewFlyteAdminErrorf(codes.Internal, "failed to create storage location, Error: %v", err)
+		}
+		metadata, err := s.dataStore.Head(ctx, knownLocation)
+		if err != nil {
+			return nil, errors.NewFlyteAdminErrorf(codes.Internal, "failed to check if file exists at location [%s], Error: %v", knownLocation.String(), err)
+		}
+		if metadata.Exists() {
```
EngHabu, line 80: Let's add a comment that this is best effort and acknowledge the racing issue?

EngHabu (approved):

wild-endeavor:
@EngHabu bumped stdlib and added comment.... one more look whenever you get a chance please?

EngHabu (approved):

PR #588, opened 2023-07-13
Fix functional tests

eapolinario:
# TL;DR
Use restricted set of tests in functional tests

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Use a subset of tests on the demo cluster in functional tests.

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -27,10 +27,9 @@ jobs:
   endtoend:
     name: End2End Test
     needs: [ docker-build ]
-    uses: flyteorg/flytetools/.github/workflows/end2end.yml@master
+    uses: flyteorg/flytetools/.github/workflows/end2end.yml@fix-functional-tests
     with:
       priorities: "P0"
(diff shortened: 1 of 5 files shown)


eapolinario:
This is the wrong approach. We need the flexibility of replacing components prior to running the tests. The way we do this now is by relying on the locally built component images and injecting them into the old sandbox as part of the process of running tests. 

The ultimate goal is to fully deprecate the old sandbox, but in order to do that we need the capability described in the previous paragraph. This is one more datapoint to expedite the move to a monorepo, where it'll be easy to setup these locally built single binary images containing updated code.

eapolinario:
Functional tests passed: https://github.com/flyteorg/flyteadmin/actions/runs/5557184259/jobs/10150754879?pr=588. Will replicate this fix everywhere now.

PR #376, opened 2022-03-22
Read and pass along cluster assignment to workflow executor

katrogan:
# TL;DR
Read cluster assignment attribute from the execution spec. Cluster assignment can now be designated by matching against preferred affinity selectors

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added
 - [ ] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description
 _How did you fix the bug, make the feature etc. Link to any design docs etc_

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2248

## Follow-up issue
_NA_

PR #600, opened 2023-08-10
Add default cluster pool assignments to config

katrogan:
# TL;DR
Add default cluster pool assignments as application config options

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue

https://github.com/flyteorg/flyte/issues/2248

## Follow-up issue
_NA_


pkg/runtime/cluster_pool_assignment_provider.go
```diff
@@ -0,0 +1,24 @@
+package runtime
+
+import (
+	"github.com/flyteorg/flyteadmin/pkg/runtime/interfaces"
+
+	"github.com/flyteorg/flytestdlib/config"
+)
+
+const clusterPoolsKey = "cluster_pools"
```
pmahindrakar-oss, line 9: Dont we use camel case every where for the keys. Would be good to keep this consistent with other configs

pmahindrakar-oss (approved):

pkg/manager/impl/execution_manager.go
```diff
@@ -444,8 +444,11 @@ func (m *ExecutionManager) getClusterAssignment(ctx context.Context, request *ad
 	if resource != nil && resource.Attributes.GetClusterAssignment() != nil {
 		return resource.Attributes.GetClusterAssignment(), nil
 	}
-	// Defaults to empty assignment with no selectors
-	return &admin.ClusterAssignment{}, nil
+	clusterPoolAssignment := m.config.ClusterPoolAssignmentConfiguration().GetClusterPoolAssignments()[request.GetDomain()]
+
+	return &admin.ClusterAssignment{
+		ClusterPoolName: clusterPoolAssignment.Pool,
```
pmahindrakar-oss, line 450: What happens in case this comes as an empty pool name

pkg/runtime/interfaces/cluster_pools.go
```diff
@@ -0,0 +1,17 @@
+package interfaces
+
+//go:generate mockery -name ClusterPoolAssignmentConfiguration -output=mocks -case=underscore
+
+type ClusterPoolAssignment struct {
+	Pool string `json:"pool"`
+}
+
+type ClusterPoolAssignments = map[DomainName]ClusterPoolAssignment
```
pmahindrakar-oss, line 9: Are these strictly going to be defined only per domain, what if we needed it also at project-domain level or is that not something we are considering for this

pkg/manager/impl/execution_manager.go
katrogan: Existing behavior is preserved, we just don't assign a clusterpool which is fine.

pkg/runtime/interfaces/cluster_pools.go
katrogan: Already implemented as a matchable attribute https://github.com/flyteorg/flyteadmin/pull/376

pkg/runtime/cluster_pool_assignment_provider.go
katrogan: we're inconsistent :(
see https://github.com/flyteorg/flyteadmin/blob/master/flyteadmin_config.yaml#L165 and https://github.com/flyteorg/flyteadmin/blob/master/flyteadmin_config.yaml#L192

but I think the convention leans to camel case, updated

pmahindrakar-oss (approved):

PR #606, opened 2023-08-22
Validate sort order in admin.ResourceListRequest

iaroslav-ciupin:
# TL;DR
Validate sort order by checking it against supported columns. This PR only checks sort order in `admin.ResourceListRequest`, other requests will be added in separate PR. Also validation for filters will be added in separate PR.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/pkg/manager/impl/execution_manager.go
+++ b/pkg/manager/impl/execution_manager.go
@@ -6,49 +6,42 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/flyteorg/flytestdlib/promutils/labeled"
-
-	"github.com/flyteorg/flyteadmin/plugins"
-
+	"github.com/benbjohnson/clock"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/flytek8s"
-
-	"github.com/flyteorg/flyteadmin/auth"
-
-	"github.com/flyteorg/flyteadmin/pkg/manager/impl/resources"
-
-	dataInterfaces "github.com/flyteorg/flyteadmin/pkg/data/interfaces"
 	"github.com/flyteorg/flytestdlib/contextutils"
+	"github.com/flyteorg/flytestdlib/logger"
 	"github.com/flyteorg/flytestdlib/promutils"
+	"github.com/flyteorg/flytestdlib/promutils/labeled"
+	"github.com/flyteorg/flytestdlib/storage"
+	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/grpc/codes"
 
-	"github.com/flyteorg/flyteadmin/pkg/common"
-
-	"github.com/flyteorg/flytestdlib/logger"
-	"github.com/flyteorg/flytestdlib/storage"
-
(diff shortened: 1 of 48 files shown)


pkg/manager/impl/execution_manager.go
```diff
@@ -1434,12 +1427,10 @@ func (m *ExecutionManager) ListExecutions(
...
 		return nil, err
 	}
-	var sortParameter common.SortParameter
-	if request.SortBy != nil {
-		sortParameter, err = common.NewSortParameter(*request.SortBy)
-		if err != nil {
-			return nil, err
-		}
+
+	sortParameters, err := common.NewSortParameters(&request, gormimpl.ExecutionColumns)
```
katrogan, line 1431: Because we [join](https://github.com/flyteorg/flyteadmin/blob/master/pkg/repositories/gormimpl/execution_repo.go#L79-L97) on multiple tables for the list request we should support sort columns from any of the supported tables

See https://docs.flyte.org/en/latest/concepts/admin.html#putting-it-all-together for how we document table_name.field as an option sort

iaroslav-ciupin: @katrogan from what I see from code and docs, we are joining with tables references in filters, not in sort_by. 
I was able to successfully perform this query `https://dogfood.cloud-staging.union.ai/api/v1/executions/flytesnacks/development?limit=5&filters=eq(workflow.name,ray_example.ray_example.ray_workflow)`

But I got an error for
`https://dogfood.cloud-staging.union.ai/api/v1/executions/flytesnacks/development?limit=5&sort_by.key=workflow.name&filters=eq(workflow.name,ray_example.ray_example.ray_workflow)`
and 
`https://dogfood.cloud-staging.union.ai/api/v1/executions/flytesnacks/development?limit=5&sort_by.key=workflow.name`
Error
```
{
  "error": "cannot query with specified table attributes: missing FROM-clause entry for table \"workflow\"",
  "code": 3,
  "message": "cannot query with specified table attributes: missing FROM-clause entry for table \"workflow\""
}
```

katrogan: thanks for the verification @iaroslav-ciupin 
do you mind filing a follow-up issue to ensure that we update [JoinTableEntities](https://github.com/flyteorg/flyteadmin/blob/6c97bb9c5558af240b2d67290bb35393990678cb/pkg/repositories/interfaces/common.go#L27) to include other tables when they're included in the sort key as well?

katrogan (approved):

pkg/manager/impl/execution_manager.go
katrogan: Actually, can you still ensure that we pass in the join table columns here as allowed columns?

katrogan (changes requested):

pkg/manager/impl/execution_manager.go
iaroslav-ciupin: @katrogan I don't get, why? we currently don't support sorting by join table columns and return a DB error for that? wouldn't it be better to return a clearer validation error ?

katrogan: ah I see - missed your second example, so even for the successful join call, e.g.
`/api/v1/executions/flytesnacks/development?limit=5&filters=eq(workflow.name,ray_example.ray_example.ray_workflow)&sort_by.key=workflow.created_at&sort_by.direction=DESCENDING`

we still fail to sort by so we're not removing functionality in this change

iaroslav-ciupin: @katrogan exactly, although we specify join table column in filter and we still can't sort by join table column

PR #607, opened 2023-08-28
Fix SQL injection in List filters

iaroslav-ciupin:
# TL;DR
Validate filters in List API endpoints so that they only contain supported columns and columns of joined tables. 
SQL injection of sort keys is done in separate [PR](https://github.com/flyteorg/flyteadmin/pull/606) 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/pkg/manager/impl/util/filters.go
+++ b/pkg/manager/impl/util/filters.go
@@ -3,6 +3,7 @@ package util
 
 import (
 	"context"
+	"fmt"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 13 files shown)


pkg/manager/impl/util/filters.go
```diff
@@ -120,6 +120,54 @@ func prepareValues(field string, values []string) (interface{}, error) {
 	return preparedValues, nil
 }
 
+var allowedJoinTableColumns = map[common.Entity]map[common.Entity]sets.String{
+	common.Execution: {
+		common.Execution:  gormimpl.ExecutionColumns,
```
katrogan, line 125: instead of duplicating the mapping between common.Entity -> allowedcolumns, can we create two sets here?
1) for entity -> allowed join entities
2) for entity -> allowed columns

PR #616, opened 2023-09-22
FlyteAdmin will always add `base_exec_id` unless it is already added

kumare3:
# TL;DR
Automatically adds `base_exec_id` label with the value of current execution or a previous `base_exec_id` if it exists.

Reasons:
1. Make it possible to retrieve all executions launched by the same base execution id (even recursively)
2. users could group executions using their own base exec id
3. flytectl get executions or remote list executions can use this label as a filter to retrieve high level progress of all subworkflows

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

### DIFF
--- a/pkg/manager/impl/execution_manager.go
+++ b/pkg/manager/impl/execution_manager.go
@@ -553,10 +553,7 @@ func (m *ExecutionManager) launchSingleTaskExecution(
 	if err != nil {
 		return nil, nil, err
 	}
-	labels, err = m.addBaseExecutionLabel(ctx, workflowExecutionID.Name, labels)
-	if err != nil {
-		return nil, nil, err
-	}
+	labels = m.addBaseExecutionLabel(ctx, workflowExecutionID.Name, labels)
 
 	var annotations map[string]string
 	if executionConfig.Annotations != nil {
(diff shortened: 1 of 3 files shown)


katrogan (approved):
LGTM

wild-endeavor:
This isn't a minor change right?

pkg/manager/impl/execution_manager.go
```diff
@@ -1687,6 +1690,19 @@ func (m *ExecutionManager) addProjectLabels(ctx context.Context, projectName str
 	return initialLabels, nil
 }
 
+// Adds base execution label to execution labels. Base execution label is ignored if a corresponding label is set on the execution already.
```
EngHabu, line 1693:
```suggestion
// addBaseExecutionLabel adds base execution label to execution labels. Base execution label is ignored if a corresponding label is set on the execution already.
```

pkg/manager/impl/shared/constants.go
```diff
@@ -33,6 +33,9 @@ const (
 	Attributes            = "attributes"
 	MatchingAttributes    = "matching_attributes"
 	// Parent of a node execution in the node executions table
-	ParentID        = "parent_id"
-	WorkflowClosure = "workflow_closure"
+	ParentID                = "parent_id"
+	WorkflowClosure         = "workflow_closure"
+	BaseExecutionIDLabelKey = "base_exec_id"
```
EngHabu, line 38: Should we call it `parent_exec_id`? just to keep names consistent... I understand that for "self" execution that will sound weird... 🤷🏽 or maybe keep them as two separate labels?

eapolinario:
Hi, we are moving all Flyte development to a monorepo. In order to help the transition period, we're moving open PRs to monorepo automatically and your PR was moved to https://github.com/flyteorg/flyte/pull/4140. Notice that if there are any conflicts in the resulting PR they most likely happen due to the change in the import path of the flyte components.

